PR #1, opened 2018-02-22
Fixed trailing / behind resource in registration API. As described in…

cylonbrain:
… RegistrationAPI "/resource"

### DIFF
--- a/Development/nmos/node_registration.cpp
+++ b/Development/nmos/node_registration.cpp
@@ -77,7 +77,7 @@ namespace nmos
                 });
 
                 // block and wait for the response
-                auto response = client.request(web::http::methods::POST, make_api_version(registry_version) + U("/resource/"), body).get();
+                auto response = client.request(web::http::methods::POST, make_api_version(registry_version) + U("/resource"), body).get();
 
                 if (web::http::status_codes::OK == response.status_code())
                     slog::log<slog::severities::more_info>(gate, SLOG_FLF) << "Registration updated for " << type.name << ": " << id;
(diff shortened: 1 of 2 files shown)


garethsb-sony:
Thanks for the PR!

I think the proposed change to nmos/registration_api.cpp (used in **nmos-cpp-registry**) is incorrect as things stand. The current implementation follows the example in the specification at:
https://github.com/AMWA-TV/nmos-discovery-registration/blob/v1.2.x/examples/registrationapi-v1.1-base-get-200.json

However I agree the first change to nmos/node_registration.cpp (used in **nmos-cpp-node**) for the POST to /resource may be a good idea. This has been discussed previously, resulting in @andrewbonney drafting the following guidance in https://github.com/AMWA-TV/nmos-discovery-registration/issues/15:

> A server implementing the APIs MUST support PUT/PATCH/POST/DELETE/OPTIONS (ie. write related) operations on the non-trailing slash version of each resource. The implementation MAY additionally support the same methods on the trailing slash version but this is not strictly required.

Unfortunately, I don’t think that has yet been merged into the spec. Thanks for reminding me of this! It’s possible that when it is, the response for the API “base” response should be changed also.

As for the other change to nmos/node_registration.cpp for the DELETE, unless I’m mistaken the ``path`` to be appended at that point doesn’t contain a leading slash, so the trailing slash on ``"/resource/"`` is required?

garethsb-sony:
Resolved by commit 7cd46cdf97181befebaf1a4e4825aed5b1378a75.

PR #2, opened 2018-03-05
Fix segfault in case of adding resources dynamically

TSPMP:
Resource(s) Added Events (see
https://github.com/AMWA-TV/nmos-discovery-registration/blob/v1.2/docs/4.2.%20Behaviour%20-%20Querying.md)
only have a `post` field but no `pre` field. Hence, the condition for `creation` was erroneous and
the segfault occurred when we accessed the `pre` field without checking for existence.


garethsb-sony:
Many thanks for the clear report and patch. I would like to add a few additional comments to the patched code and will then merge.

TSPMP:
Sure!

PR #4, opened 2018-03-06
Fix wrong type for `active` field in senders and receivers

TSPMP:
There is no implicit `web::json::value` constructor for `bool`s, but there are implicit constructors for numbers. C++ implicit conversions strike again! Result: the Node API server delivers a JSON value with `"active": 0` instead of `"active": false`

### DIFF
--- a/Development/nmos/node_resources.cpp
+++ b/Development/nmos/node_resources.cpp
@@ -197,7 +197,7 @@ namespace nmos
 
             value subscription;
             subscription[U("receiver_id")] = value::null();
-            subscription[U("active")] = value(false);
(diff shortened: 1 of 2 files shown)


garethsb-sony:
Thank you for the PR, @TSPMP. With a quick review, I found one other likely mistake. Unit tests required... (I consider the explicit `bool` constructor vs. implicit `int` constructor fiasco to be a bug in cpprestsdk's `json::value` class, which I've kept meaning to open an issue/PR for.)

TSPMP:
I agree that this is either a bug or questionable design in that framework.

PR #22, opened 2018-08-13
Handle IS-05 GET requests.

billt-hlit:
These are the changes I shared with Gareth and Martin in Wuppertal for implementing IS-05 GET requests. Martin pushed these to github as a single combined patch. This pull request contains each of my local commits in excruciating detail.


garethsb-sony:
Thanks again for this,  Bill. Merged into is05 branch as part of #23. Let's get this to master!

PR #33, opened 2018-12-01
Jwp/cmake modules

jwpwh:
These changes split the CMakeLists.txt file into a few pieces so that it is easier to reuse these portions when nmos-cpp is included with an existing backend which is defining the different senders and receivers. 

The nmos files are put into a library and used in the nmos-cpp-node and nmos-cpp-registry executables. 

The various pieces are the general options and dependency checking, making the libraries, and making the registry test program. 

The nmos_resources_thread.* files were moved into the nmos-cpp-node directory since they (along with main.cpp) are likely changed by anyone including nmos-cpp in a different application. As such, they don't seem to belong in the nmos library.

I picked the nmos-cpp up in early November and included it in my application. When I got an update, the cmake files I made the first time needed quite a bit of rework. With the pieces as shown here, I would be able to just reuse these pieces with no changes on my side.

Thanks.


garethsb-sony:
I've been meaning to get around to this for a while, so will review and would hope to merge ASAP. Thanks.

Point about ``nmos::experimental::node_resources_thread`` is well made. We hope to bring some parts of it into the library, but at the moment it does 'belong' to the example **nmos-cpp-node** so I may move it to that directory as part of this.

garethsb-sony:
Looked good. Thank you!

jwpwh:
Thanks.

-----Original Message-----

PR #34, opened 2018-12-03
Added #ifdef __cpluscplus to protect .c files from boost C++ header

jwpwh:
I have a .c file in my application and while the 
/FI\"${NMOS_CPP_DIR}/cpprest/details/boost_u_workaround.h\" 

in NmosCppCommon.cmake seems like it should only apply to C++ files, in practice, it seems to apply to .C files also, even when they have been marked as having the LANGUAGE C.

set_source_files_properties (example.c PROPERTIES LANGUAGE C)

By adding #ifdef __cplusplus in the boost_u_workaround.h file, the fight with Cmake is avoided and the the .c file avoids the errors which occur when the boost file type_traits.hpp is included in a C file.


garethsb-sony:
Yep, that's not worth fighting. 😆
Thanks for the PR.

PR #35, opened 2018-12-06
Application hooks

jwpwh:
After updating to your latest changes, I decided to add "application hooks" into the nmos-cpp-node files along with default implementations so that my changes are isolated to replacing the default implementations instead of changing node_implementation.cpp.

As part of this, the creation of the sample resources was moved to a separate function which is called from one of the default application hook. I also added sample audio resources in addition to the sample video resources. The audio resources are not duplicated and the resolve_auto function was changed to check the size of the transport_params instead of assuming there were 2 elements.

The main.cpp file was split with most of the code going to node_main_thread.cpp so that the main() function doesn't have to be included and doesn't collide with the main() in my app.

In addition, when updating the IS-05 connection resource on an activation, there was a call to set_connection_sender_transportfile with the original sdp_params.  After moving the creation of the resources to a separate function, sdp_params was no longer defined. With only 1 sender and only 1 sdp file ever defined, this call probably doesn't hurt anything. With multiple senders, it does not seem correct and it seems like the receiver gets the appropriate sdp file even without the call.

When the resources are created, the ip addresses were generated in a slightly different way so that different addresses are used for each resource.


garethsb-sony:
Nice idea. We have a couple of Node implementations we can try this out with.
@simonlo-sony Let's review.

A couple of comments at first glance:

* We were just discussing yesterday pulling the example resource construction out of the ``node_implementation_thread``. Definitely a win.
* I hope the 'boundary' between the nmos-cpp-node example ``main`` and the ``node_main_thread`` function could change slightly, or maybe there need to be a couple more application hooks (e.g. for logging config), in order to make ``node_main_thread`` more generally useful.
* I believe the call to ``set_connection_sender_transportfile`` was there so that when a sender is activated, its SDP file is updated to reflect transport_params that have changed. And the SDP can't be regenerated without the non-transport stuff which is in ``sdp_params``. Does that make sense? We will look closer at this!

If this hangs together, we could end up putting ``node_implementation_thread`` back into ``nmos::experimental``, and perhaps the ``node_main(_thread)`` function too. Or how about those two functions are merged, and the example ``main`` pretty much just starts a thread to run that and then does ``wait_term_signal``...

jwpwh:
I am assuming that additional hooks would be added or the boundaries could move. I tried to make them generic but I was just starting with the hooks I added. Mileage may vary.

Currently, I am driving the activations with the Riedel nmos explorer (as well as my own app) and the changes are mainly telling a receiver to listen or not listen to a sender. This doesn't changes the sender sdp at all. Assuming there are other cases (like maybe unicast), if the call to set_connection_sender_transportfile is needed, then if multiple senders are defined, does some sort of cache of the original sdp_params need to be kept in order to have the sdp_params for the particular sender available? I think the call was just setting the sdp to the original parameters so it wasn't creating a new sdp. Maybe there was a missing line which created the sdp_params?

These changes were made to provide an example to help move in this direction. I am still pretty new to this and know there are probably other aspects to consider. I like the separation so that it is easier to pick up recent commits without having to rework a lot of code. This obviously depends on what the recent commits are changing. The application hooks help with the updates and are also suggestive as to what a new node implementation needs to provide.

Thanks. 


-----Original Message-----

garethsb-sony:
Up until now, code under ``nmos-cpp-node/`` has been supposed to be example code you might want to copy for a real Node implementation, whereas code under ``nmos/`` was supposed to be reusable as-is. As you found and fixed with your last PR, there were one or two bits in ``nmos/`` that were really example code. You're right with this new PR, there's more of the code that can go back the other way. As you say, ``nmos-cpp-node`` should serve as both a simulated Node for testing and as a demo of how a real Node can hook into the ``nmos-cpp`` implementation.

When we're considering the ``set_connection_sender_transportfile`` call, that's only applicable when processing an activation of a Sender, not a Receiver. When you stage transport_params to a sender, the sender's SDP file needs to change to reflect that. The sdp_params capture the *other* information that goes in an SDP file, they must be combined with the transport_params to make the actual SDP file. There are functions to make an *appropriate* sdp_params from the IS-04 Source, Flow and Sender, or you could parse the current SDP file and make sdp_params from it. Or given these parts of the SDP file can't be changed by an NMOS API, you could do what we had done and fix them for the life of the application - though obviously you need one set for each Sender.

garethsb-sony:
Just noting here that when we look at this we need to fix up the out-of-date documentation describing [vendor-specific Node behaviour] (Documents/Architecture.md#vendor-specific-node-behaviour)

garethsb-sony:
Superseded by #49.

PR #49, opened 2019-01-19
Origin/application hooks 3

jwpwh:
This is a second stab at the application hooks. This tries to address the issues you raised with the set_connection_sender_transportfile and activating a sender. You were quite correct that this was an issue. 

This new version has all the sample resources generated and managed in main.cpp. The sample resources have been added to so that both audio and video examples are created. They are managed so that the sender activations can be done and get the correct information for the sender being activated instead of just always generating hard-coded information which works ok for a single sender/receiver but not in anything more general.  

This version uses function pointers which can be lambdas instead of specific function names. Actually, a structure with a number of function pointers is passed in since this seemed cleaner than listing all the functions.

I think this version is current with your most recent changes.

### DIFF
--- a/Development/CMakeLists.txt
+++ b/Development/CMakeLists.txt
@@ -17,6 +17,7 @@ include (${NMOS_CPP_DIR}/cmake/NmosCppLibraries.cmake)
 
 set(NMOS_CPP_NODE_SOURCES
     ${NMOS_CPP_DIR}/nmos-cpp-node/main.cpp
+    ${NMOS_CPP_DIR}/nmos-cpp-node/node_main_thread.cpp
     ${NMOS_CPP_DIR}/nmos-cpp-node/node_implementation.cpp
     )
 set(NMOS_CPP_NODE_HEADERS
--- a/Development/mdns/service_discovery.h
+++ b/Development/mdns/service_discovery.h
@@ -34,6 +34,7 @@ namespace mdns
     };
 
     // return true from the browse result callback if the operation should be ended before its specified timeout once no more results are "imminent"
+    // the callback must not throw
     typedef std::function<bool(const browse_result&)> browse_handler;
 
     struct resolve_result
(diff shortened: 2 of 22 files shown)


garethsb-sony:
Thanks for persevering with this. Now that we're back from the AMWA Workshop, I should have a better chance to review and merge!

pkeroulas:
@jwpwh, this doesn't look like a trivial conflict, is there any chance you could update this branch? Thank you.
Regards,
PK

garethsb-sony:
Hi @pkeroulas, I bear responsibility for conflicts here really, as @jwpwh has been kind enough to rebase and make some changes base on feedback on his original PR, but I haven't had time to invest in evaluating how much of this we want to take to master. It _is_ still on my list of things to do...

jwpwh:
I am looking at making another update to this and wanted to make sure you were not starting to look at this yourself or if you had any comments you wanted to pass along.

Thanks.

pkeroulas:
@jwpwh, correct me if I'm wrong, this feature would allow to hook a st2110-ready player/transcoder (like ffmpeg) in order to build a NMOSified video receiver. That would be really really appreciated. Let me know if I can help anyhow. Thanks.

jwpwh:
The purpose of this change was to reorganize the example node so that example specific code like the creation of the example resources would be compartmentalized in "application hook" functions. By doing this, it would make it easier to provide a different sent of application hooks which would provide real resources based on, in my case, real signals from an existing system. There is still considerable work to write the alternate set of hooks. The hoped for advantage was that additional changes could be made to the example node, and it would be easier for my application to pick up the new changes. And the changes made would be more obvious as to whether they are just specific examples (code in one of the hook functions) or something more generic like the implementation thread.

A similar set of changes was suggested for the CMake files and those changes were incorporated. (thank you) And it has been much very easy to pick up new changes to the CMake files. These changes are a bit messier so I understand why there is hesitation to pick them up.

Making these changes up-to-date with the current master involves adding some changes for the event/tally examples and probably adding one new hook for processing events. The changes are not real complicated but I need to run the test suites and I am not real adept with Git so it takes a bit more time than I would like. I expect to be done with this latest update around the middle of next week.

garethsb-sony:
Superseded by #77.

PR #52, opened 2019-01-30
New "settings_address" setting

billt-hlit:
Proof-of-concept: new "settings_address" setting to make it possible to specify the (one) IP address to which the settings API will be bound at start-up.

I've limited my changes to this one setting not (only) because I'm lazy, but because I expect there will be compilation problems on other platforms and would rather give you the chance to sort those out before other services get the ability to be bound to a specific address. Furthermore, the settings API was the interface I really wanted to specifically bind the listening address of.

### DIFF
--- a/Development/cpprest/api_router.h
+++ b/Development/cpprest/api_router.h
@@ -23,17 +23,17 @@ namespace web
                 // RAII helper for http_listener sessions (could be extracted to another header)
                 typedef pplx::open_close_guard<http_listener> http_listener_guard;
 
+#if defined(_WIN32) && !defined(CPPREST_FORCE_HTTP_LISTENER_ASIO)
+                const auto host_wildcard = _XPLATSTR("*"); // "weak wildcard"
+#else
+                const auto host_wildcard = _XPLATSTR("0.0.0.0");
+#endif
+
                 // using namespace api_router_using_declarations; // to make defining routers less verbose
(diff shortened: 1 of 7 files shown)


garethsb-sony:
Hi @billt-hlit,

Having implemented support for forward and reverse proxies recently, I've been looking at this next.
One thing that I've been wondering about is whether we should allow the same port number to be set up for both APIs that will bind to the wildcard address, and for specific address(es).

Do you feel strongly whether we should allow or forbid the following?

```json
{
  "settings_address": "127.0.0.1",
  "settings_port": 80,
  "node_port": 80
}
```

I think behaviour is platform specific, but at least in my initial tests, what I see is that specific-address binding wins vs. wildcard-address binding, e.g.

http://203.0.113.42:80/ contains the Node API, as you'd expect
http://127.0.0.1:80/ contains only the Settings API, not both

billt-hlit:
As a Unix hacker, the behavior you're seeing is what I would expect. Whether it is particularly (un)desirable -- I don't have a strong opinion. For my purposes, the behavior you demonstrate might even be useful, although I don't plan on overloading the port numbers.

garethsb-sony:
There's an interaction with the "proxy_map" configuration setting that I'm a little uncomfortable with, but let's see how these various settings get used.

PR #61, opened 2019-02-26
Log PUT, PATCH, and POST requests using a new "change log message category.

billt-hlit:
I need to log attempts to change settings differently, hence this new category.

I also have a change adding request::remote_address() to api_stash, but that seems trivial and I'm not sure anyone else is interested.

### DIFF
--- a/Development/nmos/connection_api.cpp
+++ b/Development/nmos/connection_api.cpp
@@ -890,7 +890,7 @@ namespace nmos
 
                 auto& patches = body.as_array();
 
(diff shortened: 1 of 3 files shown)


garethsb-sony:
I understand the desire. Would using the existing stashed http method work?

billt-hlit:
After an out-of-band discussion with Gareth, I've concluded I can combine the stashed request's http_method with the "access" category to detect configuration change requests.

PR #68, opened 2019-04-17
IS-07 maweit/nmos-cpp implementation

garethsb-sony:
[AMWA IS-07 NMOS Event & Tally](https://github.com/AMWA-TV/nmos-event-tally/)

I am making a PR out of @maweit's experimental implementation (Dec 2018) to facilitate discussion and comparison with the implementation in #59 by @bakaleks (Feb 2019). Thank you both for your contributions.

My intention is to create a new branch, merge the uncontroversial bits of both PRs, and proceed from there.

PR #69, opened 2019-04-24
IS-07 Event & Tally

garethsb-sony:
Here's my start on IS-07 Event & Tally, strongly influenced by @maweit PR #68 and @bakaleks PR #59. I'll cherry-pick and attribute where possible. So far, I think I already had the underlying WebSocket server bits necessary, and I've now taken the Events API itself from #59, checked against #68, just with a few renames, extra specs links, etc.

I'll be looking at the factory functions for IS-07 event state/type next, and then the WebSocket implementation.

Review would be much appreciated!

### DIFF
--- a/Development/boost/asio/ssl/set_cipher_list.hpp
+++ b/Development/boost/asio/ssl/set_cipher_list.hpp
@@ -9,6 +9,11 @@
 
 #include <boost/asio/detail/push_options.hpp>
 
+#ifndef BOOST_ASIO_SYNC_OP_VOID
+# define BOOST_ASIO_SYNC_OP_VOID void
+# define BOOST_ASIO_SYNC_OP_VOID_RETURN(e) return
+#endif
+
 namespace boost {
 namespace asio {
 namespace ssl {
--- a/Development/boost/asio/ssl/use_tmp_ecdh.hpp
+++ b/Development/boost/asio/ssl/use_tmp_ecdh.hpp
@@ -11,6 +11,11 @@
 
 #include <boost/asio/detail/push_options.hpp>
 
+#ifndef BOOST_ASIO_SYNC_OP_VOID
+# define BOOST_ASIO_SYNC_OP_VOID void
+# define BOOST_ASIO_SYNC_OP_VOID_RETURN(e) return
+#endif
+
 namespace boost {
 namespace asio {
 namespace ssl {
--- a/Development/cmake/NmosCppLibraries.cmake
+++ b/Development/cmake/NmosCppLibraries.cmake
@@ -566,6 +566,7 @@ set(NMOS_CPP_NMOS_HEADERS
     ${NMOS_CPP_DIR}/nmos/interlace_mode.h
     ${NMOS_CPP_DIR}/nmos/is04_versions.h
     ${NMOS_CPP_DIR}/nmos/is05_versions.h
+    ${NMOS_CPP_DIR}/nmos/is07_versions.h
     ${NMOS_CPP_DIR}/nmos/json_fields.h
     ${NMOS_CPP_DIR}/nmos/json_schema.h
     ${NMOS_CPP_DIR}/nmos/log_gate.h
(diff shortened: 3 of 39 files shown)


garethsb-sony:
FWIW, I've also now got several [open issues/PRs](https://github.com/AMWA-TV/nmos-event-tally/issues?q=+is%3Aopen+author%3Agarethsb-sony) on the IS-07 spec itself.

bakaleks:
Great Gareth. You may try to run you change towards IS05 tests from my pull request to AMWA https://github.com/AMWA-TV/nmos-testing/pull/162

Running it towards my pull-request #59 uncovered that we need special treatment for non-RTP senders/receivers in connection_api.cpp and node_implementation.cpp when doing IS05 activations, in order to not to trigger SDP-related stuff. Please check latest update in the #59 (it's a hack, maybe you come up with something better). I also discovered issue with my websocket implementation that is fixed in the same update, but you might as well do something completely different for that part.

garethsb-sony:
Thanks, Alexei. Yes, once I've got a bit further, I'll definitely look at the nmos-testing PR.

I think we could do with some factory functions in node_resources.cpp/.h or elsewhere for the IS-07 event type/state JSON objects. Does that make sense?

garethsb-sony:
> we need special treatment for non-RTP senders/receivers in connection_api.cpp and node_implementation.cpp when doing IS05 activations

That makes sense as well. I'll pay attention to your work on #59. I hope once I get a bit further with this #69, you could potentially rework what you're doing on top of this.

garethsb-sony:
> I also discovered issue with my websocket implementation

If you mean your websocket_listener changes, yes, I think I saw several bugs in those - but I think I've got all the necessary features in current master, so I hope you/we have everything we need in that?

bakaleks:
> Thanks, Alexei. Yes, once I've got a bit further, I'll definitely look at the nmos-testing PR.
> 
> I think we could do with some factory functions in node_resources.cpp/.h or elsewhere for the IS-07 event type/state JSON objects. Does that make sense?
Absolutely

garethsb-sony:
Making progress after a few days side-tracked onto other tasks...

> > I think we could do with some factory functions in node_resources.cpp/.h or elsewhere for the IS-07 event type/state JSON objects. Does that make sense?
>
> Absolutely

We've now got Events API state/event and type factory functions, so you can now write things like: https://github.com/sony/nmos-cpp/pull/69/commits/dd88c903fa17161afc76763f9b2874a47bc3e741#diff-b4d387b43c81adb747478b0dba692964. (The factory function declarations are in: https://github.com/sony/nmos-cpp/pull/69/commits/dd88c903fa17161afc76763f9b2874a47bc3e741#diff-0946bb1794b79acc9306283ecb701680.)

garethsb-sony:
As of 1c907490be484f874997c86d11901c58d423f4b0, the Events WebSocket API implementation is basically working and demonstrated in nmos-cpp-node/node_implementation.cpp (I used async tasks in `node_implementation_thread` rather than a separate thread, just for simplicity).

What's left is just expiry, 12 seconds after the last heartbeat.

garethsb-sony:
While I remember, I normally use [Postman](https://www.getpostman.com/) for manual tests of the REST APIs (PATCH, POST, DELETE). For WebSockets, I've just found the Chrome [Simple WebSocket Client extension]https://chrome.google.com/webstore/detail/simple-websocket-client/pfdhoblngboilpfeibdedpjgfnlcodoo, which seems to work well for manual testing.

garethsb-sony:
It's not had a lot of testing yet, but I believe this now represents complete Events API (WebSocket) functionality.

bakaleks:
It all looks clean and logical, I especially like how you reworked events_ws_api.cpp

garethsb-sony:
> It all looks clean and logical, I especially like how you reworked events_ws_api.cpp

Thanks, that's encouraging! I actually credited you with git authorship of several of the commits, because in places I was really only tweaking what you'd done, double-checking and refactoring with the Query WebSocket API, and Registry expiry, implementations.

It's reassuring to be able to reuse so much code, it suggests we've got reasonable consistency between NMOS specs' required behaviours, even if there's perhaps a bit too much copy-and-paste rather than actual common implementation.

Some time today after a bit more testing on various platforms, I'll merge this to master. Thanks for your original work and reviews.

garethsb-sony:
I've just realised that this doesn't include an implementation of an IS-07 WebSocket *Receiver*... I can rectify some of this (resource construction) easily, but I may not get around to the WebSocket client implementation for a whil.e (I have Query WebSocket client code in a private repo that could serve as inspiration.)

PR #71, opened 2019-05-03
Creating install target for libraries generated

tassianors:
Signed-off-by: Tassiano Neuhaus <tneuhaus@rossvideo.com>

### DIFF
--- a/Development/cmake/NmosCppLibraries.cmake
+++ b/Development/cmake/NmosCppLibraries.cmake
@@ -655,6 +655,11 @@ add_library(
     ${NMOS_CPP_SLOG_HEADERS}
     )
 
+set(DETAIL_HEADERS
+    ${NMOS_CPP_DIR}/detail/for_each_reversed.h
+    ${NMOS_CPP_DIR}/detail/private_access.h
+    )
+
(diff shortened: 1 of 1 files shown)


tassianors:
Hi guys, my intention with this commit is to allow me to install the libraries generated using the cmake infra instead of working around it and copying it manually.

thanks for your time.

garethsb-sony:
Thank you, Tassiano. Having some install targets is a good idea. I wonder whether we should also be annotating the include directories too?

tassianors:
Hi  Gareth,

I will check what headers are being used outside the repo and add them to
this review latter today.

thanks

  Tassiano Neuhaus
  +1 613 700 3587

tassianors:
hi Gereth, 

I've updated the patch with the installation of the header files that are needed from an external application using the libs produced by this repo, please let me know if the solution is aligned with your expectations.

thanks.

garethsb-sony:
Hi Tassiano,

Thanks. I agree that `install(FILES ...)` using the existing list variables is perfect for now.
I've just tweaked it a little and merged to master. Let me know if this doesn't do what you need!

PR #77, opened 2019-05-29
App hooks 4

jwpwh:
This is similar to the App hooks 3 except based on current code. The change is mostly rearranging some files. 

- App startup in main.cpp has mostly moved intact to the new file node_main_thread.cpp.
- The code in main.cpp represents the code specific to the sample application. 
  - It provides some sample application hook functions and then starts up the main thread.
  - The sample resources are created and managed within main.cpp
- The code in node_implementation.cpp is mostly intact except for these two types of changes
  - The code to create and manage the sample resources has moved to main.cpp
  - Calls to the application hooks are placed at strategic locations.

- The code to periodically update the sample events is in main.cpp. The code which used to directly update the events now indicates that there is work to do and then does the work when the application hook is called.

I ran the tests and get the same results as the master code. Lots of assertions but the tests pass.

I have noticed that when the application (both the master and the changed code in this pull request) is terminated (press the exit button in upper right of the console window on windows) that the application will frequently take an exception. Assume there is a thread someplace referring to an object which has been deleted but not sure.

### DIFF
--- a/Development/CMakeLists.txt
+++ b/Development/CMakeLists.txt
@@ -18,6 +18,7 @@ include (${NMOS_CPP_DIR}/cmake/NmosCppLibraries.cmake)
 set(NMOS_CPP_NODE_SOURCES
     ${NMOS_CPP_DIR}/nmos-cpp-node/main.cpp
     ${NMOS_CPP_DIR}/nmos-cpp-node/node_implementation.cpp
+    ${NMOS_CPP_DIR}/nmos-cpp-node/node_main_thread.cpp
     )
 set(NMOS_CPP_NODE_HEADERS
     ${NMOS_CPP_DIR}/nmos-cpp-node/node_implementation.h
--- a/Development/nmos-cpp-node/main.cpp
+++ b/Development/nmos-cpp-node/main.cpp
@@ -1,222 +1,493 @@
 #include <fstream>
 #include <iostream>
-#include "cpprest/ws_listener.h"
-#include "cpprest/ws_utils.h"
+#include "pplx/pplx_utils.h" // for pplx::complete_after, etc.
+#include "cpprest/host_utils.h"
 #include "nmos/admin_ui.h"
 #include "nmos/api_utils.h"
 #include "nmos/connection_api.h"
-#include "nmos/events_api.h"
-#include "nmos/events_ws_api.h"
-#include "nmos/log_gate.h"
 #include "nmos/logging_api.h"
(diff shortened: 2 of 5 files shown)


jwpwh:
Another change: The CMakeLists.txt file was changed to add the new node_ain_thread.cpp file.

garethsb-sony:
> Lots of assertions but the tests pass.

Failing assertions? The nmos-cpp-test app or the NMOS Test Suites?

> when the application (both the master and the changed code in this pull request) is terminated (press the exit button in upper right of the console window on windows) that the application will frequently take an exception. 

I've seen this... It seems to depend on build options for cpprestsdk. It would be good to track down.

garethsb-sony:
You've convinced me of the need to put `nmos_implementation_thread` back into the nmos/ module, and to give it a hook for connection resource activation (covering resolve auto, etc.). With the other GitHub issues on SDP parsing (and to support other transport types for IS-07), I'm also convinced of the need for the Connection API implementation to have a hook for connection resource staging (covering SDP parsing, etc.).

I'm not yet convinced of the need to have resource creation or the check for/process work hooks in the impl thread. Other app threads can work with the model, its mutex and condition variable, directly.

You've also convinced me the bulk of the `main` function should also be pulled into the nmos/ module. I hope this can be achieved with a factory function rather than a thread, but I haven't fully thought it through yet.

jwpwh:
I think what you have described sounds reasonable and would be a good direction to move to.

Thanks.

garethsb-sony:
As of https://github.com/sony/nmos-cpp/commit/9ce86373049c62e8a34c16fc763fd6f8a134786f on master, I hope we've finally got something that is easier to integrate into an existing app. Thank you for inspiration and patience!

jwpwh:
I was on vacation most of the last 2 weeks and have not been able to look at the details. The comments sound very good and very workable. After taking care of the vacation pile-up stuff, I should be able to look at this in more detail later this week. Thanks for all of your effort.

garethsb-sony:
> I have noticed that when the application (both the master and the changed code in this pull request) is terminated (press the exit button in upper right of the console window on windows) that the application will frequently take an exception. Assume there is a thread someplace referring to an object which has been deleted but not sure.

Hi @jwpwh, I finally found time to dissect why Ctrl-C exit worked OK, but closing the console window didn't on Windows. It's now fixed for me by a simple addition - 05953292e26c7c0257bf40b98c17d34097c76019.

PR #80, opened 2019-06-26
Fixing compilation problems when the flag -Wswitch-default on dependent proj

tassianors:
it is a very minor problem that I have fixed on my side by removing the flag, but I dont think it would be a problem to have this extra default line on this switch case. main reason for it is that the switch case code is on the header file.


garethsb-sony:
Thanks.

PR #87, opened 2019-07-19
Minor changes related to callback functions

jwpwh:
These are some minor changes to the callback functions to make it more clear what all the available callbacks are and what their arguments are. The example implementation makes all the callbacks instead of just the functions that are not {}. I think the placeholders would help indicate which callbacks are available.

There are also some preprocessor macros for allowing the main() function to be renamed to nmos_main() in case the application can use the body of main() as is. e.g. nmos_main() would be called from the application and the main() in the application would not collide with the sample implementation.

### DIFF
--- a/Development/nmos-cpp-node/main.cpp
+++ b/Development/nmos-cpp-node/main.cpp
@@ -7,7 +7,11 @@
 #include "nmos/server.h"
 #include "node_implementation.h"
 
+#if (NMOS_IS_MAIN == 1)
 int main(int argc, char* argv[])
+#else
+int nmos_main(int argc, char* argv[])
+#endif
 {
     // Construct our data models including mutexes to protect them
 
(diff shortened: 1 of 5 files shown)


garethsb-sony:
Thank you!

I've adapted all of these changes into master with a few minor tweaks (1ba7bbdec82eef03d03dd62a5eb682a99a39876e, 71e1540ff064abab4bd2a152dfb4b35e9bc23e3f), apart from the ability to rename `main` to `nmos_main`. I'm just not convinced that reading config from file and ultimately waiting for `Ctrl-C` is useful except as an example. If it's close to what someone needs, at least after the recent effort, it's not much code to copy & rename.

jwpwh:
Thanks. I am re-using main as is for now. In the long run, this will probably be replaced so not picking up the name change is probably the correct thing to do.

Issue #112, opened 2020-01-24
Making an audio receiver with a single provided bit-depth results in extra value


grimmn:
Calling the form of `make_audio_receiver` from node_resources.cpp which takes a single `bit_depth` argument constructs a vector of bit depths in order to call the other form. I believe that the intention was for the `{ 1, bit_depth }` argument to construct a vector containing one value equal to `bit_depth` but the initializer list constructor takes precedence resulting in the receiver being created supporting the provided bit depth as well as a bit depth of 1.

garethsb-sony:
Resolved by #113

PR #113, opened 2020-01-24
Fix extra audio receiver bit depth capability #112

grimmn:
When providing a single supported bit depth for audio receivers they
would end up getting an extra supported bit depth of 1.


grimmn:
Yeah, I just realized that myself when testing in my setup. Updated my changes to fix that.

garethsb-sony:
That looks good to me.
Apologies for the bug, thanks for catching this!

Are you in Burlington? :-)

grimmn:
No worries, I've been bitten more than once by constructors on vectors of integers.

Yes, I am in Burlington. :D

PR #115, opened 2020-01-29
Add argument to set PTP domain in SDP parameters

grimmn:
Tried out an implementation with `boost::optional` for the optional PTP domain argument. Let me know what you think.

I'm also wondering what you think about, when the PTP domain is not specified but the PTP clock's `traceable` field is set, using the `ptp=<ptp_server>:traceable` format of `ts-refclk` instead? And then only falling back to just putting the GMID with no domain if no domain is given and the clock is not traceable?

### DIFF
--- a/Development/bst/optional.h
+++ b/Development/bst/optional.h
@@ -47,6 +47,7 @@ namespace bst
 {
     using std::nullopt_t;
     using std::nullopt;
+    using std::bad_optional_access;
 }
 
 #elif defined(BST_OPTIONAL_BOOST)
@@ -58,6 +59,12 @@ namespace bst
 {
     typedef boost::none_t nullopt_t;
     const boost::none_t nullopt = boost::none;
+#if BOOST_VERSION >= 105600
+    using boost::bad_optional_access;
+#else
+    // Prior to Boost 1.56.0, boost::optional provided an interface that differed in a number of ways from that ultimately defined for C++17 std::optional
+    // e.g. no member function value
+#endif
 }
 
 #endif
@@ -66,7 +73,6 @@ namespace bst
 {
     using bst_optional::optional;
     using bst_optional::make_optional;
-    using bst_optional::bad_optional_access;
 }
 
 #endif
--- a/Development/bst/optional.h
+++ b/Development/bst/optional.h
@@ -1,0 +1,72 @@
+#ifndef BST_OPTIONAL_H
+#define BST_OPTIONAL_H
+
+// Provide bst::optional, etc. using either std:: or boost:: symbols
+// C++17 feature test macro: __has_include(<optional>), __cpp_lib_optional
+
+#if !defined(BST_OPTIONAL_STD) && !defined(BST_OPTIONAL_BOOST)
+
+#if defined(__GNUC__)
+// std::optional is available from GCC 7.1, with -std=c++17
+//#if __GNUC__  > 7 || (__GNUC__ == 7 && __GNUC_MINOR__ >= 1)
+#if __cplusplus >= 201703L
+#define BST_OPTIONAL_STD
+#else
+#define BST_OPTIONAL_BOOST
+#endif
+
+#elif defined(_MSC_VER)
+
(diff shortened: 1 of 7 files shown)


garethsb-sony:
Looks good, thanks! In other code bases, I'd be completely happy with using `boost::optional` but we've so far managed to avoid it in `nmos` interfaces, so I just have a small pause to consider the right approach here.

Your point about using the traceable form is interesting... We have to consider which holds more useful information, the GMID or the fact that it's traceable, and whether trying to always produce an ST 2110 compliant value overrides the potential surprise that not knowing the domain changes what happens? (In this case, I think it's an oversight in the SDP spec that it doesn't allow you to indicate both traceability and the GMID.) 

I'll have a play here too.

garethsb-sony:
> We have to consider which holds more useful information, the GMID or the fact that it's traceable, and whether trying to always produce an ST 2110 compliant value overrides the potential surprise that not knowing the domain changes what happens?

I garbled that, but what I'm wondering is whether we should actually go further and always indicate `traceable` (if it is). RFC 7273 says "Marking a clock as traceable allows additional information (e.g., IP addresses, PTP master identifiers, and the like) to be omitted from the SDP since any traceable clock available at the answerer is considered to be an appropriate timestamp reference clock."

I.e. `traceable` would always be preferred form, `<gmid>:<domain>` second choice and bare `<gmid>` third.

I think that's reasonable according to the way we wrote the tests in [`IS0502Test.test_17`](https://github.com/AMWA-TV/nmos-testing/blob/ef2719f5bd49965120dea817ee84633843855aaa/nmostesting/suites/IS0502Test.py#L1027-L1032) but I'd like to check what other users think.

grimmn:
I personally like the idea of indicating `traceable` whenever possible. All traceable references should be interchangeable with each other so additional information on which particular one you're using doesn't seem necessary at that point.

tkernen:
I think this is something you want to input into the NMOS TR1001-1 or ST2110-10 1-year review to build a hierarchy between those options. Traceable is generally prefered, AFAICR there is no explicit language that suggests such a hierarchy is used.

andrewbonney:
@tkernen do you have any opinion on how important knowledge of the `gmid` may be for diagnostics? Whilst in a correctly functional system `traceable` ought to be sufficient, it seems helpful to have the `gmid` available in case you are diagnosing an issue where perhaps two `traceable` GMs have in fact drifted, either due to bugs or due to issues in the GPS network for example. Ideally I'd have thought it useful to have all of this information present, but the RFC doesn't permit this.

garethsb-sony:
Good points, thanks! (20:20 hindsight tells us we ought to have had IS-04 PTP clock descriptions include `domain_number`, and we ought to have proposed extending the RFC 7273 attribute to allow `<gmid>[:<domain>]:traceable` as well as the existing `<gmid>[:<domain>]` and `traceable`.)

garethsb-sony:
RFC 7273 allows that "Two or more IEEE 1588 clocks MAY be listed at the same level in the session description to indicate that all of the listed clocks are candidate grandmaster clocks for the domain or deliver the same reference time and may be used interchangeably."

How would you feel about putting two attributes in when we have a traceable clock?

```
a=ts-refclk:ptp=IEEE1588-2008:39-A7-94-FF-FE-07-CB-D0:42
a=ts-refclk:ptp=IEEE1588-2008:traceable
```

garethsb-sony:
Another issue we've identified in discussion is that the NMOS `clocks` are associated to each Source to indicate how that media is clocked. But the SDP `ts-refclk` indicates how the ST 2110 stream is being clocked at the interface. In the case of ST 2022-7 with multiple interfaces, it's possible that each leg could indicate a different clock.

So as well as the 'hierarchy' or multiplicity of `ts-refclk` options that is an issue for ST 2110-10 and/or TR-1001-1, simply mapping from NMOS `clocks` to SDP `ts-refclk` probably can't cover every case.

garethsb-sony:
Thank you, @grimmn!

PR #116, opened 2020-01-30
ST2022-6 SDP creation/parsing

### LINKED ISSUE #74: Feature request: 2022-6 SDP creation/parsing
Hello,

When looking at the implementation I see SDP generation for 2022-6 (mux flows) is not implemented. It wasn't mentioned in the current feature list, so it was expected. Are there any plans for adding support for 2022-6 SDP files?

Thanks!

grimmn:
Should largely resolve #74 most significant missing component is generating the proper group attributes and media descriptions of ST2022-5 FEC streams. I'm not sure what all of the valid or expected combinations of ST2022-5 and ST2022-7 are yet or how they should properly be handled in the SDP so I've left that out for now.

I've included all of the required content in the SDP that I could determine from ST2022-8 although it's not super clear so it's possible that I've missed something.

### DIFF
--- a/Development/nmos/sdp_utils.cpp
+++ b/Development/nmos/sdp_utils.cpp
@@ -192,6 +192,14 @@ namespace nmos
         return{ sender.at(nmos::fields::label).as_string(), params, 100, media_stream_ids, details::make_ts_refclk(node, source, sender) };
     }
 
+    static sdp_parameters make_mux_sdp_parameters(const web::json::value& node, const web::json::value& source, const web::json::value& flow, const web::json::value& sender, const std::vector<utility::string_t>& media_stream_ids)
+    {
+        sdp_parameters::mux_t params;
+        params.tp = sdp::type_parameters::type_NL;
+
+        return{ sender.at(nmos::fields::label).as_string(), params, 98, media_stream_ids, details::make_ts_refclk(node, source, sender) };
+    }
+
     sdp_parameters make_sdp_parameters(const web::json::value& node, const web::json::value& source, const web::json::value& flow, const web::json::value& sender, const std::vector<utility::string_t>& media_stream_ids)
     {
         const auto& format = nmos::fields::format(flow);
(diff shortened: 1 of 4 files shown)


Development/nmos/sdp_utils.cpp
```diff
@@ -192,6 +192,14 @@ namespace nmos
         return{ sender.at(nmos::fields::label).as_string(), params, 100, media_stream_ids, details::make_ts_refclk(node, source, sender) };
     }
 
+    static sdp_parameters make_mux_sdp_parameters(const web::json::value& node, const web::json::value& source, const web::json::value& flow, const web::json::value& sender, const std::vector<utility::string_t>& media_stream_ids)
+    {
+        sdp_parameters::mux_t params;
+        params.tp = sdp::type_parameters::type_NL;
```
garethsb-sony: Why default to Narrow Linear here?

garethsb-sony:
Hmm, checking ST 2022-8:2019 myself, I don't see it mentions using type parameter (TP) at all?

Development/nmos/sdp_utils.cpp
grimmn: ST2022-8 Section 6 specifies that

> Senders shall comply with either the "Narrow Linear Senders (Type NL)" requirements, or the "Wide Senders (Type W)" requirements.

grimmn:
The TP requirement is in Section 6:

> Senders shall signal Media Type Parameters TP and TROFF as specified in ST 2110-21.

Slightly annoying that there are sessions description requirements listed outside of the section of the document titled Session Description Requirements.

Development/nmos/sdp_utils.cpp
garethsb-sony: Ah! How did I miss that? More haste, less speed reading. Sorry! Makes sense. Only Gapped is irrelevant.

garethsb-sony:
> The TP requirement is in Section 6:

Yep, and should presumably have been in Section 8 also.

> > Senders shall signal Media Type Parameters TP and TROFF as specified in ST 2110-21.

...which we take to mean "Senders shall do what ST 2110-21 says about these two parameters, i.e. TP is required and TROFF is optional."

grimmn:
Exactly, that is how I interpreted it as well.

garethsb-sony:
OK, I'm going to merge this to master. Thanks for the work on this, Jonathan.

Given that ST 2022-8 says (and similarly in ST 2110-21 actually), "Video RTP Senders whose streams utilize a value of TR-OFFSET which differs from TRO-DEFAULT shall signal the prevailing value" should we demo how to add it to the generated format specific parameters after the fact, or should we just add support for that attribute too in both `nmos::sdp_parameters::video` and `mux`?

PR #117, opened 2020-01-31
Node API registration post order

wsneijers:
Order of posting resources to a registration API should happen on resource type, not on creation date.

### DIFF
--- a/Development/nmos/query_utils.cpp
+++ b/Development/nmos/query_utils.cpp
@@ -317,16 +317,20 @@ namespace nmos
 
         std::vector<web::json::value> events;
 
-        // resources are traversed in order of increasing creation timestamp, so that events for super-resources are inserted before events for sub-resources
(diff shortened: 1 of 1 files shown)


garethsb-sony:
Yes... the [IS-04 v1.3 documentation on referential integrity](https://amwa-tv.github.io/nmos-discovery-registration/branches/v1.3.x/docs/4.1._Behaviour_-_Registration.html#referential-integrity) says:

"Typically registration should proceed in the following order: Node, Devices, Sources, Flows, Senders, Receivers"

In the circumstances where an nmos-cpp Node creates a Sender with `"flow_id": null` initially, and then creates the Flow and adds the id to the Sender, the current implementation will result in the NMOS Testing Tool warning about the non-recommended ordering...

IS0401Test-test_10_01-warning

garethsb-sony:
Thanks! I've investigated the effect of this in a number of use cases and it looks OK.

PR #123, opened 2020-02-22
Update is04 json schemas

prince-chrismc:
A few minor fixes were missing.


garethsb-sony:
As per the naming of the directories in nmos-cpp, they were supposed to reflect released spec versions and the changes you've included haven't made it into a spec patch release yet. But it is awkward as I agree it's better to be on the bleeding edge of the ".x" branches usually, just makes it hard to track. Hmm

garethsb-sony:
Master is now sync'd with the spec ".x" branches not the release tags. Thanks.

prince-chrismc:
Thank you very much, I was not aware of your convention.

PR #152, opened 2020-06-09
Add source map extension to the admin UI HTTP server

DouglasHeriot:
When using sony/nmos-js this allows viewing the Javascript source maps in the web inspector.

Mime-type application/json seems like a good choice, but doesn't actually matter:
https://stackoverflow.com/a/19912684/466698


garethsb-sony:
Seems like a good idea. If `application/json` is a good choice, why pick `application/javascript`?

DouglasHeriot:
Oops, you're right. That was a poor copy/paste mistake.
Fixed. Just tested it, and it works correctly. (Will be easier to learn how the nmos-js project works now!)

garethsb-sony:
Thanks for this.

PR #155, opened 2020-06-11
disable testMdnsResolveAPIs on mac

alanb-sony:
run disabled tests and ignore their result


garethsb-sony:
Thanks, looks good. We'll have to remember to check into this and any other currently 'expected' failures...

PR #158, opened 2020-08-13
Fix build with boost 1.73.0

connorfeeley:
### Issue:
Building with boost 1.73.0 causes the following compiler error:

> nmos-cpp/Development/nmos/connection_api.cpp: In function ‘const web::json::experimental::json_validator& nmos::details::staged_core_validator()’:
> nmos-cpp/Development/nmos/connection_api.cpp:103:149: error: ‘_1’ was not declared in this scope
>   103 |                     is05_versions::all | boost::adaptors::transformed(boost::bind(experimental::make_connectionapi_staged_patch_request_schema_uri, _1, nmos::types::sender)),
> 

This can be reproduced by specifying boost 1.73.0 in `conanfile.txt`

As of boost 1.73.0, the boost placeholders (_1, _2, etc) are no longer imported into the global namespace by default.

[boost/bind.hpp](https://www.boost.org/doc/libs/1_73_0/boost/bind.hpp) now has a comment stating:

>   For backward compatibility, this header includes
>   <boost/bind/bind.hpp> and then imports the placeholders _1, _2,
>   _3, ... into the global namespace. Definitions in the global
>   namespace are not a good practice and this use is deprecated.
>   Please switch to including <boost/bind/bind.hpp> directly,
>   adding the using directive locally where appropriate.
>   Alternatively, the existing behavior may be preserved by defining
>   the macro BOOST_BIND_GLOBAL_PLACEHOLDERS.
> 

### Solution:
To build with boost 1.73.0, the `boost/bind/bind.hpp` header should be included and the `boost::placeholders` namespace should be used.

There are three cases that need to be considered:
- **Boost version < 1.60.0:** placeholders defined directly in the global namespace
- **Boost version >= 1.73.0:** placeholders defined in `boost::placeholders`, unless `BOOST_BIND_GLOBAL_PLACEHOLDERS` is defined.
- **Boost version >= 1.60.0 and < 1.73.0:** placeholders defined in `boost::placeholders`, but imported into the global namespace

Because the `boost::placeholders` namespace is defined for versions 1.60.0 and up, this can be reduced to a single version check.

To maintain backwards compatibility with old Boost versions, a new namespace (`bst::boost_placeholders`) has been introduced which is aliased to `boost::placeholders` if the Boost version is >= 1.60.0, and aliased to the placeholders in the global namespace if the Boost version is < 1.60.0.

`bst::placeholders` would be a better name than `bst::boost_placeholders`, but `slog/all_in_one.h` already aliases it to `std::placeholders`, which isn't compatible with `boost::bind`.

### DIFF
--- a/Development/bst/placeholders.h
+++ b/Development/bst/placeholders.h
@@ -1,0 +1,51 @@
+#ifndef BST_PLACEHOLDERS_H
+#define BST_PLACEHOLDERS_H
+
+// Provide bst::placeholders using the boost::placeholders namespace (to be compatible with Boost 1.73.0 and up),
+// or the global namespace (to be compatible with Boost versions less than 1.60.0). Boost versions between 1.60.0 and 1.73.0
+// have the boost::placeholders namespace defined, and import it into the global namespace by default. Versions less than
+// 1.60.0 don't have the boost::placeholders namespace defined at all, and simply define the placeholders in the global
+// namespace.
+// As of Boost version 1.73.0, the boost::placeholders namespace is no longer globally imported by default.
+
+// From boost/bind.hpp in Boost 1.73.0:
+//   For backward compatibility, this header includes
+//  <boost/bind/bind.hpp> and then imports the placeholders _1, _2,
+//  _3, ... into the global namespace. Definitions in the global
+//  namespace are not a good practice and this use is deprecated.
+//  Please switch to including <boost/bind/bind.hpp> directly,
+//  adding the using directive locally where appropriate.
+//  Alternatively, the existing behavior may be preserved by defining
+//  the macro BOOST_BIND_GLOBAL_PLACEHOLDERS.
+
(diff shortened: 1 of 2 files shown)


garethsb-sony:
Thanks for the thorough work on this. I think the best thing to do would be to use `bst::bind` and `bst::placeholders::_1` in the code in **nmos/connection_api.cpp** and fix **bst/function.h** (that is amalgamated into **slog/all_in_one.h**) with the check for the three cases for different Boost versions that you describe. I'll get to this shortly.

garethsb-sony:
Hopefully what I've just pushed to master solves this issue. Thanks again for finding this.

garethsb-sony:
It wasn't quite that straightforward - Boost 1.54.0 vintage `boost::adaptors::transformed` doesn't work with `std::bind`. I've pushed a more pragmatic fix that works around that. The **windows-latest** build is failing for an entirely unrelated reason.

PR #163, opened 2020-09-03
Bcp 003 tests

alanb-sony:


.github/workflows/src/amwa-test.yml
```diff
@@ -71,6 +71,11 @@
         find . -iname "*.py" -exec sed -i 's#\["testssl/testssl.sh#\["'"${bash_path}"'", "testssl/testssl.sh#g' {} +
     fi
 
+    if [[ "${{ runner.os }}" == "Linux" && "${{ matrix.install_mdns }}" == "false" ]]; then
+      # nmos-cpp-node doesn't currently support advertising hostnames to Avahi
+      avahi-publish-address nmos-api.local -a -R ${{ env.HOST_IP_ADDRESS }} &
```
garethsb-sony, line 76:
```suggestion
      avahi-publish -a -R nmos-api -d local ${{ env.HOST_IP_ADDRESS }} &
```

garethsb-sony:
Hmm, going back to the default `HTTP_TIMEOUT` means the Avahi build fails **IS-04-01 test_13**. Increasing it to pass that one test case on one build config is a pretty big sledgehammer. :-/

garethsb-sony:
It might be interesting to remove the HTTPS timing workaround in `IS0401Test.do_registry_basics_prereqs` and look what happens as I wonder if the cause is related.

https://github.com/AMWA-TV/nmos-testing/blob/97f3b452b8ba08588e4539ae702fbf589afe7378/nmostesting/suites/IS0401Test.py#L213-L220
```python
                # in event of testing HTTPS support, the TLS handshake seems to take nearly 2 seconds, so
                # when the first registry is disabled, an additional few seconds is needed to ensure the node
                # has a chance to make a connection to it, receive the 5xx error, and make a connection to
                # the next one
                if CONFIG.ENABLE_HTTPS:
                    heartbeat_countdown = CONFIG.HEARTBEAT_INTERVAL + 1 + 5
                else:
                    heartbeat_countdown = CONFIG.HEARTBEAT_INTERVAL + 1
```

alanbirtles:
yep, the `HTTP_TIMEOUT` workaround was for the avahi build originally

garethsb-sony:
I noticed **IS-04-01 test_13** has started failing for ubuntu-14.04 (non-Conan) on the last few commits with:

```
            "detail": "Receiver target PUT did not produce a 202 response code: 400",
```

Haven't yet checked if I failed to merge something that should have been into build-test from the src files. I assumed it initially stopped working when the mocks.testsuite.nmos.tv name was removed from hosts, but the latest build is using the latest test suite commit which should have nmos-mocks.local in the URL...

FWIW, I also see in the testoutput:

```
 * WARNING: Could not find an installation of 'testssl'. Some tests will be disabled.
```

and `"Test Disabled"` in the BCP-003-01 json output (which we aren't classifying as a failure).

PR #165, opened 2020-11-09
Fix nmos-testing on macOS

alanb-sony:
Broken by a combination of:
https://github.com/conan-io/conan/issues/8031
https://github.com/lepture/authlib/issues/290


garethsb-sony:
Thank you for tracking that down, @alanb-sony!!

PR #166, opened 2020-11-12
revert workaround for nmos-testing on mac

alanb-sony:
should no longer be necessary with conan 1.31.1


garethsb-sony:
Thanks, Alan - and I see they already have conan 1.31.2 to fix a similar issue with urllib!

PR #167, opened 2020-12-23
Change to get a bigger buffer for GetAdapterAddresses when original is not big enough

jwpwh:
The call to GetAdapterAddresses starts out with a buffer of 15000 bytes. If this buffer is not big enough, the call to GetAdapterAddresses returns with error code ERROR_BUFFER_OVERFLOW and returns the size of how big the buffer should be. This change looks for that error code, allocates a bigger buffer and calls GetAdapterAddresses again in order to avoid the error.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -158,7 +158,6 @@ jobs:
         echo "DOCKER_HOST=$DOCKER_HOST" >> $GITHUB_ENV
         echo "DOCKER_CERT_PATH=$DOCKER_CERT_PATH" >> $GITHUB_ENV
     
-    
     - name: ubuntu setup
       if: runner.os == 'Linux'
       run: |
@@ -168,6 +167,8 @@ jobs:
         ip address
         # add the CRL Distribution Point to hosts so that it's discoverable when runing the AMWA test suite in mDNS mode
         echo "$hostip crl.testsuite.nmos.tv" | sudo tee -a /etc/hosts
+        # re-synchronize the package index
+        sudo apt-get update -q
     
     - name: ubuntu mdns install
       if: runner.os == 'Linux' && matrix.install_mdns == true
@@ -519,7 +520,6 @@ jobs:
         echo "DOCKER_HOST=$DOCKER_HOST" >> $GITHUB_ENV
         echo "DOCKER_CERT_PATH=$DOCKER_CERT_PATH" >> $GITHUB_ENV
     
-    
     - name: ubuntu setup
       if: runner.os == 'Linux'
       run: |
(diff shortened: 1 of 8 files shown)


jwpwh:
For what it is worth: this branch has 1 change in 1 file in WIN32 specific code. I don't really see any way for this to break ubuntu-18.04 builds.

On the other hand, I do like to learn at least one new thing a day and maybe this is an opportunity.

garethsb-sony:
Looks like installation of nscd failed, because the file or went for in http://azure.archive.ubuntu.com/ubuntu/pool/universe/g/glibc/ doesn't exist any more, bit odd. Certainly nothing to do with the PR anyway!

When have you encountered the ERROR_BUFFER_OVERFLOW?

I didn't like a 15kB std::array but it did have the advantage over std::vector of not unnecessarily zeroing the contents. Hmm.

garethsb-sony:
> I didn't like a 15kB std::array but it did have the advantage over std::vector of not unnecessarily zeroing the contents. Hmm.

The verbose solution to this would appear to be [a special allocator that does `default-init`](https://stackoverflow.com/questions/21028299/is-this-behavior-of-vectorresizesize-type-n-under-c11-and-boost-container/21028912#comment85778704_21028912).

The [GetAdaptersAddresses docs and example code](https://docs.microsoft.com/en-us/windows/win32/api/iphlpapi/nf-iphlpapi-getadaptersaddresses) also imply that since the required buffer size can change between calls, one may have to retry several times. One retry is obviously better than none, but I wonder if there are circumstances (boot time?) where this is more likely to be encountered?

jwpwh:
On my PC, the needed size returned was about 32K instead of 15K. I have 11 items listed in the output of ipconfig. there are at least 3 ethernet adapters in use and a wifi. There are also a usb ethernet being used as a license dongle, a bluetooth network connection, a VMware port, and a couple of Wireless LAN adapters. 

I think you are right that when booting up, the system may be changing with respect to the status of the ethernet adapters. This speaks to a need to do multiple tries at GetAdapterAddresses for two different reasons. The size needed may be changing and the failure to find the desired adapter may just mean the adapter is not up yet.

The changing size seems easy to deal with. Multiple retries is one way. Another is to allow the specification of the "known" size needed so it just starts at a good place. For an embedded system, the configuration should be pretty static and a known size could be specified. My real application will be Linux based so the PC is just for development. Doing multiple retries, even with the time to initialize the space to 0 is not a problem. It just has to work. 

Not sure how to deal with being too early for the desired adapter. Probably have to delay the application or order the startup so that this is unlikely to happen. On the PC, this is not a problem for me. I am only starting the app after the system is fully initialized. For development, the app is not being started as a service. We do have other applications which are services and the volatility of the network adapters has been an issue, so this seems likes a valid concern in the general case.

garethsb-sony:
I'm going to merge this, thank you, John.

> Not sure how to deal with being too early for the desired adapter. Probably have to delay the application or order the startup so that this is unlikely to happen. On the PC, this is not a problem for me. I am only starting the app after the system is fully initialized. For development, the app is not being started as a service. We do have other applications which are services and the volatility of the network adapters has been an issue, so this seems likes a valid concern in the general case.

I think you're right, but I think this part is outside the scope of the change here. Running as a service would need other changes vs. the structure of the example 'node implementation' code.

Issue #177, opened 2021-03-31
Support for ST 2110-22 / JPEG XS?


garethsb:
Most of the NMOS functionality to support description of coded video is already provided by nmos-cpp, but some convenience functions lack support. E.g. `nmos::make_coded_video_flow` is already provided by the nmos module, but `nmos::make_sdp_parameters`, which is used in the example nmos-cpp-node to construct an SDP file for the Sender based on the Source, Flow and Sender resources, doesn't work because it assumes a raw video Flow with `"components"` and therefore throws a `json_exception`.

It would be great to fill out the missing support for JPEG-XS in ST 2110-22 at least as constrained by forthcoming VSF TR-08.

And provide better error messages for unsupported media types.

Thanks to @David-P-B for the report.

garethsb:
The I-D for the RTP payload mapping for JPEG XS is here: https://datatracker.ietf.org/doc/draft-ietf-payload-rtp-jpegxs/
The SDP format-specific parameters are covered in the latest draft: https://tools.ietf.org/html/draft-ietf-payload-rtp-jpegxs-09#section-6

garethsb:
Version 18 looks pretty close to final...

jonvmey:
IANA has a [video/jxsv](https://www.iana.org/assignments/media-types/video/jxsv) media type registration listed now (referencing draft version 18) which is the actual requirement for ST2110-22.

garethsb:
[RFC 9134](https://tools.ietf.org/html/rfc9134.html) is now published, and we have a Work-In-Progress draft of [AMWA BCP-006-01: NMOS With JPEG XS](https://specs.amwa.tv/bcp-006-01). Comments on that are welcome here: https://github.com/AMWA-TV/nmos-jpeg-xs.

PR #179, opened 2021-04-08
Preparatory work for supporting ST 2110-22 and JPEG XS

garethsb:
As per #177, some of the SDP-related convenience/utility functions like `nmos::make_video_sdp_parameters` and `nmos::get_session_description_sdp_parameters` (and the `sdp_parameters` struct itself probably!) need improving to handle `video/jxsv`, if not other coded video formats. This doesn't try to do any of any that, but adds some of the constants that will be required.

### DIFF
--- a/Development/nmos/media_type.h
+++ b/Development/nmos/media_type.h
@@ -16,7 +16,13 @@ namespace nmos
     {
         // Video media types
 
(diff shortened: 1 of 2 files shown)


jonathan-r-thorpe (approved):
LGTM

PR #181, opened 2021-04-22
Fix SEGFAULT while running testWebSocketListenerCloseOpen on macOS

lo-simon:
This could be due to `Reference cannot be bound to dereferenced null pointer` while executing `&server.get_io_service()` for determining whether to initialise or restart the io_service. This can be done, using a `boolean flag` to identify the status of the io_service instead of calling `&server.get_io_service()`


garethsb:
Ah, Apple say:
> If the compiler spots a pointer dereference, it treats that pointer as non-null. As a result, null equality checks on pointers that are guaranteed to be dereferenced may be optimized away.

https://developer.apple.com/documentation/xcode/diagnosing_memory_thread_and_crash_issues_early/null_reference_creation_and_null_pointer_dereference

jonathan-r-thorpe (approved):
LGTM

PR #186, opened 2021-06-07
Little unit test tweak

garethsb:
Little change so that the test doesn't use post _response_ example data to test the post _request_ schema.

The _response_ data does in fact validate against the _request_ schema but the `id` and `ws_href` are supposed to be allocated by the Query API, not by the client. The test could also check the `POST` response or `GET` data against the other schema, but the important thing is whether a new `POST` request validates.

### DIFF
--- a/Development/nmos/test/query_api_test.cpp
+++ b/Development/nmos/test/query_api_test.cpp
@@ -20,7 +20,8 @@ BST_TEST_CASE(testQueryAPISubscriptionsExtensionSchema)
         boost::copy_range<std::vector<web::uri>>(nmos::is04_versions::all | boost::adaptors::transformed(nmos::experimental::make_queryapi_subscriptions_post_request_schema_uri))
     };
 
-    // valid subscriptions data
+    // valid subscriptions post request data
+    // see https://github.com/AMWA-TV/nmos-discovery-registration/blob/v1.3.x/examples/queryapi-subscriptions-post-request.json
     auto data = value_of({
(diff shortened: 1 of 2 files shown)


Development/nmos/test/query_api_test.cpp
```diff
@@ -20,27 +20,23 @@ BST_TEST_CASE(testQueryAPISubscriptionsExtensionSchema)
...
         { U("resource_path"), U("/nodes") },
         { U("params"), value_of({
             { U("label"), U("host1") }
         }) },
         { U("persist"), false },
-        { U("secure"), false },
-        { U("authorization"), false },
-        { U("ws_href"), U("ws://172.29.80.52:8870/ws/?uid=7c903667-7113-4a8f-8865-1c63f9070a9e") },
-        { U("id"), U("7c903667-7113-4a8f-8865-1c63f9070a9e") }
+        { U("secure"), false }
```
lo-simon, line 32: The `secure` field is not required by the schema, we can drop that too.

lo-simon (changes requested):
Yes you are right, should be using the post request data for the check

Development/nmos/test/query_api_test.cpp
garethsb: You could, but the example in the spec does include that property. 🤷‍♂️

jonathan-r-thorpe (approved):
LGTM

PR #188, opened 2021-06-28
Allow a preferred registration URL to be specified (node behaviour)

duncwebb:
Currently in the node behaviour it's only possible to provide a fallback registration URL to try in the event that mDNS discovery doesn't yield a registration endpoint to connect to. This change adds three new config keys for hostname (or IP address), port and API version (the latter two have sensible defaults) to request that the node attempt to connect to a preferred registration endpoint ahead of the list generated by the discovery process. This allows for a specific registration endpoint to be used that isn't advertised by DNS-SD or one regardless of the PRI value in the mDNS record. If the preferred URL is unavailable, it'll be popped off the list must like mDNS discovered entries. If the preferred hostname is not specified, behaviour will be as it is now.

### DIFF
--- a/Development/nmos/node_behaviour.cpp
+++ b/Development/nmos/node_behaviour.cpp
@@ -227,9 +227,14 @@ namespace nmos
 
         // query DNS Service Discovery for any Registration API in the specified browse domain, having priority in the specified range
         // otherwise, after timeout or cancellation, returning the fallback registration service
-        web::json::value discover_registration_services(mdns::service_discovery& discovery, const std::string& browse_domain, const std::set<nmos::api_version>& versions, const std::pair<nmos::service_priority, nmos::service_priority>& priorities, const std::set<nmos::service_protocol>& protocols, const std::set<bool>& authorization, const web::uri& fallback_registration_service, slog::base_gate& gate, const std::chrono::steady_clock::duration& timeout, const pplx::cancellation_token& token = pplx::cancellation_token::none())
+        web::json::value discover_registration_services(mdns::service_discovery& discovery, const std::string& browse_domain, const std::set<nmos::api_version>& versions, const std::pair<nmos::service_priority, nmos::service_priority>& priorities, const std::set<nmos::service_protocol>& protocols, const std::set<bool>& authorization, const web::uri& fallback_registration_service, const web::uri& preferred_registration_service, slog::base_gate& gate, const std::chrono::steady_clock::duration& timeout, const pplx::cancellation_token& token = pplx::cancellation_token::none())
         {
             std::list<web::uri> registration_services;
+
+            if (!preferred_registration_service.is_empty())
+            {
+                registration_services.push_back(preferred_registration_service);
+            }
 
             if (nmos::service_priorities::no_priority != priorities.first)
             {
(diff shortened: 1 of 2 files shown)


lo-simon:
Thank you for your PR, have you thought about using the fallback registry settings and the `highest_pri` and `lowest_pri` settings to achieve the same outcome as using your suggested preferred registry settings.

Example 1, in a DNS system, you have registries A, B, and C.  Let's say A is your preferred registry has `pri` set to 100.  B and C have their `pri` set to 0. Then configuring your nmos-cpp-node's `highest_pri` and `lowest_pri` set to 100.  nmos-cpp-node will filter out any registries which are outside the `highest_pri` and `lowest_pri` range, i.e. only registry A will be discovered.

Example 2, similar to example 1 but this time registry A is not advertised via the DNS server. Then configuring your nmos-cpp-node's `highest_pri` and `lowest_pri` set to 100 and `fallback registry` set to registry A.  nmos-cpp-node will not discovered any suitable registry, so the fallback registry will be selected.

duncwebb:
Hi and thanks for replying. I had considered this but unfortunately the use case that this cannot solve is when I need to select a specific registry from a set which all have the same PRI values and the configuration of the registries is outside of my control.

This is even more pronounced on a system where there are registries advertised and others not advertised with the same PRI. I need the flexibility to be able to select a specific non-advertised registry (which would be impossible here as the advertised ones would always be selected first). 

These use cases all derive from the fact that the application under development is to be transported between many different environments where the configuration of nodes and registries is constantly being changed by engineers so I need to be able to communicate with a specific registry by IP address & port.

garethsb:
Hi Duncan, those sound like messy network environments... Presumably these are dev environments rather then production? But devs are running Registries without agreeing a strategy for `pri`?

> I need the flexibility to be able to select a specific non-advertised registry (which would be impossible here as the advertised ones would always be selected first).

Set `lowest_`/`highest_pri` to an unused `pri` value, e.g. `9999`, so that they do not match any of the advertised Registries, and pass a fallback Registry, and it will get used, I believe.

lo-simon:
As Gareth suggested using 9999 (a value that is not used by any of your advertised registries) for the `lowest_pri`/`highest_pri` would definitely overcome your issue, which will do exactly the same as having the preferred registry settings.  By doing so no registries will be discovered, then your fallback registry will be selected.  

The fallback registry there is really for the non-production setup.  NMOS node should already be using DNS-SD as NMOS spec staged, having the preferred registry settings will taking away the main concept of using DNS-SD.  See https://specs.amwa.tv/is-04/releases/v1.1.3/docs/3.1._Discovery_-_Registered_Operation.html#registration

duncwebb:
Some of the environments are indeed 'in development' - often with NMOS equipment that's either in development or not 100% compliant with standards. Thanks for the suggestion, I'll spin up some typical environments that I've seen and will try the suggested approach. Thanks again Simon and you too Gareth for the help.

jonathan-r-thorpe:
Hi @duncwebb - did the above suggestions solve your problem as I'm looking to close this PR?

duncwebb:
Hi @jonathan-r-thorpe, I'm looking at this again at the moment. Initial examination looks like it will do the trick though there's an edge case I want to check will not cause any practical problems (if there is a registry with a PRI of intmax - the upper boundary - it's not possible to ignore it). I'll get back to you ASAP.

lo-simon:
Hi Duncan, Registry disables mDNS advertisement if its `pri` is set to intmax, so there is no edge case at all.  In the production environment, `pri` should only be set between 0-99, anything from 100 is for development purposes.

see Development/nmos-cpp-registry/config.json for `pri` description.

```
    // pri [registry, node]: used for the 'pri' TXT record; specifying nmos::service_priorities::no_priority (maximum value) disables advertisement completely
    //"pri": 100,
```

duncwebb:
Hi Simon - that's great. Our kit is used in production and test environments so it was a concern but that's great. OK let's close this one then. Thanks again, Duncan.

Issue #189, opened 2021-06-30
Segfault when a request is received and the server instance is deleted at the same time


AtsushiMorinaga:
It is required to re-create the nmos::server instance when listener port settings(IP address, port, etc) are changed.
Segmentation fault may happen if a request comes in the middle of tear down process.

As a result of investigation, I found that
1. The request handler has a lambda callback function that has capture of `api_router` by reference.
  https://github.com/sony/nmos-cpp/blob/master/Development/nmos/api_utils.cpp#L666
2. When a request is received, the handler is invoked and an async task is generated.
  https://github.com/sony/nmos-cpp/blob/master/Development/cpprest/api_router.cpp#L66
3. Then the instance of nmos::server is deleted, resources are destructed but the async task is not cannceled by destructor.
4. Segfault when the task is invoked because it has the dangling reference of `api_router`.

* Backtrace

```
#0  0x004823d4 in bool std::__detail::__regex_algo_impl<__gnu_cxx::__normal_iterator<char const*, std::string>, std::allocator<std::sub_match<__gnu_cxx::__normal_iterator<char const*, std::string> > >, char, std::regex_traits<char>, (std::__detail::_RegexExecutorPolicy)0, true>(__gnu_cxx::__normal_iterator<char const*, std::string>, __gnu_cxx::__normal_iterator<char const*, std::string>, std::match_results<__gnu_cxx::__normal_iterator<char const*, std::string>, std::allocator<std::sub_match<__gnu_cxx::__normal_iterator<char const*, std::string> > > >&, std::basic_regex<char, std::regex_traits<char> > const&, std::regex_constants::match_flag_type) ()
#1  0x0047c7dc in web::http::experimental::listener::api_router::operator()(web::http::http_request, web::http::http_response, std::string const&, std::unordered_map<std::string, std::string, std::hash<std::string>, std::equal_to<std::string>, std::allocator<std::pair<std::string const, std::string> > > const&, std::_List_iterator<web::http::experimental::listener::api_router::route>) ()
[...]

PR #190, opened 2021-07-05
Terminate runing request before closing the server to fix issue #189

lo-simon:
Ensure the api router is completed/terminated before returning to the http_listener request handler.  This will prevent the API routers to be deleted before the http_listeners while deleting the server instance.

### DIFF
--- a/Development/cpprest/api_router.cpp
+++ b/Development/cpprest/api_router.cpp
@@ -31,14 +31,14 @@ namespace web
                     }
                 }
 
-                api_router::api_router()
+                api_router::api_router() : canceled(false)
                 {}
 
                 void api_router::operator()(web::http::http_request req)
                 {
                     web::http::http_response res;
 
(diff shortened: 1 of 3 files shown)


lo-simon:
@garethsb please review

garethsb:
Is it a good idea to make the `api_router::operator()(web::http::http_request req)` function blocking? I think this is likely to cause thread pool starvation and deadlock under load. Basically _any_ blocking `wait` in `http_listener` handler code is an anti-pattern, isn't it?

garethsb:
Obviously I ought to have added unit tests to show the problem that was solved by 345e344d99710817e1269b7396dbb46f3b1ce5ac, and equally this needs unit tests to show the problem identified by #189.

garethsb:
My suggestion is to bind the `api_router` into the handlers by value rather than reference, and to give `api_router` a shared pimpl. And keep the `server` code unchanged.

jonathan-r-thorpe:
PR  superseded by #191

PR #201, opened 2021-10-01
SDP fixes

### LINKED ISSUE #125: Handling of PTP reference clocks.
NMOS IS-04 requires that the GMID of the PTP reference clock be given with lowercase hex number notation. The corresponding SDP attribute `a=ts-refclk` has it the other way around, i.e. uppercase hex only. Somewhere a conversion must happen either way, but I don't see where.

Furthermore, nmos-cpp seems to assume that the `a=ts-refclk` attribute is always in the media section of the SDP file. Indeed that is what ST2110 (stupidly) demands. Other relevant standards don't have this restriction, for example AES67, and indeed the RFC 7273 they all depend on. One would have thought that the normal case is a common reference clock, so that putting it into the global section of the SDP file would be the most prevalent case, if it weren't for ST2110.

So I think nmos-cpp should at least recognise and properly handle either style when receiving an SDP. At the moment, it insists on ST2110 on input, too.

Please tell if you want me to produce a patch.

garethsb:
1. Fix correct setting of `source_ip` in a Receiver's `transport_params` when parsing an SDP file which doesn't contain a `source-filter`. Previously it set it from the origin `<unicast-address>` but that's wrong - as shown by the example in the IS-05 spec for [Behaviour - RTP Transport Type - Unicast](https://specs.amwa.tv/is-05/releases/v1.1.1/docs/4.1._Behaviour_-_RTP_Transport_Type.html#unicast).

2. Following on from that, generate `source-filter`s from unicast Senders' `transport_params` by default.

3. To enable creation of the effective SDP files at a Receiver when `rtp_enabled` is explicitly overridden, support creation of SDP files from Receivers' `transport_params`. Previously only worked from Senders' `transport_params`, which have a slightly different set of parameters, e.g. `destination_ip` vs. `multicast_ip`/`interface_ip`.

4. New test cases are added for most of the above improvements.

5. Also support "session-level" `ts-refclk` and `mediaclk` which will resolve #125. The `source-filter` attribute is still not handled at  "session-level", but I've added more comments to explain some of the ways the `source-filter` handling could be improved (e.g. checking `<dest-address>` actually matches the connection address being considered).

6. Fix generation of `a=mediaclk:sender` (which would previously have been rendered with a trailing `=`).

### DIFF
--- a/Development/bst/filesystem.h
+++ b/Development/bst/filesystem.h
@@ -82,6 +82,7 @@ namespace bst
         using bst_filesystem::file_size;
         using bst_filesystem::create_directory;
         using bst_filesystem::remove_all;
+        using bst_filesystem::temp_directory_path;
     }
 }
 
--- a/Development/cmake/NmosCppTest.cmake
+++ b/Development/cmake/NmosCppTest.cmake
@@ -46,6 +46,7 @@ set(NMOS_CPP_TEST_NMOS_TEST_SOURCES
     nmos/test/json_validator_test.cpp
     nmos/test/paging_utils_test.cpp
     nmos/test/query_api_test.cpp
+    nmos/test/sdp_utils_test.cpp
     nmos/test/system_resources_test.cpp
     )
 set(NMOS_CPP_TEST_NMOS_TEST_HEADERS
--- a/Development/nmos/sdp_utils.cpp
+++ b/Development/nmos/sdp_utils.cpp
@@ -243,11 +243,54 @@ namespace nmos
         return make_sdp_parameters(node, source, flow, sender, media_stream_ids, bst::nullopt);
     }
 
-    static web::json::value make_session_description(const sdp_parameters& sdp_params, const web::json::value& transport_params, const web::json::value& ptime, const web::json::value& rtpmap, const web::json::value& fmtp)
+    namespace details
+    {
+        // "<unicast-address> is the address of the machine from which the session was created."
+        // See https://tools.ietf.org/html/rfc4566#section-5.2
+        const utility::string_t& get_origin_address(const web::json::value& transport_params)
+        {
+            // this doesn't need to be the source address of the data stream
+            // e.g. a host_address which was a management interface would be OK
+            // however, by convention, use the source_ip of the first leg for
+            // an SDP file created at a Sender and the interface_ip of the
+            // first leg for an SDP created at a Receiver
+            const auto& transport_param = transport_params.at(0);
+            const auto& interface_ip = nmos::fields::interface_ip(transport_param);
+            // a Receiver always has an interface_ip
+            if (!interface_ip.is_null()) return interface_ip.as_string();
+            const auto& source_ip = nmos::fields::source_ip(transport_param);
+            // a Sender always has a source_ip
+            return source_ip.as_string();
+        }
+
+        // "If the session is multicast, the connection address will be an IP multicast group address.
+        // If the session is not multicast, then the connection address contains the unicast IP address of the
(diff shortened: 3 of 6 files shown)


garethsb:
The `nmos::get_session_description_transport_params` function which is used when parsing SDP files for IS-05 includes this comment:
```C++
        // For now, this function should handle the following cases identified in the documentation:
        // * Unicast
        // * Source Specific Multicast
        // * Any Source Multicast
        // * Operation with SMPTE 2022-7 - Separate Source Addresses
        // * Operation with SMPTE 2022-7 - Separate Destination Addresses

        // The following cases are not yet handled:
        // * Operation with SMPTE 2022-5
        // * Operation with SMPTE 2022-7 - Temporal Redundancy
        // * Operation with RTCP
```

I think this is basically true, however it's worth pointing out that not all of the _other_ relevant information is extracted from the SDP file.

For example, when parsing an SDP file for ST 2022-7 with "Separate Destination Addresses", the `<payload type>` is only extracted into the `sdp_parameters` from the `m=`/`a=rtpmap:` lines of the first media description. Similarly, the `ssrc` attributes are not extracted when parsing an SDP File for ST 2022-7 with "Separate Source Addresses". This would prevent accurate roundtrip - creation of a new SDP file from the parsed `transport_params` and `sdp_parameters`.

On the creation side, `nmos::make_session_description` does not distinguish the different ST 2022-7 cases, so I'm pretty certain the "Separate Source Addresses" case does not generate the correct SDP file.

garethsb:
Note that ST 2110-10 has this to say about representation of ST 2022-7 in SDP:

> **8.3 SDP for Duplicate RTP Streams**
> Duplicate RTP streams meeting the requirements of SMPTE ST 2022-7 may be used for redundant transmission to achieve higher system availability. Senders which transmit these duplicated RTP streams, using the mechanisms of separate source addresses (IETF RFC 7104, Section 4.1) or separate destination addresses (IETF RFC 7104, section 4.2) shall signal the RTP duplication using the session level `group` attribute of IETF RFC 5888 and the duplication grouping `DUP` semantics of IETF RFC 7104.

The example for "Separate Source Addresses" in IS-05 is taken from RFC 7104 Section 4.1 and it uses `ssrc` and `ssrc-group`attributes for synchronization source (SSRC) multiplexed RTP streams. That mechanism, which is defined by [RFC 5576](https://datatracker.ietf.org/doc/html/rfc5576), isn't mentioned in ST 2110-10, which may explain why I've never seen it in the wild...

The ST 2110-10 **Annex B SDP Example (Informative)** is of the "Separate Destination Addresses" case, and uses the same RTP Payload ID for both streams, which may explain why I've never seen different ones outside of the example in RFC 7104 Section 4.2 which is reproduced in IS-05.

garethsb:
> 2. Following on from that, generate `source-filter`s from unicast Senders' `transport_params` by default.

The default behaviour at the Sender side is now that `source-filter` is written for both unicast (new) and multicast (as before).
I wonder whether this should be tightened up to only include the `source-filter` for multicast if the address is in the source-specific multicast block, `232/8` (`232.0.0.0`-`232.255.255.255`) defined by [RFC 5771](https://datatracker.ietf.org/doc/html/rfc5771#section-8) and registered in the IANA [IPv4 Multicast Address Space Registry](https://www.iana.org/assignments/multicast-addresses/multicast-addresses.xhtml#multicast-addresses-10). The new `source_filters` optional flag could be used to force it off in that range or on in other ranges.

lo-simon (approved):
LGTM

PR #204, opened 2021-10-15
Tweaks to nmos-cpp-node example

garethsb:
Add config settings for `frame_width`, `frame_height`, `interlace_mode` and whether or not to `activate_senders` on start up.

Tweak how the primary and secondary interfaces for RTP senders and receivers are identified in order to get reproducible results using the `host_addresses` front/back rather than the front/back of the filtered interfaces.

### DIFF
--- a/Development/nmos-cpp-node/config.json
+++ b/Development/nmos-cpp-node/config.json
@@ -20,9 +20,7 @@
     //"frame_height": 2160,
 
     // interlace_mode: controls the interlace_mode of video flows, see nmos::interlace_mode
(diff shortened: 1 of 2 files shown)


lo-simon:
> One other thing to note is that the default for impl::fields::interlace_mode is currently fixed to interlace_tff which means that if you switch frame_rate to e.g. 50 Hz but forget to change interlace_mode you'll get a surprising result... Perhaps worth making the default for interlace_mode depend on frame_rate and frame_width, frame_height, or making the default progressive, though that's then a change from previous fixed behaviour.

As this is just for the example, maybe just put the comment on it, is good enough.

Issue #108, opened 2020-01-16
Support VS2019 16.3 and higher


aholzinger:
VS2019 since 16.3 has deprecated std::experimental::filesystem with #error in the corresponding header making the compikation fail.

With this small patch it's working:
```
--- a/nmos-cpp/Development/bst/filesystem.h	Thu Jan 16 14:07:34 2020
+++ b/nmos-cpp/Development/bst/filesystem.h	Thu Jan 16 14:07:36 2020
@@ -17,6 +17,9 @@
 
 #if _MSC_VER >= 1900
 #define BST_FILESYSTEM_STD_EXPERIMENTAL
+#if _MSC_VER >= 1920
+#define _SILENCE_EXPERIMENTAL_FILESYSTEM_DEPRECATION_WARNING
+#endif
 #else
 #define BST_FILESYSTEM_MICROSOFT_TR2
 #endif
```

By moving to C++17 instead filesystem would be in std::filesystem, but I guess VS2015 is the blocker for this change.

garethsb-sony:
Thanks, I'll investigate.

garethsb-sony:
It may be possible to use `std::filesystem` in this case, via `BST_FILESYSTEM_STD`, something like this:

```patch
--- a/nmos-cpp/Development/bst/filesystem.h
+++ b/nmos-cpp/Development/bst/filesystem.h
@@ -17,6 +17,8 @@
 
+#if _MSC_VER >= 1920
+#define BST_FILESYSTEM_STD
+#elif _MSC_VER >= 1900
-#if _MSC_VER >= 1900
 #define BST_FILESYSTEM_STD_EXPERIMENTAL
 #else
 #define BST_FILESYSTEM_MICROSOFT_TR2
 #endif
```

aholzinger:
Unfortunately not. Only if you set the language standard to C++17 otherwise std::filesystem is not defined.

garethsb-sony:
OK, understood, thanks. I guess ultimately then we should support both modes, depending on value of `__cplusplus` preprocessor symbol, based on [/Zc:__cplusplus compiler option](https://docs.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=vs-2019).

aholzinger:
Yes, that would be the best solution, but I guess that would mean to support choosing C++17 (or also C++14?) for the whole project, wouldn't it?

[...]

PR #207, opened 2021-11-02
Enable Conan compiler.cppstd to override nmos-cpp's preferred CMAKE_CXX_STANDARD

### LINKED ISSUE #206: Conan recipe: selecting the appropriate C++ version flags
Hello,

Platform: Ubuntu 20.04 with gcc 10.3, cmake 3.21.3, conan 1.41.0.

We have an undefined symbol (make_session_description) error linking to nmos-cpp as a Conan dependency. 
We found that Conan was building the recipe using the default C++11 and we are using C++17. This causes nmos-cpp/Development/bst/optional.h to select boost::optional and we get this:

```
nm -C lib/libnmos-cpp.a | grep make_session_description
0000000000007d6d T nmos::make_session_description(nmos::sdp_parameters const&, web::json::value const&, boost::optional<bool>)
(...)
```

When building our code (with CXX_STANDARD=17), optional.h selects std::optional and the linker obviously fails.

I would expect that passing compiler.cppstd=17 on the Conan settings would set -std=c++17 when make runs gcc but it does not. I deleted one of the "*.o" files on the Conan build folder and `make -n` does show -std=c++11.

I will try to understand what is going on but, should someone have a clue, I'd appreciate any pointers.

Thanks,

Pedro

garethsb:
Resolves #206. Follow up to #108.

The change made in commit edd896214a772b10bbcfae89824bfeb7b22743a4 enabled the user to override nmos-cpp's preferred `CMAKE_CXX_STANDARD` on the CMake command line by setting that variable into the cache.

More recently, we introduced a [Conan recipe](https://github.com/conan-io/conan-center-index/tree/master/recipes/nmos-cpp), with [packages published at Conan Center Index](https://conan.io/center/nmos-cpp). @pedro-alves-ferreira noticed that specifying `compiler.cppstd=17` (as in [How to manage C++ standard](https://docs.conan.io/en/latest/howtos/manage_cpp_standard.html)) was ignored on the first build.

This turns out to be because Conan sets `CMAKE_CXX_STANDARD` as a "normal" variable, and on first run, the current CMake behaviour is that setting it as a cache variable removes the normal variable. Using idiomatic existence checks on all the `CMAKE_CXX_` variables instead is the right way to solve this issue.

This could be artificially demonstrated in GitHub Actions, but I haven't gone that far with this PR.
Once it's merged, there will still need to be a new Conan package version published.

### DIFF
--- a/Development/cmake/NmosCppCommon.cmake
+++ b/Development/cmake/NmosCppCommon.cmake
@@ -19,12 +19,18 @@ endif()
 
 # enable C++11
 enable_language(CXX)
-set(CMAKE_CXX_STANDARD 11 CACHE STRING "Default value for CXX_STANDARD property of targets")
+if(NOT DEFINED CMAKE_CXX_STANDARD)
+    set(CMAKE_CXX_STANDARD 11)
+endif()
 if(CMAKE_CXX_STANDARD STREQUAL "98")
     message(FATAL_ERROR "CMAKE_CXX_STANDARD must be 11 or higher; C++98 is not supported")
 endif()
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
-set(CMAKE_CXX_EXTENSIONS OFF)
+if(NOT DEFINED CMAKE_CXX_STANDARD_REQUIRED)
+    set(CMAKE_CXX_STANDARD_REQUIRED ON)
+endif()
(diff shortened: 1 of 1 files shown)


lo-simon (commented):
LGTM

PR #211, opened 2021-11-12
Add missing dependency on Boost

garethsb:
This didn't affect the Ubuntu 14.04 build in GitHub Actions only because Boost headers are found in the system include dir _/usr/include/_, but for other systems with GCC 4.8, and Boost installed in non-standard location, it is necessary.


lo-simon (commented):
Shouldn't the target_link_libraries be inside the if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS 4.9)?

garethsb:
Yes!!

garethsb:
Fixed, @lo-simon, thanks!

PR #215, opened 2021-11-30
Conan default package id mode considered harmful

garethsb:
The [build-test #297](https://github.com/sony/nmos-cpp/actions/runs/1367671562) logs show that bumping Boost from 1.76.0 to 1.77.0 caused nmos-cpp-node to terminate with a segmentation fault on Ubuntu 18.04 and 20.04.

This is because by default [Conan assumes semver compatibility](https://docs.conan.io/en/1.42/creating_packages/define_abi_compatibility.html#versioning-schema). Boost and other C++ libraries do not meet this expectation and break ABI compatibility between e.g. minor versions. The recipes in Conan Center Index do not generally customize their `package_id` method to take this into account.

Conan outputs this message in the log, but it's **highly** misleading!

```
WARN: cpprestsdk/2.10.18: requirement boost/1.76.0 overridden by your conanfile to boost/1.77.0
```

The cpprestsdk binary package used is still one built against Boost 1.76.0. When nmos-cpp(-node) is then built against Boost 1.77.0, this results in a seg fault.

A workaround is to change Conan's default package id mode to `minor_mode` or a stricter mode such as `recipe_revision_mode`.

```sh
conan config set general.default_package_id_mode=minor_mode
```

I wish there were something better we could do than put this in CI and documentation, but I haven't come up with anything.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -79,6 +79,8 @@ jobs:
       if: matrix.use_conan == true
       run: |
         pip install conan
+        # override default 'semver_direct_mode' due to boost
+        conan config set general.default_package_id_mode=minor_mode
         conan config set general.revisions_enabled=1
     
     - name: install cmake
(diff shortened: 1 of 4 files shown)


garethsb:
Since the mode affects the binary package ids, this has the downside that packages cannot be downloaded from Conan Center Index and must be built locally.

Without further work, builds on GitHub Actions are all going to take a lot longer, about an hour rather than 30 minutes.

I think the solution would be to use the [Conan Cache GitHub Action](https://github.com/marketplace/actions/conan-cache).

PR #219, opened 2021-12-10
Add application/sdp+json as the precise media type for the experimental SDP-as-JSON response

garethsb:
As suggested by @rjb1000 in https://github.com/garethsb/nmos-cpp-examples/discussions/1#discussioncomment-1682324.


rjb1000:
Cheers!

PR #220, opened 2021-12-10
Fix many spec URLs, mainly those that were returning 404 Not Found due to the retagging of v1.x as v1.x.0

garethsb:
I made the decision to fix the links to still point at GitHub for now, but I hacked a quick script out of a regex search for all the links to check no `404 Not Found` responses.

Switching them over to specs.amwa.tv could be done later, but it might mean losing some of the precision in e.g. links to specific lines of the RAML or JSON Schemas.

The script found some other URL typos as well that I've fixed.

### DIFF
--- a/Development/nmos/group_hint.h
+++ b/Development/nmos/group_hint.h
@@ -5,7 +5,7 @@
 #include "nmos/string_enum.h"
 
 // Group Hint
-// See https://github.com/AMWA-TV/nmos-parameter-registers/blob/main/tags/grouphint.md
+// See https://specs.amwa.tv/nmos-parameter-registers/branches/main/tags/grouphint.html
 namespace nmos
 {
     namespace fields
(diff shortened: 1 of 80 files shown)


garethsb:
Now switched all of these over to the closest generated specs.amwa.tv links.

I have checked the 197 distinct specs.amwa.tv links with a script like this...

```sh
curl -s -S -f -I -o temp https://specs.amwa.tv/...
```

This prints nothing if the page is successfully fetched or something like this if not...

```
curl: (22) The requested URL returned error: 404 Not Found
```

There are none of those right now. Some of the fragment identifiers could still be wrong (or in the case of the API docs generated from the RAML, correct, but not effective in bringing up the right pop-up).

Development/nmos/group_hint.h
```diff
@@ -38,4 +38,4 @@ namespace nmos
     group_hint parse_group_hint(const utility::string_t& group_hint);
 }
 
-#endif
+/#endif
```
garethsb, line 41:
```suggestion
#endif
```

PR #224, opened 2021-12-14
Add support for allowing dot being used inside the rql query key path

lo-simon:
In RQL, the `.` character is used to represent the key separator for the key path, use of the `.` escape sequence which allows `.` can be used inside the key.
e.g. handle RQL on `tags`, where the tag names might include `.`

### DIFF
--- a/Development/cpprest/json_utils.cpp
+++ b/Development/cpprest/json_utils.cpp
@@ -43,15 +43,24 @@ namespace web
 
         // insert a field into the specified object at the specified key path (splitting it on '.' and inserting sub-objects as necessary)
         // only if the value doesn't already contain a field matching that key path (except for the required sub-objects or null values)
-        bool insert(web::json::object& object, const utility::string_t& key_path, const web::json::value& field_value)
+        bool insert(web::json::object& object, const utility::string_t& key_path_, const web::json::value& field_value)
+        {
+            std::vector<utility::string_t> key_path;
+            boost::algorithm::split(key_path, key_path_, [](utility::char_t c) { return _XPLATSTR('.') == c; });
+
+            return insert(object, key_path, field_value);
+        }
+
+        // insert a field into the specified object at the specified key path
+        // only if the value doesn't already contain a field matching that key path (except for the required sub-objects or null values)
+        bool insert(web::json::object& object, const std::vector<utility::string_t>& key_path, const web::json::value& field_value)
         {
             web::json::object* pobject = &object;
-            utility::string_t::size_type key_first = 0;
-            do
+
+            size_t count = 0;
+            for (auto key : key_path)
             {
-                const utility::string_t::size_type key_last = key_path.find_first_of(_XPLATSTR("."), key_first);
-                const utility::string_t key = key_path.substr(key_first, details::count(key_first, key_last));
-                if (utility::string_t::npos != key_last)
+                if (++count < key_path.size())
                 {
                     auto& field = (*pobject)[key];
                     // do not replace (non-null) values for duplicate keys; other policies (replace, promote to array, or throw) would be possible...
@@ -66,8 +75,8 @@ namespace web
                     if (field.is_null()) field = field_value;
                     else return false;
                 }
-                key_first = utility::string_t::npos != key_last ? key_last + 1 : key_last;
-            } while (utility::string_t::npos != key_first);
+            }
+
             return true;
         }
 
--- a/Development/cpprest/json_utils.cpp
+++ b/Development/cpprest/json_utils.cpp
@@ -3,6 +3,7 @@
 #include <list>
 #include <boost/algorithm/string/find.hpp>
 #include <boost/algorithm/string/predicate.hpp>
+#include <boost/algorithm/string/split.hpp>
 #include "cpprest/base_uri.h" // for web::uri::decode
 #include "cpprest/regex_utils.h"
 
@@ -75,18 +76,24 @@ namespace web
         // if any arrays are encountered on the key path, results is an array, otherwise it's a non-array value
         bool extract(const web::json::object& object, web::json::value& results, const utility::string_t& key_path)
         {
+            std::vector<utility::string_t> key_path_;
+            boost::algorithm::split(key_path_, key_path, [](utility::char_t c) { return U('.') == c; });
+
+            return extract(object, results, key_path_);
+        }
+
+        // find the value of a field or fields from the specified object and searching key path arrays as necessary
+        // returns true if the value has at least one field matching the key path
+        // if any arrays are encountered on the key path, results is an array, otherwise it's a non-array value
+        bool extract(const web::json::object& object, web::json::value& results, const std::vector<utility::string_t>& key_path)
+        {
             bool match = false;
             results = web::json::value::null();
 
             std::list<const web::json::object*> pobjects(1, &object);
-            utility::string_t::size_type key_first = 0;
-            do
+            for (auto key : key_path)
             {
-                const utility::string_t::size_type key_last = key_path.find_first_of(_XPLATSTR("."), key_first);
-                utility::string_t key = key_path.substr(key_first, details::count(key_first, key_last));
-                // key may still contain the dot escape sequences, must decode them before key matching
-                key = web::uri::decode(key);
-                if (utility::string_t::npos != key_last)
+                if (key != *(key_path.end() - 1))
                 {
                     // not the leaf key, so map each object to the specified field, searching arrays and filtering out other types
                     for (auto it = pobjects.begin(); pobjects.end() != it; it = pobjects.erase(it))
@@ -155,8 +162,7 @@ namespace web
                         }
                     }
                 }
-                key_first = utility::string_t::npos != key_last ? key_last + 1 : key_last;
-            } while (utility::string_t::npos != key_first);
+            }
 
             return match;
         }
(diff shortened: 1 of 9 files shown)


garethsb:
This would be great to solve, and requiring dots within names to be percent-encoded is a good idea.

However, to implement this, I think we need to treat `.` specially, not `%2E`.

Consider `foo.bar.baz%252Equx`.

We want it to mean `foo`.`bar`.`baz%2Equx` but I think if you handle it with two rounds of percent-decode, you'll end up with `foo`.`bar`.`baz.qux`.

Development/rql/test/rql_test.cpp
```diff
@@ -0,0 +1,27 @@
...
+#include "bst/test/test.h"
+#include "cpprest/json_utils.h"
+
+////////////////////////////////////////////////////////////////////////////////////////////
+BST_TEST_CASE(testRqlParse)
+{
+    const auto test_value = web::json::value_of({
+        { U("foo"), web::json::value_of({
+            { U("bar"), web::json::value_of({
+                { U("baz.qux"), U("quux")}
```
garethsb, line 13:
```suggestion
                { U("baz.qux"), U("quux")},
                { U("baz%2Equx"), U("quuux")}
```

Development/rql/test/rql_test.cpp
```diff
@@ -0,0 +1,27 @@
...
+
+    const utility::string_t query_rql = U("matches(foo.bar.baz%2Equx,quux)");
+    const auto rql_query = rql::parse_query(query_rql);
+
+    const auto args = rql_query.at(U("args"));
+    const auto key_path = args.as_array().at(0).as_string();
+    web::json::value results;
+
+    BST_REQUIRE_EQUAL(true, web::json::extract(test_value.as_object(), results, key_path));
+    BST_REQUIRE_EQUAL(U("quux"), results.as_string());
```
garethsb, line 26:
```suggestion
    {
        const utility::string_t query_rql = U("matches(foo.bar.baz%2Equx,quux)");
        const auto rql_query = rql::parse_query(query_rql);

        const auto args = rql_query.at(U("args"));
        const auto key_path = args.as_array().at(0).as_string();
        web::json::value results;

        BST_REQUIRE_EQUAL(true, web::json::extract(test_value.as_object(), results, key_path));
        BST_REQUIRE_EQUAL(U("quux"), results.as_string());
    }

    {
        const utility::string_t query_rql = U("matches(foo.bar.baz%252Equx,quux)");
        const auto rql_query = rql::parse_query(query_rql);

        const auto args = rql_query.at(U("args"));
        const auto key_path = args.as_array().at(0).as_string();
        web::json::value results;

        BST_REQUIRE_EQUAL(true, web::json::extract(test_value.as_object(), results, key_path));
        BST_REQUIRE_EQUAL(U("quuux"), results.as_string());
    }
```

garethsb (commented):
Initial suggested test case tweak...

garethsb:
I think an approach to satisfy the requirements could be something like this: 

* Modify original `web::json::extract` to be:
  ```C++
  bool extract(const web::json::object& object, web::json::value& results, const std::vector<utility::string_t>& key_path)
  ```
  and make the existing function do a simple split on `.` (using `boost::algorithm::split`) before calling this new one. Neither function should do `web::uri::decode`.
* Modify the `nmos::make_extractor` in _nmos/query_utils.cpp_ to do split on `.` and `web::uri::decode` each component (using `boost::transformed`) before calling the new `web::json::extract` overload
* Modify `rql::make_value` so that it doesn't necessarily `web::uri::decode` the type (maybe?) and value, e.g. by adding a Boolean flag and/or a 'user'-specified decode function
* Modify `rql::parse_query` to take that flag too and pass it to `make_value`

To maintain backward compatibility and flexibility, we would need to think harder about the last two bullet points, to ensure that type look-up and value equality works they way we want, e.g. is `f%6F%6F` equal to `f%6f%6f` and/or `foo`.

Development/cpprest/json_utils.cpp
```diff
@@ -74,17 +75,25 @@ namespace web
...
 
             std::list<const web::json::object*> pobjects(1, &object);
-            utility::string_t::size_type key_first = 0;
-            do
+            for (auto key : key_path)
             {
-                const utility::string_t::size_type key_last = key_path.find_first_of(_XPLATSTR("."), key_first);
-                const utility::string_t key = key_path.substr(key_first, details::count(key_first, key_last));
-                if (utility::string_t::npos != key_last)
+                if (key != *(key_path.end() - 1))
```
garethsb, line 96: That checks that the _value_ of the key isn't equal to the last one in the `key_path`... but what about paths like `foo.bar.foo`?

Development/cpprest/json_utils.cpp
```diff
@@ -74,17 +75,25 @@ namespace web
         // returns true if the value has at least one field matching the key path
         // if any arrays are encountered on the key path, results is an array, otherwise it's a non-array value
         bool extract(const web::json::object& object, web::json::value& results, const utility::string_t& key_path)
+        {
+            std::vector<utility::string_t> key_path_;
+            boost::algorithm::split(key_path_, key_path, [](utility::char_t c) { return U('.') == c; });
```
garethsb, line 80: I think we try to avoid `U` preprocessor macro in the library code? Use `_XPLATSTR` as below.

garethsb (commented):
This is heading in the right direction apart from one bug. However it doesn't address the issue I pointed out with value equality.

Consider the request URL:

`/x-nmos/query/v1.3/nodes?query.rql=eq(label,f%6F%6F)`

Does that match a Node with a `label` of `foo` or not?

Similarly what about a query like:

`/x-nmos/query/v1.3/flows?query.rql=ge(bit_depth,%31%36)`

Does that match audio Flows with `bit_depth` greater than or equal to 16?

Development/cpprest/json_utils.cpp
lo-simon: Okay, will change to use `_XPLATSTR` instead of `U`.

FYI `U` is already used in the preprocess function.
            

>utility::string_t preprocess(const utility::string_t& value)
>{
>    // regex pattern matches JSON strings, or single or multi-line comments
>    // only strings are captured
>    static const utility::regex_t string_or_comment(U(R"-regex-(("[^"\\]*(?:\\.[^"\\]*)*")|(?:\/\/[^\r\n]+)|(?:\/\*[\s\S]*?\*\/))-regex-"));
>    // format pattern uses the first capture group to copy strings into the output
>    // having inserted a single space to ensure tokens are not coalesced
>    return bst::regex_replace(value, string_or_comment, U(" $1"));
>}

Development/cpprest/json_utils.cpp
lo-simon: Well-spotted will fix it

lo-simon:
> This is heading in the right direction apart from one bug. However it doesn't address the issue I pointed out with value equality.
> 
> Consider the request URL:
> 
> `/x-nmos/query/v1.3/nodes?query.rql=eq(label,f%6F%6F)`
> 
> Does that match a Node with a `label` of `foo` or not?
> 
> Similarly what about a query like:
> 
> `/x-nmos/query/v1.3/flows?query.rql=ge(bit_depth,%31%36)`
> 
> Does that match audio Flows with `bit_depth` greater than or equal to 16?

The not-to-decode is only applied on the key path, in the above examples, they are the `label` and the `bit_depth`, where decode is always applied on the value part. So the above queries do work as expected.

Development/cpprest/json_utils.cpp
garethsb: > FYI `U` is already used in the preprocess function.

You're right. I think the intention may have been that anything we might conceivably have wanted to push back to Microsoft/cpprestsdk should use `_XPLATSTR` so that the code could be built with `U` disabled. So that probably means anything in the _Development/cpprest_ (or _pplx_) directories... However, there are obviously a few files where I failed to follow my own rule already, including the instance you pointed out! Moreover, since cpprestsdk is in maintenance mode, it's not going to happen, even for the least controversial additions. For now, why not just fix the rest of _this_ file so it uses `_XPLATSTR` consistently, and leave it at that. Thanks. :-)

garethsb:
> The not-to-decode is only applied on the key path,

As discussed on Slack, the proposed implementation is currently to apply the Boolean flag not-to-decode to all but the last argument of each function. This works for many functions which are of the form `operator(<property>,<value>)`, e.g. `eq(label,foo.bar)`, but could fail for some functions we have already like `matches(foo,b%61r,i)` (case-insensitive match) and breaks the current separation between parsing, according to RQL syntax, on one hand and evaluating, using a set of application-specific functions, on the other. Only the particular functions know whether their arguments should be treated as property extractors or not.

We now propose to adopt the RQL-specified array-form for "nested" properties (some way down the [RQL Rules](https://github.com/persvr/rql#rql-rules)) to solve the issue instead:

> Any property can be nested by using an array of properties. To search by the bar property of the object in the foo property we can do:
> `(foo,bar)=3`

(The example has `=3` because it uses the FIQL syntactic sugar, the normalized form (required by IS-04 [Advanced RQL Queries]https://specs.amwa.tv/is-04/releases/v1.3.1/docs/2.5._APIs_-_Query_Parameters.html#advanced-rql-queries-optional)) would be `eq((foo,bar),3)`.)

Since `parse_query` already implements the array form, the changes to it can be reverted, and the extractor created by `nmos::make_extractor` in _nmos/query_utils.cpp_ just needs updating to handle an array key as well as a string key.

This would mean that the following are equivalent:
```
eq(foo.bar.baz,qux)
```
```
eq((foo,bar,baz),qux)
```
But property names with dots would be specified by a client like so:
```
eq((foo,bar.baz),qux)
```

Development/cpprest/json_utils.cpp
lo-simon: okay

Issue #213, opened 2021-11-23
Enhance support of NMOS stream mappings for JPEG XS and e.g. H264, H265, VP9 and AV1


garethsb:
NMOS supports compressed video streams; at the most basic via the `media_type` attribute in the Flow and the `media_types` capability in the Receiver.

Since IS-04 v1.3, we can [register new Flow attributes](https://specs.amwa.tv/nmos-parameter-registers/branches/main/flow-attributes/), and with BCP-004-01 Receiver Capabilities and the [Capabilities register](https://specs.amwa.tv/nmos-parameter-registers/branches/main/capabilities/), new Receiver capabilities. For example, for JPEG XS (see #177), it was agreed that bit rate is an important characteristic to add (as well as being able to indicate color (sub-)sampling for coded video Flows via the same `components` attribute used for raw video Flows).

The nmos-cpp _[sdp](https://github.com/sony/nmos-cpp/tree/master/Development/sdp)_ module is generic and extensible.

However, the nmos-cpp _[nmos/sdp_utils.h](https://github.com/sony/nmos-cpp/blob/master/Development/nmos/sdp_utils.h)_ module and the `nmos::sdp_parameters` structure it defines are not.

I propose we extend the `nmos::sdp_parameters` and related functions to provide SDP read/write for additional media types. If it cannot be done maintaining backward compatibility we can do something alongside instead.

There may be no point making something _fully_ generic if it wouldn't provide any abstraction over the lower level `sdp` namespace functions.

Top priorities are to support [`video/jxsv`](https://www.iana.org/assignments/media-types/video/jxsv), [`video/H264`](https://www.iana.org/assignments/media-types/video/H264), [`video/H265`](https://www.iana.org/assignments/media-types/video/H265), [`video/VP9`](https://www.iana.org/assignments/media-types/video/VP9), [`video/AV1`](https://www.iana.org/assignments/media-types/video/AV1) and `audio/AM824` (which doesn't appear to have an IANA registration...).

I feel the `nmos::sdp_parameters` structure should expose all the required and optional format-specific parameters of these media types as first-class values. There are currently a few missing even for [`video/raw`](https://www.iana.org/assignments/media-types/video/raw) (when used as per ST 2110-20) such as `CMAX` ([noted in the code](https://github.com/sony/nmos-cpp/blob/95536ae32341046dabf66286b373d70e97e3a59a/Development/nmos/sdp_utils.cpp#L1206)).

garethsb:
[...]

Issue #202, opened 2021-10-08
nmos::details::make_sdp_parameters in sdp_utils in case of audio always reprting ptime 1 and payload id 97


aholzinger:
I'm currently stumpling upon this one (sdp_utils.cpp line 190):
```
params.packet_time = 1;

return{ sender.at(nmos::fields::label).as_string(), params, 97, media_stream_ids, details::make_ts_refclk(node, source, sender, ptp_domain) };
```

Wouldn't it be nice if the function would correctly report packet time and payload id?

The problem seems to be that in the data provided in the function parameters neither ptime nor pid are contained.

garethsb:
Those values are not found in the NMOS resources. If your app knows them, it can set them. Like the example app, https://github.com/sony/nmos-cpp/blob/master/Development/nmos-cpp-node/node_implementation.cpp#L911

aholzinger:
That's how I worked around it.

Wouldn't it then be better to return invalid values for the two parameters so that the user will immediately be aware to update these parameters?

garethsb:
It's a little awkward because the format-specific functions in `nmos::details` are exposed to the user only via:

https://github.com/sony/nmos-cpp/blob/4fba3ab6db1b0361d625fc3d7edf307145f5a653/Development/nmos/sdp_utils.h#L23

Adding an `optional<uint64_t> payload_type = nullopt` to that function declaration would be OK, I guess.

But it feels a little wrong to add the (basically) audio-specific `double packet_time` to this function's parameter list.

So I'm not sure of the best approach...

> pre-fetched correct values from the SDP

For what it's worth, depending on the structure of your app, the `get_session_description_sdp_parameters` function and therefore also `parse_session_description` can do this for you.

PR #225, opened 2022-01-10
Refactor (and a few breaking interface changes) of sdp_utils

garethsb:
Refactor (and a few breaking interface changes) of _nmos/sdp_utils.{cpp,h}_ to tease apart format-agnostic SDP creation, parsing and validation from format-specific functionality.

This is intended to simplify future addition of support for `video/jxsv` and other media types mentioned in #213, whether within nmos-cpp or in application code for specific use cases.

### Details

1. `sdp_parameters` is now format-agnostic.
   It no longer includes any format (media type/subtype) specific fields, like `video.width`.
   Instead, the format-specific parameters for the `a=fmtp:` line are serialized to a vector of pairs of string in this object. (It can't be a map because some keys may appear multiple times, like `DID_SDID`.)
   Other common (RFC4566-defined) attributes that are likely to be used, like `a=ptime:`, `a=maxptime:` and `a=framerate:` are added to the format-agnostic struct, as is direct support for one bandwidth `b=` line.
2. There are now free functions to get and set typed format-specific parameters from/to the `fmtp`, `rtpmap` and other member variables.
   E.g. audio sample rate and channel count from the `m=` line, packet time from `a=ptime:`, and in the future, JPEG XS bit rate from the `b=` line.
   These are `make_<media type/subtype>_sdp_parameters` but `get_<media type/subtype>_parameters` rather than `parse_` because there are other members of `sdp_parameters` which are not reflected into the `<media type/subtype>_parameters` objects.

### Breaking interface changes

* Constructors of `sdp_parameters` that took `sdp_parameters::video_t`, `audio_t`, etc. are deprecated (so not a breaking change, yet!).
  Equivalent constructors for new media types should not be added in the future. The free function approach should be adopted. See below...
* Instead of `sdp.video`, `sdp.audio`, etc. application code must now use `get_video_raw_parameters(sdp)`, `get_audio_L_parameters(sdp)`, etc.
* `sdp_parameters::audio_t::sample_rate` is now a `uint64_t` rather than an `nmos::rational` to correspond more closely with the SDP spec and roundtrip better

### Adding a new media type

0. Create a struct `<media type/subtype>_parameters` with members for the necessary format parameters.
   You may not want to include everything mentioned in the IANA registration, but that's a good place to start.
   See e.g. `video_raw_parameters`.
1. Add `make_<media type/subtype>_parameters` which constructs the struct from info in the IS-04 resources and possibly commonly required additional info not carried in IS-04 resources (e.g. packet time for audio).
   See e.g. `make_video_raw_parameters`.
2. Add `make_<media type/subtype>_sdp_parameters` which constructs the format-agnostic `sdp_parameters` from the struct (and the non-transport, non-format-specific, parameters such as the SDP session name).
   This function can usually be very simple, for example building the rtpmap and fmtp from the `<media type/subtype>_parameters`.
   Also add an equivalent overload of `make_sdp_parameters`; should be a one-liner.
   See e.g. `make_video_raw_sdp_parameters`.
3. Add `get_<media type/subtype>_parameters` which constructs the struct from `sdp_parameters`.
   This function can usually be very simple, for example extracting the values for the `<media type/subtype>_parameters` from the rtpmap and fmtp.
   See e.g. `get_video_raw_parameters`.
4. Support creation of SDP files of the new format based on the IS-04 resources.
   Within the nmos-cpp library this would be done by updating the `nmos::make_sdp_parameters` overload to use the new `make_<media type/subtype>_sdp_parameters` function.
   This nmos-cpp library function obviously couldn't be updated by application code, but since this function is only called from application code, such applications would use their own function for the application-specific media types instead.
5. Support the new format for validation of SDP files for receiver capabilities.
   Within the nmos-cpp library, this would be done by updating `nmos::validate_sdp_parameters`.
   This nmos-cpp library function couldn't be updated by application code. It _is_ called by a library function, `nmos::parse_rtp_transport_file`, but since _that_ function isn't used if the application provides an `nmos::transport_file_parser`. An application can use `nmos::details::parse_rtp_transport_file` which now provides a way for application code to easily plug in the application-specific SDP parameters validation.

### DIFF
--- a/Development/nmos-cpp-node/node_implementation.cpp
+++ b/Development/nmos-cpp-node/node_implementation.cpp
@@ -22,6 +22,7 @@
 #include "nmos/connection_resources.h"
 #include "nmos/connection_events_activation.h"
 #include "nmos/events_resources.h"
+#include "nmos/format.h"
 #include "nmos/group_hint.h"
 #include "nmos/interlace_mode.h"
 #ifdef HAVE_LLDP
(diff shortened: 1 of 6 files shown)


garethsb:
### Next steps

1. Another PR to extend the existing formats, with the missing format-specific parameters etc.

* [`video/raw`](https://www.iana.org/assignments/media-types/video/raw) (when used as per ST 2110-20/-21)
  * Required parameters
    * `PM`, `SSN` (currently written with hard-coded values and not read)
  * Optional parameters
    * `RANGE`
    * `MAXUDP`
    * `PAR`
    * `TROFF` (also for `video/SMPTE2022-6` when used as per ST 2022-8)
    * `CMAX`
* `video/SMPTE2022-6`
  * Other
    * Frame rate as `a=framerate:`
* `audio/L`
  * Other
    * Maximum packet time in an `a=maxptime:` attribute (maybe it's good enough it's in the `sdp_parameters` now? just inconsistent with `packet_time` in `sdp_parameters::audio_t`/`audio_L_parameters`)

2. A PR to add e.g. `video/jxsv` (when the NMOS mapping, [BCP-006-01](https://specs.amwa.tv/bcp-006-01) is fully spec'ed...) and/or a PR to show how e.g. nmos-cpp-node application code could add support for its own custom media type.

garethsb:
@aholzinger, after our discussions around #202, you may be interested in this also.

garethsb:
@peterbrightwell, @David-P-B, @pedro-alves-ferreira, you may also be interested in this.

PR #229, opened 2022-01-18
openssl1.1.1m

garethsb:
Cf. #198.

### DIFF
--- a/Development/conanfile.txt
+++ b/Development/conanfile.txt
@@ -2,7 +2,7 @@
 boost/1.77.0
 cpprestsdk/2.10.18
 websocketpp/0.8.2
-openssl/1.1.1l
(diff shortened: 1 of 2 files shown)


garethsb:
CI failure is unrelated. Looks like the ppa deadsnakes repo (used to install Python) is no longer supporting Ubuntu 14.04 Trusty since end of last week. They've also killed off support for Ubuntu 16.04 Xenial, since it too is end of life. One possibility if you want to maintain the Ubuntu 14.04 job, would be to build Python 3.6 from scratch, e.g. following Method 2 rather than Method 1 in http://devopspy.com/python/install-python-3-6-ubuntu-lts/. Obviously it'll increase the running time further...

PR #233, opened 2022-01-20
Use nmos::get_domain and nmos::get_host_name to get the provided defaults for the DNS-SD advertisements

garethsb:


lo-simon (commented):
LGTM, will merge once all passed.

PR #242, opened 2022-02-04
More helpers for for supporting additional media types

garethsb:


Development/nmos/node_resources.cpp
```diff
@@ -318,9 +319,19 @@ namespace nmos
 
         data[U("media_type")] = value::string(media_type.name);
 
+        if (!color_sampling.name.empty())
+        {
```
lo-simon, line 323: Would it make sense, you update this PR with the master after the #241 merged, then we can use write
`if (!color_sampling.empty())` instead of
`if (!color_sampling.name.empty())`

garethsb: Yes. I'll rebase and push.

PR #250, opened 2022-04-22
Test unicast DNS-SD as well as multicast, with both Avahi and mDNSResponder, on Ubuntu 20.04

garethsb:
Test unicast DNS-SD as well as multicast, with both Avahi and mDNSResponder, on Ubuntu 20.04

* Add `matrix.dns_sd_mode` as 'unicast' or 'multicast', used in job name and results filenames
* For now, unicast DNS-SD testing is not supported on Windows and macOS
* Add _api.testsuite.nmos.tv_ and _mocks.testsuite.nmos.tv_ to _/etc/hosts_ (as per https://github.com/AMWA-TV/nmos-testing/blob/master/test_data/BCP00301/README.md#hosts-files)
* Stomp on _/etc/resolv.conf_ (`systemd-resolve --set-dns` didn't replace only add) to ensure that only the mock DNS server is used and restart the Avahi or mDNSResponder daemon (reloading and/or invalidating/flushing systemd-resolve and nscd caches between test suite runs was found to be unnecessary)
* Configure nmos-cpp-node to timeout registration requests (after 5s) and retry long-running DNS-SD queries (after 10s), well before the testing tool's `DNS_SD_ADVERT_TIMEOUT` (30s by default)
* Run the testing tool with elevated permissions when testing unicast DNS-SD (unfortunately necessitating quick hack of `pip install`-ing packages as root)
* Note the domain names passed to _run_nmos_testing.sh_ mustn't have trailing dots to avoid "[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: Hostname mismatch, certificate is not valid for 'api.testsuite.nmos.tv.'. (_ssk.c:1131)" from e.g. IS-07-02 `test_05`

Unrelated changes:
* Update to conan-cmake 0.18.1 to support Visual Studio 2022 (a.k.a. "17", or "MSVC 19.31.31105.0" as now [installed on GitHub windows-latest runnner](https://github.com/actions/virtual-environments/issues/4856))
* However, switch to windows-2019 rather than windows-latest (now an alias for windows-2022) because [there aren't yet any Conan binary packages for VS 2022](https://github.com/conan-io/conan-center-index/issues/8795) and _testssl.sh_ is consistently failing with "Fatal error: No IPv4/IPv6 address(es) for "nmos-api.local" available" on the newer virtual environment
* Add `host_addresses` to config because on Windows I'm seeing "No matching mDNS announcement" failures in IS-04-02 and IS-04-03 and a second "Registered address" in the _nodeoutput_ log for the "vEthernet (nat)" adapter, despite it being disabled in the "windows setup" step...
* Use latest https://github.com/DannyBen/kojo which means whitespace-only lines due to indented imports are gone

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -23,13 +23,20 @@ jobs:
         force_cpprest_asio: [false]
         dns_sd_mode: [multicast, unicast]
         exclude:
+          # install_mdns is only meaningful on Linux
           - os: macos-11
             install_mdns: true
           - os: windows-2019
             install_mdns: true
+          # for now, unicast DNS-SD tests are only implemented on Linux
           - os: macos-11
             dns_sd_mode: unicast
           - os: windows-2019
+            dns_sd_mode: unicast
+          # for now, exclude unicast DNS-SD with mDNSResponder due to
+          # intermittent *** buffer overflow detected *** in mdnsd
+          - os: ubuntu-20.04
+            install_mdns: true
             dns_sd_mode: unicast
         include:
           - os: windows-2019
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -444,8 +444,8 @@ jobs:
         if [[ "${{ runner.os }}" == "Linux" ]]; then
           if [[ "${{ matrix.install_mdns }}" == "true" ]]; then
             echo "Restarting mdnsd"
-            #sudo /etc/init.d/mdns start
-            sudo /usr/sbin/mdnsd -debug &
+            sudo /etc/init.d/mdns start
+            #sudo /usr/sbin/mdnsd -debug &
             sleep 2
 
             dns-sd -V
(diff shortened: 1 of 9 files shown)


garethsb:
Commit caabf8e resolves failures for Avahi unicast via the suggestion made by @wsneijers in https://github.com/sony/nmos-cpp/issues/128, without any apparent effect on the mDNSResponder results. However, we still need to do a bit of digging...

garethsb:
@lo-simon This is as far as I'm going to take this for now. I've resquashed all the commits to tell a sensible story, with the main work to enable unicast DNS-SD testing in the first two commits separate from the bug fix for Avahi. I've finally excluded the mDNSResponder unicast job because of the intermittent failures. But I believe the local testing we've done with unicast DNS-SD with mDNSResponder after the bug fix is encouraging... If https://github.com/AMWA-TV/nmos-testing/pull/677 is merged soon, I'll force push once more to use `master` rather than the PR branch.

Issue #259, opened 2022-05-06
Compilation under Ubuntu 22.04 (Jammy) and gcc 11.2.0 fails with error during linking


rhastie:
Just testing moving the container to the new Ubuntu 22.04 (Jammy) LTS release. This means moving to gcc 11.2.0 which currently isn't tested in the CI/CD pipeline...

I get a compile error during linking...

```
[ 92%] Building CXX object CMakeFiles/nmos-cpp.dir/sdp/sdp_grammar.cpp.o
[ 92%] Linking CXX static library libnmos-cpp.a
[ 92%] Built target nmos-cpp
[ 92%] Building CXX object CMakeFiles/nmos-cpp-registry.dir/nmos-cpp-registry/main.cpp.o
[ 93%] Building CXX object CMakeFiles/nmos-cpp-node.dir/nmos-cpp-node/node_implementation.cpp.o
[ 93%] Building CXX object CMakeFiles/nmos-cpp-node.dir/nmos-cpp-node/main.cpp.o
[ 93%] Building CXX object CMakeFiles/nmos-cpp-test.dir/nmos-cpp-test/main.cpp.o
In file included from /usr/include/signal.h:328,
                 from /home/nmos-cpp/Development/third_party/catch/catch.hpp:6448,
                 from /home/nmos-cpp/Development/bst/test/detail/catch-1.10.0.h:18,
                 from /home/nmos-cpp/Development/bst/test/test.h:48,
                 from /home/nmos-cpp/Development/nmos-cpp-test/main.cpp:2:
/home/nmos-cpp/Development/third_party/catch/catch.hpp:6471:33: error: size of array 'altStackMem' is not an integral constant-expression
 6471 |         static char altStackMem[SIGSTKSZ];
      |                                 ^~~~~~~~
/home/nmos-cpp/Development/third_party/catch/catch.hpp:6522:45: error: size of array 'altStackMem' is not an integral constant-expression
 6522 |     char FatalConditionHandler::altStackMem[SIGSTKSZ] = {};
      |                                             ^~~~~~~~
make[2]: *** [CMakeFiles/nmos-cpp-test.dir/build.make:76: CMakeFiles/nmos-cpp-test.dir/nmos-cpp-test/main.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:326: CMakeFiles/nmos-cpp-test.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
[...]

Issue #251, opened 2022-04-22
Add ubuntu-22.04 test platform


garethsb:
Ubuntu 22.04 is now LTS.

Unfortunately not yet available as a GitHub virtual env - https://github.com/actions/virtual-environments/issues/5428

garethsb:
Public beta now available: https://github.com/actions/virtual-environments/issues/5490
Main changes: default GCC is 11.2, and default Python is 3.10

PR #262, opened 2022-05-21
Ubuntu 22.04

garethsb:
To start with, apply patch for non-constant SIGSTKSZ. See #259.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -44,6 +44,11 @@ jobs:
             use_conan: true
             force_cpprest_asio: true
             dns_sd_mode: multicast
+          - os: ubuntu-22.04
+            install_mdns: false
+            use_conan: true
+            force_cpprest_asio: false
+            dns_sd_mode: multicast
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/src/build-test.yml
+++ b/.github/workflows/src/build-test.yml
@@ -44,6 +44,11 @@ jobs:
             use_conan: true
             force_cpprest_asio: true
             dns_sd_mode: multicast
+          - os: ubuntu-22.04
+            install_mdns: false
+            use_conan: true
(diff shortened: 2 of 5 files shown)


garethsb:
I think the Ubuntu 14.04 failure was an intermittent one.

This patch seems to solve the Catch compile issue on Ubuntu 22.04, but... I'm seeing the _install test_ step fail... looks like an issue with `bst::optional` being `std::optional` when building `my-nmos-node`, but presumably being `boost::optional` when building the rest of `nmos-cpp`. That's odd, because C++17 _is_ the default language standard with GCC 11.2, so I expected `std::optional` throughout.

garethsb:
Ah, the problem will be related to https://github.com/sony/nmos-cpp/issues/206...

We currently have:

https://github.com/sony/nmos-cpp/blob/dff1585b6056ffa8d419b7a3715b38aa2c8f3e00/Development/cmake/NmosCppCommon.cmake#L22-L24

This CMake file is included when building the library (and example apps in the same go) but not when building `my-nmos-node`. The `nmos-cpp::compile-settings` have this to specify the _minimum_ requirements:

https://github.com/sony/nmos-cpp/blob/dff1585b6056ffa8d419b7a3715b38aa2c8f3e00/Development/cmake/NmosCppCompileSettings.cmake#L2

Generally, maybe we should default to the compiler default (remove setting `CMAKE_CXX_STANDARD` completely), so that with GCC 11.2 it would build with C++17 unless overridden by the user. The reason we didn't is that the default for GCC 4 and GCC is C++98, which ain't gonna work.

garethsb:
> The reason we didn't is that the default for GCC 4 and GCC 5 is C++98, which ain't gonna work.

Huh, so the Ubuntu 14.04 build has succeeded too. OK then...

garethsb:
Assuming the latest build-test is all green ✔️ I think this could be merged and resolves #251 and resolves #259.

PR #267, opened 2022-06-20
Bump Boost to 1.79.0

garethsb:
Boost 1.79.0 was released 16 April 2022.
https://www.boost.org/users/history/version_1_79_0.html

(Merge after #266.)


garethsb:
@lo-simon Note I have assumed you will update the Raspberry Pi instructions independently as I'm not able to test those here.

PR #269, opened 2022-06-20
Bump nlohmann/json to 3.10.5

garethsb:
Was 3.9.1 but Conan recipe is already on 3.10.5 since 15 January 2022. (Current pboettch/json-schema-validator GitHub Actions workflow uses 3.10.2)


garethsb:
@lo-simon although there has been some activity on pboettch/json-schema-validator since 2020, there hasn't been a release since 2.1.0, so I haven't attempted to take any of the fixes or features from it

PR #272, opened 2022-08-25
Relax fmtp requirement for whitespace after semicolon

garethsb:
ST 2110-20 requires whitespace after the semicolon within format-specific parameters:

tempFileForShare_20220825-064715

However, RFC 4566 (and RFC 8866) doesn't specify this syntax, and many examples, in both RFCs and SMPTE specs, show semicolon used as separator with no following whitespace.

We relaxed the SDPoker check some time ago... https://github.com/AMWA-TV/sdpoker/pull/6.

Note that this does not accept whitespace _preceding_ the semicolon, partly because I'm lazy and handling trailing whitespace after the last parameter (with no semicolon) is a harder change, and because ST 2110-20 also says no whitespace within the name or value.

### DIFF
--- a/Development/sdp/sdp_grammar.cpp
+++ b/Development/sdp/sdp_grammar.cpp
@@ -160,8 +160,8 @@ namespace sdp
         // ST 2110-20:2022 says "the <format specific parameters> section shall consist of a sequence of
         // media type parameter entries, separated by the semicolon (";") character followed by whitespace"
         // but RFC 4566 does not itself specify the syntax of format-specific parameters and many examples
-        // in other RFCs and SMPTE standards are inconsistent, so make the following whitespace optional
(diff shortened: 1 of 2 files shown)


garethsb:
Should we also change this line:

https://github.com/sony/nmos-cpp/blob/3a904a3fcc39057a8db74656a697f0d97d8a3651/Development/sdp/sdp_grammar.cpp#L566

to:

```C++
s += named_values_converter.format(params);
```

i.e. no trailing separator?

ST 2110-20:2022 requires no trailing semicolon, but some readers would argue that ST 2110-20:2017 required it to be present. (I argued for making it optional in the revision but seems like I didn't win that.)

garethsb:
> Note that this does not accept whitespace _preceding_ the semicolon, partly because I'm lazy and handling trailing whitespace after the last parameter (with no semicolon) is a harder change

And now it does both of those, @lo-simon.

lo-simon (approved):
LGTM

PR #275, opened 2022-09-29
OSCP Stapling & HSTS support

jonathan-r-thorpe:
Added OSCP Stapling & HSTS support to comply with *all* BCP-003-01 test cases

### DIFF
--- a/Development/cpprest/http_utils.cpp
+++ b/Development/cpprest/http_utils.cpp
@@ -476,25 +476,25 @@ namespace web
                 return results;
             }
 
-            utility::string_t make_hsts_header(const htst& value)
+            utility::string_t make_hsts_header(const hsts& value)
             {
                 directives result;
 
-                // invalid htst default to HTST disable
-                auto htst_ = value;
-                if (htst_.max_age < 0) htst_ = htst{};
+                // invalid hsts default to HSTS disable
+                auto hsts_ = value;
+                if (hsts_.max_age < 0) hsts_ = hsts{};
 
-                if (htst_.max_age >= 0)
+                if (hsts_.max_age >= 0)
                 {
-                    result.push_back({ U("max-age"), utility::ostringstreamed(htst_.max_age) });
-                    if (htst_.max_age > 0 && htst_.includeSubDomains) result.push_back({ U("includeSubDomains"), {} });
+                    result.push_back({ U("max-age"), utility::ostringstreamed(hsts_.max_age) });
+                    if (hsts_.max_age > 0 && hsts_.includeSubDomains) result.push_back({ U("includeSubDomains"), {} });
                 }
                 return make_directives_header(result);
             }
 
-            htst parse_htst_header(const utility::string_t& value)
+            hsts parse_hsts_header(const utility::string_t& value)
             {
-                htst result;
+                hsts result;
                 auto directives = parse_directives_header(value);
                 const auto max_age = std::find_if(directives.begin(), directives.end(), [](const ptoken_param& param) { return boost::algorithm::iequals(param.first, U("max-age")); });
                 if (directives.end() != max_age) result.max_age = utility::istringstreamed(max_age->second, 0);
--- a/Development/cpprest/http_utils.cpp
+++ b/Development/cpprest/http_utils.cpp
@@ -330,6 +330,111 @@ namespace web
                 return result;
             }
 
+            utility::string_t make_directives_header(const directives& values)
+            {
+                utility::string_t result;
+                for(auto it = values.begin(); it != values.end(); ++it)
+                {
+                    if (it != values.begin()) { result.push_back(U(';')); }
+                    result.append(it->first);
+                    if (!it->second.empty())
+                    {
+                        result.push_back(U('='));
+                        result.append(it->second);
+                    }
+                }
+                return result;
+            }
+
+            directives parse_directives_header(const utility::string_t& value)
+            {
+                enum {
+                    pre_param,
+                    pre_param_name,
+                    param_name,
+                    pre_param_value,
+                    param_value,
+                    param_value_token,
+                    param_value_quoted_string,
+                    param_value_quoted_string_escape
+                } state = pre_param_name;
+
+                // token      = 1*<any CHAR except CTLs or separators>
+                // separators = "(" | ")" | "<" | ">" | "@"
+                //            | "," | ";" | ":" | "\" | <">
+                //            | "/" | "[" | "]" | "?" | "="
+                //            | "{" | "}" | SP | HT
+                // see https://datatracker.ietf.org/doc/html/rfc2616#section-2.2
+                auto is_tchar = [](utility::char_t c)
+                {
+                    static const utility::string_t separators{ U("()<>@,;:\\\"/[]?={} \t") };
+                    return std::isalnum(c) || (!std::iscntrl(c) && std::string::npos == separators.find(c));
+                };
+
+                directives result;
+                utility::string_t name;
+                for (auto c : value)
+                {
+                    switch (state)
+                    {
+                    case pre_param:
+                        if (U(';') == c) { state = pre_param_name; break; }
+                        if (U(' ') == c || U('\t') == c) { break; }
+                        throw std::invalid_argument("invalid value, expected ';'");
+                    case pre_param_name:
+                        if (is_tchar(c)) { name.push_back(c); state = param_name; break; }
+                        if (U(' ') == c || U('\t') == c) { break; }
+                        throw std::invalid_argument("invalid directive name, expected tchar");
+                    case param_name:
+                        if (is_tchar(c)) { name.push_back(c); break; }
+                        result.push_back({ name, {} }); name.clear();
+                        if (U('=') == c) { state = param_value; break; }
+                        if (U(';') == c) { state = pre_param_name; break; }
+                        if (U(' ') == c || U('\t') == c) { state = pre_param_value; break; }
+                        throw std::invalid_argument("invalid directive name, expected tchar");
+                    case pre_param_value:
+                        if (U('=') == c) { state = param_value; break; }
+                        if (U(' ') == c || U('\t') == c) { break; }
+                        if (U(';') == c) { state = pre_param_name; break; }
+                        throw std::invalid_argument("invalid directive, expected '='");
+                    case param_value:
+                        if (is_tchar(c)) { result.back().second.push_back(c); state = param_value_token; break; }
+                        if (U('"') == c) { state = param_value_quoted_string; break; }
+                        if (U(' ') == c || U('\t') == c) { break; }
+                        throw std::invalid_argument("invalid directive value, expected tchar or '\"'");
+                    case param_value_token:
+                        if (is_tchar(c)) { result.back().second.push_back(c); break; }
+                        if (U(';') == c) { state = pre_param_name; break; }
+                        if (U(' ') == c || U('\t') == c) { state = pre_param; break; }
+                        throw std::invalid_argument("invalid directive value, expected tchar");
+                    case param_value_quoted_string:
+                        if (U('"') == c) { state = pre_param_name; break; }
+                        if (U('\\') == c) { state = param_value_quoted_string_escape; break; }
+                        result.back().second.push_back(c);
+                        break;
+                    case param_value_quoted_string_escape:
+                        result.back().second.push_back(c);
+                        state = param_value_quoted_string;
+                        break;
+                    default:
+                        throw std::logic_error("unreachable code");
+                    }
+                }
+
+                if (!name.empty())
+                {
+                    switch (state)
+                    {
+                    case param_name:
+                        result.push_back({ name, {} }); break;
+                    default:
+                        throw std::logic_error("unreachable code");
+                    }
+                }
+
+                return result;
+            }
+
             namespace details
             {
                 template <typename TimePoint>
@@ -369,6 +474,33 @@ namespace web
                     results.push_back(std::move(metric));
                 }
                 return results;
+            }
+
+            utility::string_t make_hsts_header(const htst& value)
+            {
+                directives result;
+
+                // invalid htst default to HTST disable
+                auto htst_ = value;
+                if (htst_.max_age < 0) htst_ = htst{};
+
+                if (htst_.max_age >= 0)
+                {
+                    result.push_back({ U("max-age"), utility::ostringstreamed(htst_.max_age) });
+                    if (htst_.max_age > 0 && htst_.includeSubDomains) result.push_back({ U("includeSubDomains"), {} });
+                }
+                return make_directives_header(result);
+            }
+
+            htst parse_htst_header(const utility::string_t& value)
+            {
+                htst result;
+                auto directives = parse_directives_header(value);
+                const auto max_age = std::find_if(directives.begin(), directives.end(), [](const ptoken_param& param) { return boost::algorithm::iequals(param.first, U("max-age")); });
+                if (directives.end() != max_age) result.max_age = utility::istringstreamed(max_age->second, 0);
+                const auto includeSubDomains = std::find_if(directives.begin(), directives.end(), [](const ptoken_param& param) { return boost::algorithm::iequals(param.first, U("includeSubDomains")); });
+                if (directives.end() != includeSubDomains) result.includeSubDomains = true;
+                return result;
             }
         }
 
--- a/Development/cpprest/http_utils.h
+++ b/Development/cpprest/http_utils.h
@@ -133,6 +133,19 @@ namespace web
             utility::string_t make_ptokens_header(const ptokens& values);
             ptokens parse_ptokens_header(const utility::string_t& value);
 
+            // Strict-Transport-Security = "Strict-Transport-Security" ":"
+            // directives           = [directive] *(";"[directive])
+            // directive            = directive-name["=" directive-value]
+            // directive-name       = token
+            // directive-value      = token / quoted - string
+            // E.g Strict-Transport-Security uuses this format
+            // See https://datatracker.ietf.org/doc/html/rfc6797#section-6.1
+
+            typedef std::pair<token, utility::string_t> directive;
+            typedef std::vector<directive> directives;
+
+            utility::string_t make_directives_header(const directives& values);
+            directives parse_directives_header(const utility::string_t& value);
 
             namespace header_names
             {
@@ -143,6 +156,10 @@ namespace web
                 // Resource Timing 2
                 // See https://www.w3.org/TR/resource-timing-2/#sec-timing-allow-origin
                 const web::http::http_headers::key_type timing_allow_origin{ _XPLATSTR("Timing-Allow-Origin") };
+
+                // Strict Transport Security
+                // See https://datatracker.ietf.org/doc/html/rfc6797#section-6.1
+                const web::http::http_headers::key_type strict_transport_security{ _XPLATSTR("Strict-Transport-Security") };
             }
 
             struct timing_metric
@@ -162,6 +179,20 @@ namespace web
 
             utility::string_t make_timing_header(const timing_metrics& values);
             timing_metrics parse_timing_header(const utility::string_t& value);
+
+            struct htst
+            {
+                int32_t max_age; // seconds
+                bool includeSubDomains;
+                htst(int32_t max_age = 0, bool includeSubDomains = false) : max_age(max_age), includeSubDomains(includeSubDomains) {}
+
+                auto tied() const -> decltype(std::tie(max_age, includeSubDomains)) { return std::tie(max_age, includeSubDomains); }
+                friend bool operator==(const htst& lhs, const htst& rhs) { return lhs.tied() == rhs.tied(); }
+                friend bool operator!=(const htst& lhs, const htst& rhs) { return !(lhs == rhs); }
+            };
+
+            utility::string_t make_hsts_header(const htst& value);
+            htst parse_htst_header(const utility::string_t& value);
         }
 
         // Determine whether http_request::reply() has been called already
--- a/Development/cpprest/http_utils.h
+++ b/Development/cpprest/http_utils.h
@@ -180,19 +180,19 @@ namespace web
             utility::string_t make_timing_header(const timing_metrics& values);
             timing_metrics parse_timing_header(const utility::string_t& value);
 
-            struct htst
+            struct hsts
             {
                 int32_t max_age; // seconds
                 bool includeSubDomains;
-                htst(int32_t max_age = 0, bool includeSubDomains = false) : max_age(max_age), includeSubDomains(includeSubDomains) {}
+                hsts(int32_t max_age = 0, bool includeSubDomains = false) : max_age(max_age), includeSubDomains(includeSubDomains) {}
 
                 auto tied() const -> decltype(std::tie(max_age, includeSubDomains)) { return std::tie(max_age, includeSubDomains); }
-                friend bool operator==(const htst& lhs, const htst& rhs) { return lhs.tied() == rhs.tied(); }
-                friend bool operator!=(const htst& lhs, const htst& rhs) { return !(lhs == rhs); }
+                friend bool operator==(const hsts& lhs, const hsts& rhs) { return lhs.tied() == rhs.tied(); }
+                friend bool operator!=(const hsts& lhs, const hsts& rhs) { return !(lhs == rhs); }
             };
 
-            utility::string_t make_hsts_header(const htst& value);
-            htst parse_htst_header(const utility::string_t& value);
+            utility::string_t make_hsts_header(const hsts& value);
+            hsts parse_hsts_header(const utility::string_t& value);
         }
 
         // Determine whether http_request::reply() has been called already
--- a/Development/nmos-cpp-node/config.json
+++ b/Development/nmos-cpp-node/config.json
@@ -236,5 +236,16 @@
     //"system_interval_min": 3600,
     //"system_interval_max": 3660,
 
+    // hsts_max_age [registry, node]: used for the Strict-Transport-Security HTTP response header's max-age field to enforce the HSTS Policy is to be used for the nth number of seconds; default is 365 days
+    // using value 0 to disable the Strict-Transport-Security HSTS Policy, HSTS is also disable when server_secure is switched off
+    // See https://datatracker.ietf.org/doc/html/rfc6797#section-6.1.1
+    //"hsts_max_age": 31536000,
+
+    // hsts_include_sub_domains [registry, node]: used for the Strict-Transport-Security HTTP response header's includeSubDomains field for signaling the client that the
+    // HSTS Policy applies to this HSTS Host as well as any subdomains of the host's domain name
+    // hsts_include_sub_domains will only be valid if and only if server_secure is switched on and hsts_max_age is set to > 0
+    // See https://datatracker.ietf.org/doc/html/rfc6797#section-6.1.2
+    //"hsts_include_sub_domains": false,
+
     "don't worry": "about trailing commas"
 }
--- a/Development/nmos-cpp-node/config.json
+++ b/Development/nmos-cpp-node/config.json
@@ -247,5 +247,13 @@
     // See https://datatracker.ietf.org/doc/html/rfc6797#section-6.1.2
     //"hsts_include_sub_domains": false,
 
+    // ocsp_interval_min/ocsp_interval_max [registry, node]: used to request for certificate status (OCSP response); default is about one hour
+    // Note: if the half life of the server cerificate expiry time is shorter, then the ocsp_interval_min/max will be overridden by it
+    //"ocsp_interval_min": 3600,
+    //"ocsp_interval_max": 3660,
+
+    // ocsp_request_max [registry, node]: timeout for interactions with the OCSP API
+    //"ocsp_request_max": 30,
+
     "don't worry": "about trailing commas"
 }
--- a/Development/nmos/api_utils.h
+++ b/Development/nmos/api_utils.h
@@ -124,14 +124,14 @@ namespace nmos
     void set_error_reply(web::http::http_response& res, web::http::status_code code, const std::exception& debug);
 
     // add handler to set appropriate response headers, and error response body if indicated - call this only after adding all others!
-    void add_api_finally_handler(web::http::experimental::listener::api_router& api, const web::http::experimental::htst& htst, slog::base_gate& gate);
+    void add_api_finally_handler(web::http::experimental::listener::api_router& api, const web::http::experimental::hsts& hsts, slog::base_gate& gate);
 
     // modify the specified API to handle all requests (including CORS preflight requests via "OPTIONS") and attach it to the specified listener
-    void support_api(web::http::experimental::listener::http_listener& listener, web::http::experimental::listener::api_router api, const web::http::experimental::htst& htst, slog::base_gate& gate);
+    void support_api(web::http::experimental::listener::http_listener& listener, web::http::experimental::listener::api_router api, const web::http::experimental::hsts& hsts, slog::base_gate& gate);
 
     // construct an http_listener on the specified address and port, modifying the specified API to handle all requests
     // (including CORS preflight requests via "OPTIONS")
-    web::http::experimental::listener::http_listener make_api_listener(bool secure, const utility::string_t& host_address, int port, web::http::experimental::listener::api_router api, web::http::experimental::listener::http_listener_config config, const web::http::experimental::htst& htst, slog::base_gate& gate);
+    web::http::experimental::listener::http_listener make_api_listener(bool secure, const utility::string_t& host_address, int port, web::http::experimental::listener::api_router api, web::http::experimental::listener::http_listener_config config, const web::http::experimental::hsts& hsts, slog::base_gate& gate);
 
     // construct an http_listener on the specified port, modifying the specified API to handle all requests
     // (including CORS preflight requests via "OPTIONS")
@@ -183,7 +183,7 @@ namespace nmos
         web::http::experimental::listener::route_handler make_api_version_handler(const std::set<api_version>& versions, slog::base_gate& gate);
 
         // make handler to set appropriate response headers, and error response body if indicated
-        web::http::experimental::listener::route_handler make_api_finally_handler(const web::http::experimental::htst& htst, slog::base_gate& gate);
+        web::http::experimental::listener::route_handler make_api_finally_handler(const web::http::experimental::hsts& hsts, slog::base_gate& gate);
     }
 }
 
--- a/Development/nmos/api_utils.h
+++ b/Development/nmos/api_utils.h
@@ -124,20 +124,20 @@ namespace nmos
     void set_error_reply(web::http::http_response& res, web::http::status_code code, const std::exception& debug);
 
     // add handler to set appropriate response headers, and error response body if indicated - call this only after adding all others!
-    void add_api_finally_handler(web::http::experimental::listener::api_router& api, slog::base_gate& gate);
+    void add_api_finally_handler(web::http::experimental::listener::api_router& api, const web::http::experimental::htst& htst, slog::base_gate& gate);
 
     // modify the specified API to handle all requests (including CORS preflight requests via "OPTIONS") and attach it to the specified listener
-    void support_api(web::http::experimental::listener::http_listener& listener, web::http::experimental::listener::api_router api, slog::base_gate& gate);
+    void support_api(web::http::experimental::listener::http_listener& listener, web::http::experimental::listener::api_router api, const web::http::experimental::htst& htst, slog::base_gate& gate);
 
     // construct an http_listener on the specified address and port, modifying the specified API to handle all requests
     // (including CORS preflight requests via "OPTIONS")
-    web::http::experimental::listener::http_listener make_api_listener(bool secure, const utility::string_t& host_address, int port, web::http::experimental::listener::api_router api, web::http::experimental::listener::http_listener_config config, slog::base_gate& gate);
+    web::http::experimental::listener::http_listener make_api_listener(bool secure, const utility::string_t& host_address, int port, web::http::experimental::listener::api_router api, web::http::experimental::listener::http_listener_config config, const web::http::experimental::htst& htst, slog::base_gate& gate);
 
     // construct an http_listener on the specified port, modifying the specified API to handle all requests
     // (including CORS preflight requests via "OPTIONS")
     inline web::http::experimental::listener::http_listener make_api_listener(int port, web::http::experimental::listener::api_router api, web::http::experimental::listener::http_listener_config config, slog::base_gate& gate)
     {
-        return make_api_listener(false, web::http::experimental::listener::host_wildcard, port, api, config, gate);
+        return make_api_listener(false, web::http::experimental::listener::host_wildcard, port, api, config, {}, gate);
     }
 
     // construct a websocket_listener on the specified address and port - captures handlers by reference!
@@ -183,7 +183,7 @@ namespace nmos
         web::http::experimental::listener::route_handler make_api_version_handler(const std::set<api_version>& versions, slog::base_gate& gate);
 
         // make handler to set appropriate response headers, and error response body if indicated
-        web::http::experimental::listener::route_handler make_api_finally_handler(slog::base_gate& gate);
+        web::http::experimental::listener::route_handler make_api_finally_handler(const web::http::experimental::htst& htst, slog::base_gate& gate);
     }
 }
 
--- a/Development/nmos/node_server.cpp
+++ b/Development/nmos/node_server.cpp
@@ -33,7 +33,7 @@ namespace nmos
             // Set up the APIs, assigning them to the configured ports
 
             const auto server_secure = nmos::experimental::fields::server_secure(node_model.settings);
-            const auto htst = server_secure ? web::http::experimental::htst{ nmos::experimental::fields::hsts_max_age(node_model.settings), nmos::experimental::fields::hsts_include_sub_domains(node_model.settings) } : web::http::experimental::htst{};
+            const auto hsts = server_secure ? web::http::experimental::hsts{ nmos::experimental::fields::hsts_max_age(node_model.settings), nmos::experimental::fields::hsts_include_sub_domains(node_model.settings) } : web::http::experimental::hsts{};
 
             // Configure the Settings API
 
@@ -74,7 +74,7 @@ namespace nmos
                 const auto& host = !api_router.first.first.empty() ? api_router.first.first : web::http::experimental::listener::host_wildcard;
                 // map the configured client port to the server port on which to listen
                 // hmm, this should probably also take account of the address
-                node_server.http_listeners.push_back(nmos::make_api_listener(server_secure, host, nmos::experimental::server_port(api_router.first.second, node_model.settings), api_router.second, http_config, htst, gate));
+                node_server.http_listeners.push_back(nmos::make_api_listener(server_secure, host, nmos::experimental::server_port(api_router.first.second, node_model.settings), api_router.second, http_config, hsts, gate));
             }
 
             // Set up the handlers for each WebSocket API port
--- a/Development/nmos/node_server.cpp
+++ b/Development/nmos/node_server.cpp
@@ -32,6 +32,7 @@ namespace nmos
             // Set up the APIs, assigning them to the configured ports
 
             const auto server_secure = nmos::experimental::fields::server_secure(node_model.settings);
+            const auto htst = server_secure ? web::http::experimental::htst{ nmos::experimental::fields::hsts_max_age(node_model.settings), nmos::experimental::fields::hsts_include_sub_domains(node_model.settings) } : web::http::experimental::htst{};
 
             // Configure the Settings API
 
@@ -72,7 +73,7 @@ namespace nmos
                 const auto& host = !api_router.first.first.empty() ? api_router.first.first : web::http::experimental::listener::host_wildcard;
                 // map the configured client port to the server port on which to listen
                 // hmm, this should probably also take account of the address
-                node_server.http_listeners.push_back(nmos::make_api_listener(server_secure, host, nmos::experimental::server_port(api_router.first.second, node_model.settings), api_router.second, http_config, gate));
+                node_server.http_listeners.push_back(nmos::make_api_listener(server_secure, host, nmos::experimental::server_port(api_router.first.second, node_model.settings), api_router.second, http_config, htst, gate));
             }
 
             // Set up the handlers for each WebSocket API port
--- a/Development/nmos/node_server.cpp
+++ b/Development/nmos/node_server.cpp
@@ -10,6 +10,7 @@
 #include "nmos/model.h"
 #include "nmos/node_api.h"
 #include "nmos/node_behaviour.h"
+#include "nmos/ocsp_behaviour.h"
 #include "nmos/server.h"
 #include "nmos/server_utils.h"
 #include "nmos/settings_api.h"
@@ -65,7 +66,7 @@ namespace nmos
 
             // Set up the listeners for each HTTP API port
 
-            auto http_config = nmos::make_http_listener_config(node_model.settings, node_implementation.load_server_certificates, node_implementation.load_dh_param, gate);
+            auto http_config = nmos::make_http_listener_config(node_model.settings, node_server.ocsp_settings, node_implementation.load_server_certificates, node_implementation.load_dh_param, gate);
 
             for (auto& api_router : node_server.api_routers)
             {
@@ -78,7 +79,7 @@ namespace nmos
 
             // Set up the handlers for each WebSocket API port
 
-            auto websocket_config = nmos::make_websocket_listener_config(node_model.settings, node_implementation.load_server_certificates, node_implementation.load_dh_param, gate);
+            auto websocket_config = nmos::make_websocket_listener_config(node_model.settings, node_server.ocsp_settings, node_implementation.load_server_certificates, node_implementation.load_dh_param, gate);
             websocket_config.set_log_callback(nmos::make_slog_logging_callback(gate));
 
             for (auto& ws_handler : node_server.ws_handlers)
@@ -100,6 +101,7 @@ namespace nmos
             auto set_transportfile = node_implementation.set_transportfile;
             auto connection_activated = node_implementation.connection_activated;
             auto channelmapping_activated = node_implementation.channelmapping_activated;
+            auto load_server_certificates = node_implementation.load_server_certificates;
             node_server.thread_functions.assign({
                 [&, load_ca_certificates, registration_changed] { nmos::node_behaviour_thread(node_model, load_ca_certificates, registration_changed, gate); },
                 [&] { nmos::send_events_ws_messages_thread(events_ws_listener, node_model, events_ws_api.second, gate); },
@@ -114,6 +116,12 @@ namespace nmos
                 node_server.thread_functions.push_back([&, load_ca_certificates, system_changed] { nmos::node_system_behaviour_thread(node_model, load_ca_certificates, system_changed, gate); });
             }
 
+#if !defined(_WIN32) || defined(CPPREST_FORCE_HTTP_CLIENT_ASIO)
+            if (server_secure)
+            {
+                node_server.thread_functions.push_back([&, load_ca_certificates, load_server_certificates] { nmos::ocsp_behaviour_thread(node_model, node_server.ocsp_settings, load_ca_certificates, load_server_certificates, gate); });
+            }
+#endif
             return node_server;
         }
 
--- a/Development/nmos/ocsp_behaviour.cpp
+++ b/Development/nmos/ocsp_behaviour.cpp
@@ -8,7 +8,7 @@
 #include "nmos/ocsp_utils.h"
 #include "nmos/random.h"
 #include "nmos/slog.h"
-#include "nmos/ssl_utils.h"
+#include "ssl/ssl_utils.h"
 
 namespace nmos
 {
@@ -258,11 +258,11 @@ namespace nmos
                 // get the shorter expiry time from all the certificates
                 for (const auto& cert_data : certificate_chains_data)
                 {
-                    const auto expiry_time_ = nmos::experimental::certificate_expiry_from_now(utility::us2s(cert_data), 0.5);
+                    const auto expiry_time_ = ssl::experimental::certificate_expiry_from_now(utility::us2s(cert_data), 0.5);
                     expiry_time = expiry_time < 0 ? expiry_time_ : std::min(expiry_time, expiry_time_);
                 }
             }
-            catch (const nmos::experimental::ssl_exception& e)
+            catch (const ssl::experimental::ssl_exception& e)
             {
                 slog::log<slog::severities::error>(gate, SLOG_FLF) << "Get certificate expiry time with SSL error: " << e.what();
             }
--- a/Development/nmos/ocsp_behaviour.cpp
+++ b/Development/nmos/ocsp_behaviour.cpp
@@ -1,0 +1,402 @@
+#include "nmos/ocsp_behaviour.h"
+
+#if !defined(_WIN32) || defined(CPPREST_FORCE_HTTP_CLIENT_ASIO)
+#include "pplx/pplx_utils.h" // for pplx::complete_at
+#include "nmos/client_utils.h"
+#include "nmos/model.h"
+#include "nmos/ocsp_settings.h"
+#include "nmos/ocsp_utils.h"
+#include "nmos/random.h"
+#include "nmos/slog.h"
+#include "nmos/ssl_utils.h"
+
+namespace nmos
+{
+    namespace details
+    {
+        struct ocsp_shared_state
+        {
+            load_ca_certificates_handler load_ca_certificates;
+            web::uri base_uri;
+            bool ocsp_service_error;
+            std::vector<uint8_t> ocsp_request;
+
+            // how many seconds before next certificate status request
+            int next_request;
+
+            nmos::details::seed_generator seeder;
+            std::default_random_engine engine;
+            std::unique_ptr<web::http::client::http_client> client;
+
+            explicit ocsp_shared_state(load_ca_certificates_handler load_ca_certificates)
+                : load_ca_certificates(std::move(load_ca_certificates))
+                , ocsp_service_error(false)
+                , next_request(0)
+                , engine(seeder)
+            {}
+        };
+
+        void ocsp_behaviour_thread(nmos::model& model, nmos::experimental::ocsp_settings& ocsp_settings, load_ca_certificates_handler load_ca_certificates, load_server_certificates_handler load_client_certificate, slog::base_gate& gate);
+
+        int half_certificate_expiry_from_now(const std::vector<utility::string_t>& certificate_chains_data, slog::base_gate& gate);
+        std::vector<web::uri> get_ocsp_uris(const std::vector<utility::string_t>& certificates_chain_data, slog::base_gate& gate);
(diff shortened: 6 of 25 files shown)


Development/nmos/ocsp_behaviour.cpp
```diff
@@ -0,0 +1,402 @@
...
+        {
+            try
+            {
+                // make OCSP request from multi cert chains
+                const auto certificate_chains_data_ = boost::copy_range<std::vector<std::string>>(certificate_chains_data | boost::adaptors::transformed([](const utility::string_t& cert) { return utility::us2s(cert); }));
+                return nmos::experimental::make_ocsp_request(certificate_chains_data_);
+            }
+            catch (const nmos::experimental::ocsp_exception& e)
+            {
+                slog::log<slog::severities::error>(gate, SLOG_FLF) << "Create OCSP request with OCSP error: " << e.what();
```
garethsb, line 331:
```suggestion
                slog::log<slog::severities::error>(gate, SLOG_FLF) << "OCSP error while creating OCSP request: " << e.what();
```

Development/nmos/ocsp_behaviour.cpp
```diff
@@ -0,0 +1,402 @@
...
+            }
+            catch (const std::exception& e)
+            {
+                slog::log<slog::severities::error>(gate, SLOG_FLF) << "Create OCSP request with unexpected exception: " << e.what();
```
garethsb, line 335:
```suggestion
                slog::log<slog::severities::error>(gate, SLOG_FLF) << "Unexpected exception while creating OCSP request: " << e.what();
```

Development/nmos/ocsp_behaviour.cpp
```diff
@@ -0,0 +1,402 @@
...
+            }
+            catch (...)
+            {
+                slog::log<slog::severities::severe>(gate, SLOG_FLF) << "Create OCSP request with nexpected unknown exception";
```
garethsb, line 339:
```suggestion
                slog::log<slog::severities::severe>(gate, SLOG_FLF) << "Unexpected unknown exception while creating OCSP request";
```

Development/nmos/ocsp_behaviour.cpp
```diff
@@ -0,0 +1,402 @@
...
+            }
+            return (expiry_time < 0 ? 0 : expiry_time);
+        }
+
+        // construct a list of OCSP URIs from a list of server certificate chains
+        std::vector<web::uri> get_ocsp_uris(const std::vector<utility::string_t>& certificate_chains_data, slog::base_gate& gate)
+        {
+            std::vector<web::uri> ocsp_uris;
+
+            try
```
garethsb, line 285: Does this `try`...`catch` block belong in this function (and `make_ocsp_request`) or would it be better closer to 'the edge'?

Development/ssl/ssl_utils.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+        // calculate the number of seconds to expire with the given ratio
+        int certificate_expiry_from_now(const std::string& cert_data, double ratio)
+        {
+            const auto cert_info = cert_information(cert_data);
+            const auto now = time(NULL);
+            const auto from_now = difftime(cert_info.not_after, now);
+            return (int)(from_now > 0 ? from_now * ratio : 0);
+        }
+	}
```
garethsb, line 281: 👀

Development/ssl/ssl_utils.cpp
```diff
@@ -0,0 +1,282 @@
+#include "ssl/ssl_utils.h"
+
+#include <boost/algorithm/string.hpp> // for boost::split
+#include <openssl/err.h>
+#include <openssl/pem.h>
+
+namespace ssl
+{
+	namespace experimental
+	{
```
garethsb, line 10: Please replace tabs with 4 spaces!

Development/ssl/ssl_utils.cpp
```diff
@@ -0,0 +1,282 @@
...
+                    certs.push_back(cert_data.substr(start, end - start + end_delimiter.length()));
+                    start = end + end_delimiter.length();
+                }
+
+            } while (std::string::npos != start && std::string::npos != end);
+
+            return certs;
+        }
+
+        // calculate the number of seconds to expire with the given ratio
```
garethsb, line 273: This seems like a slightly odd function to me. Combining `cert_information`, `time` and `difftime` seems natural, but doing the `> 0`, `* ratio` and truncate to `int`, less so.

```suggestion
        // calculate the number of seconds until expiry with the given ratio
```

Development/nmos/ocsp_behaviour.cpp
```diff
@@ -0,0 +1,402 @@
...
+                , next_request(0)
+                , engine(seeder)
+            {}
+        };
+
+        void ocsp_behaviour_thread(nmos::model& model, nmos::experimental::ocsp_settings& ocsp_settings, load_ca_certificates_handler load_ca_certificates, load_server_certificates_handler load_client_certificate, slog::base_gate& gate);
+
+        int half_certificate_expiry_from_now(const std::vector<utility::string_t>& certificate_chains_data, slog::base_gate& gate);
+        std::vector<web::uri> get_ocsp_uris(const std::vector<utility::string_t>& certificates_chain_data, slog::base_gate& gate);
+        std::vector<uint8_t> make_ocsp_request(const std::vector<utility::string_t>& certificate_chains_data, slog::base_gate& gate);
```
garethsb, line 43: Why `_data` in the parameter names? How is a certificate distinct from its data?

Development/nmos/ocsp_utils.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+            return details::make_ocsp_request(issuer_data, cert_chains_data);
+        }
+
+        // send OCSP response in TLS handshake
+        bool send_ocsp_response(SSL* s, const std::vector<uint8_t>& ocsp_resp_data)
+        {
+            using ssl::experimental::BIO_ptr;
+
+            if(ocsp_resp_data.size() <= 0) { return false; }
```
garethsb, line 150:
```suggestion
            if (ocsp_resp_data.size() <= 0) { return false; }
```

Development/nmos/ocsp_behaviour.cpp
```diff
@@ -0,0 +1,402 @@
...
+                // get the shorter expiry time from all the certificates
+                for (const auto& cert_data : certificate_chains_data)
+                {
+                    const auto expiry_time_ = ssl::experimental::certificate_expiry_from_now(utility::us2s(cert_data), 0.5);
+                    expiry_time = expiry_time < 0 ? expiry_time_ : std::min(expiry_time, expiry_time_);
+                }
+            }
+            catch (const ssl::experimental::ssl_exception& e)
+            {
+                slog::log<slog::severities::error>(gate, SLOG_FLF) << "Get certificate expiry time with SSL error: " << e.what();
```
garethsb, line 267:
```suggestion
                slog::log<slog::severities::error>(gate, SLOG_FLF) << "SSL error while getting certificate expiry time: " << e.what();
```

Development/nmos/ocsp_behaviour.cpp
```diff
@@ -0,0 +1,402 @@
...
+            }
+            catch (const std::exception& e)
+            {
+                slog::log<slog::severities::error>(gate, SLOG_FLF) << "Get certificate expiry time with unexpected exception: " << e.what();
```
garethsb, line 271:
```suggestion
                slog::log<slog::severities::error>(gate, SLOG_FLF) << "Unexpected exception while getting certificate expiry time: " << e.what();
```

Development/nmos/ocsp_behaviour.cpp
```diff
@@ -0,0 +1,402 @@
...
+            }
+            catch (...)
+            {
+                slog::log<slog::severities::severe>(gate, SLOG_FLF) << "Get certificate expiry time with nexpected unknown exception";
```
garethsb, line 275:
```suggestion
                slog::log<slog::severities::severe>(gate, SLOG_FLF) << "Unexpected unknown exception while getting certificate expiry time";
```

Development/cpprest/http_utils.cpp
```diff
@@ -330,6 +330,111 @@ namespace web
...
+
+                directives result;
+                utility::string_t name;
+                for (auto c : value)
+                {
+                    switch (state)
+                    {
+                    case pre_param:
+                        if (U(';') == c) { state = pre_param_name; break; }
+                        if (U(' ') == c || U('\t') == c) { break; }
```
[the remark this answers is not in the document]
garethsb: Oh, thank goodness, [RFC 7230 makes line folding obsolete](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2), so forget about this point!

Development/cpprest/http_utils.cpp
```diff
@@ -330,6 +330,111 @@ namespace web
...
+                    param_value_quoted_string,
+                    param_value_quoted_string_escape
+                } state = pre_param_name;
+
+                // token      = 1*<any CHAR except CTLs or separators>
+                // separators = "(" | ")" | "<" | ">" | "@"
+                //            | "," | ";" | ":" | "\" | <">
+                //            | "/" | "[" | "]" | "?" | "="
+                //            | "{" | "}" | SP | HT
+                // see https://datatracker.ietf.org/doc/html/rfc2616#section-2.2
```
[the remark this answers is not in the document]
garethsb: Since RFC 2616 is obsoleted by RFC 7230, I thought it was worth checking the definition of `token` and `tchar` there, and for which we already had the `is_tchar` function above.
https://tools.ietf.org/html/rfc7230#section-3.2.6
I believe the two specifications are equivalent, so despite RFC 6797 referring to the older RFC 2616, would it be better to reuse the existing function?

Development/cpprest/http_utils.cpp
```diff
@@ -330,6 +330,111 @@ namespace web
...
+                    param_value_quoted_string_escape
+                } state = pre_param_name;
+
+                // token      = 1*<any CHAR except CTLs or separators>
+                // separators = "(" | ")" | "<" | ">" | "@"
+                //            | "," | ";" | ":" | "\" | <">
+                //            | "/" | "[" | "]" | "?" | "="
+                //            | "{" | "}" | SP | HT
+                // see https://tools.ietf.org/html/rfc2616#section-2.2
+                auto is_tchar = [](utility::char_t c)
```
garethsb, line 368: Since RFC 2616 is obsoleted by RFC 7230, I checked the definition of `token` and `tchar` there. Although it's expressed differently,  the result is the equivalent. So maybe you could get rid of this whole comment and function and reuse the `is_tchar` function above?

Development/ssl/ssl_utils.cpp
garethsb: Resolved but not yet pushed?

Development/nmos/server_utils.cpp
```diff
@@ -18,9 +20,26 @@ namespace nmos
...
+        // This callback is called when client includes a certificate status request extension in the TLS handshake
+        static int server_certificate_status_request(SSL* s, void* arg)
+        {
+            nmos::experimental::ocsp_settings* settings = (nmos::experimental::ocsp_settings*)arg;
+            nmos::experimental::send_ocsp_response(s, nmos::with_read_lock(*settings->mutex, [&] { return settings->ocsp_response; }));
+            return SSL_TLSEXT_ERR_OK;
+        }
+
+        // setup server certificate status callback when client includes a certificate status request extension in the TLS handshake
+        void make_server_certificate_status_handler(boost::asio::ssl::context& ctx, const nmos::experimental::ocsp_settings& settings)
```
garethsb, line 33: I think this function, which returns `void` ought to be called something like `set_server_certificate_status_handler` (not `make`), didn't it? Also... just wondering... did you consider moving this function into _nmos/ocsp_utils.h_ so that the interface is at the level of `boost::asio::ssl::context`, similar to `http_listener_config`?

Development/nmos/server_utils.cpp
```diff
@@ -76,6 +95,11 @@ namespace nmos
                     {
                         ctx.use_tmp_dh(boost::asio::buffer(dh_param.data(), dh_param.size()));
                     }
+
+#if !defined(_WIN32) || defined(CPPREST_FORCE_HTTP_CLIENT_ASIO)
```
garethsb, line 99: Unnecessary `#if`...`#endif` here because the whole function is already inside the same?

Development/ssl/ssl_utils.h
```diff
@@ -0,0 +1,46 @@
+#ifndef SSL_SSL_UTILS_H
+#define SSL_SSL_UTILS_H
+
+#include <memory>
+#include <stdexcept>
+#include <string>
+#include <vector>
+#include <openssl/x509v3.h> // for X509V3_EXT_conf_nid
```
garethsb, line 8: Hmm, `X509V3_EXT_conf_nid` doesn't appear in this header file, so that can't be it... `X509_free` maybe??

Development/ssl/ssl_utils.h
```diff
@@ -0,0 +1,46 @@
...
+        // get last openssl error
+        std::string last_openssl_error();
+
+        struct cert_info
+        {
+            std::string common_name;
+            std::string issuer_name;
+            time_t not_before;
+            time_t not_after;
+            std::vector<std::string> sans;
```
garethsb, line 33: Should we spell out `subject_alternative_names` if we're spelling out `common_name`?

Development/nmos/ocsp_behaviour.cpp
```diff
@@ -0,0 +1,378 @@
...
+                        slog::log<slog::severities::error>(gate, SLOG_FLF) << "OCSP error while initial OCSP operation: " << e.what();
+                    }
+                    catch (const std::exception& e)
+                    {
+                        slog::log<slog::severities::error>(gate, SLOG_FLF) << "Unexpected exception while initial OCSP operation: " << e.what();
+                    }
+                    catch (...)
+                    {
+                        slog::log<slog::severities::severe>(gate, SLOG_FLF) << "Unexpected unknown exception while initial OCSP operation";
+                    }
```
garethsb, line 135:
```suggestion
                    catch (const nmos::experimental::ocsp_exception& e)
                    {
                        slog::log<slog::severities::error>(gate, SLOG_FLF) << "OCSP error during initial OCSP operation: " << e.what();
                    }
                    catch (const std::exception& e)
                    {
                        slog::log<slog::severities::error>(gate, SLOG_FLF) << "Unexpected exception during initial OCSP operation: " << e.what();
                    }
                    catch (...)
                    {
                        slog::log<slog::severities::severe>(gate, SLOG_FLF) << "Unexpected unknown exception during initial OCSP operation";
                    }
```

Development/nmos/ocsp_behaviour.cpp
```diff
@@ -0,0 +1,378 @@
...
+
+                    // get the list of server certificates chain data
+                    const auto server_certificates = load_server_certificate();
+                    const auto server_certificate_chains = boost::copy_range<std::vector<utility::string_t>>(server_certificates | boost::adaptors::transformed([](const nmos::certificate& cert) { return cert.certificate_chain; }));
+
+                    try
+                    {
+                        // get OCSP URIs
+                        ocsp_uris = details::get_ocsp_uris(server_certificate_chains, gate);
```
garethsb, line 97: Possible bug 🐛 If `get_ocsp_uris` throws, will this thread spin in the `initial_ocsp_operation` state without increasing the `backoff`?

Development/nmos/api_utils.h
```diff
@@ -125,19 +125,22 @@ namespace nmos
...
+    web::http::experimental::listener::http_listener make_api_listener(bool secure, const utility::string_t& host_address, int port, web::http::experimental::listener::api_router api, web::http::experimental::listener::http_listener_config config, const web::http::experimental::hsts& hsts, slog::base_gate& gate);
 
     // construct an http_listener on the specified port, modifying the specified API to handle all requests
     // (including CORS preflight requests via "OPTIONS")
     inline web::http::experimental::listener::http_listener make_api_listener(int port, web::http::experimental::listener::api_router api, web::http::experimental::listener::http_listener_config config, slog::base_gate& gate)
     {
-        return make_api_listener(false, web::http::experimental::listener::host_wildcard, port, api, config, gate);
+        return make_api_listener(false, web::http::experimental::listener::host_wildcard, port, api, config, {}, gate);
```
garethsb, line 143: Won't using `{}`, i.e. default-constructed `hsts`, change the behaviour - inserting `Strict-Transport-Security: max-age=0` every time rather than omitting it?

Also applies to all the overloads in _nmos/api_utils.cpp_ too...

Development/cpprest/http_utils.cpp
```diff
@@ -330,6 +330,98 @@ namespace web
...
+                {
+                    switch (state)
+                    {
+                    case directive_name:
+                        result.push_back({ name, {} }); break;
+                    default:
+                        throw std::logic_error("unreachable code");
+                    }
+                }
+                return result;
```
garethsb, line 422: I think there are syntax errors that can reach this point, e.g. `foo=meow;bar="purr` (no closing `"`), which we should probably reject? This also applies to `parse_ptokens_header` of course...

Development/cpprest/http_utils.h
```diff
@@ -162,6 +179,20 @@ namespace web
 
             utility::string_t make_timing_header(const timing_metrics& values);
             timing_metrics parse_timing_header(const utility::string_t& value);
+
+            struct hsts
+            {
+                int32_t max_age; // seconds
```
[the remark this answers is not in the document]
garethsb: Resolved by #278.

Development/cpprest/http_utils.cpp
```diff
@@ -370,6 +475,33 @@ namespace web
...
+
+            utility::string_t make_hsts_header(const hsts& value)
+            {
+                directives result;
+
+                // invalid hsts default to HSTS disable
+                auto hsts_ = value;
+                if (hsts_.max_age < 0) hsts_ = hsts{};
+
+                if (hsts_.max_age >= 0)
```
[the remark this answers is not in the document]
garethsb: Resolved by #278.

Development/nmos-cpp-node/config.json
```diff
@@ -236,5 +236,24 @@
     //"system_interval_min": 3600,
     //"system_interval_max": 3660,
 
+    // hsts_max_age [registry, node]: used for the Strict-Transport-Security HTTP response header's max-age field to enforce the HSTS Policy is to be used for the nth number of seconds; default is 365 days
+    // using value 0 to disable the Strict-Transport-Security HSTS Policy, HSTS is also disable when server_secure is switched off
```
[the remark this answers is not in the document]
garethsb: Resolved by #278.

Development/nmos-cpp-node/config.json
```diff
@@ -236,5 +236,24 @@
...
+    // See https://datatracker.ietf.org/doc/html/rfc6797#section-6.1.1
+    //"hsts_max_age": 31536000,
+
+    // hsts_include_sub_domains [registry, node]: used for the Strict-Transport-Security HTTP response header's includeSubDomains field for signaling the client that the
+    // HSTS Policy applies to this HSTS Host as well as any subdomains of the host's domain name
+    // hsts_include_sub_domains will only be valid if and only if server_secure is switched on and hsts_max_age is set to > 0
```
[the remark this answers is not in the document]
garethsb: Resolved by #278.

Development/nmos-cpp-node/config.json
```diff
@@ -236,5 +236,24 @@
...
+    // See https://datatracker.ietf.org/doc/html/rfc6797#section-6.1.2
+    //"hsts_include_sub_domains": false,
+
+    // ocsp_interval_min/ocsp_interval_max [registry, node]: used to request for certificate status (OCSP response); default is about one hour
+    // Note: if the half life of the server cerificate expiry time is shorter, then the ocsp_interval_min/max will be overridden by it
```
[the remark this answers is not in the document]
garethsb: Resolved by #278.

Development/nmos/ocsp_utils.cpp
```diff
@@ -0,0 +1,169 @@
...
+namespace nmos
+{
+	namespace experimental
+	{
+        typedef std::unique_ptr<OCSP_REQUEST, decltype(&OCSP_REQUEST_free)> OCSP_REQUEST_ptr;
+        typedef std::unique_ptr<OCSP_RESPONSE, decltype(&OCSP_RESPONSE_free)> OCSP_RESPONSE_ptr;
+
+        namespace details
+        {
+            // create OCSP request from list of server certificate chains data
```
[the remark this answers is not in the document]
garethsb: Resolved by @lo-simon in 0544da3d0decb14d9e84c4e0a746da647e1ad861 and #279.

Development/nmos/ocsp_settings.h
```diff
@@ -0,0 +1,34 @@
...
+#include "nmos/mutex.h"
+
+namespace nmos
+{
+    namespace experimental
+    {
+        struct ocsp_settings
+        {
+            // mutex to be used to protect the members of the settings from simultaneous access by multiple threads
+            std::shared_ptr<nmos::mutex> mutex;
```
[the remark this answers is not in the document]
garethsb: `shared_ptr` resolved by @lo-simon in 9eaeeccdd3ae3bfb9d1b1707baa19216d06f6f05 and #279.

But we still have some concerns about use of the mutex, alongside the model mutex, e.g. in _nmos/ocsp_behaviour.cpp_.

Development/nmos/api_utils.h
garethsb: Resolved by #278.

Development/cpprest/http_utils.cpp
garethsb: Resolved by #278.

Development/nmos/ocsp_utils.cpp
```diff
@@ -0,0 +1,188 @@
...
+        bool set_ocsp_response(SSL* ssl, const std::vector<uint8_t>& ocsp_response)
+        {
+            if (ocsp_response.empty()) return false;
+
+            auto buffer = OPENSSL_memdup(ocsp_response.data(), ocsp_response.size());
+            if (!buffer) return false;
+
+            return SSL_set_tlsext_status_ocsp_resp(ssl, buffer, (int)ocsp_response.size());
+        }
+
```
garethsb, line 177:
```suggestion
            if (0 == buffer) return false;

            return 0 != SSL_set_tlsext_status_ocsp_resp(ssl, buffer, (int)ocsp_response.size());
        }
```

Development/ssl/ssl_utils.cpp
garethsb: Still a tab here?

Development/ssl/ssl_utils.cpp
garethsb: Still tabs here?

Development/nmos/ocsp_utils.cpp
garethsb: (Mostly) resolved in 61158f0b81f92302ce0e47d731cf3208d2548830.

Development/nmos/ocsp_behaviour.h
```diff
@@ -0,0 +1,25 @@
...
+namespace nmos
+{
+    struct model;
+
+    namespace experimental
+    {
+        struct ocsp_state;
+    }
+
+    // callbacks from this function are called with the model locked, and may read or write directly to the model
```
garethsb, line 21: is this comment accurate? when are these handlers called?

Development/nmos/node_server.cpp
```diff
@@ -113,6 +117,16 @@ namespace nmos
                 node_server.thread_functions.push_back([&, load_ca_certificates, system_changed] { nmos::node_system_behaviour_thread(node_model, load_ca_certificates, system_changed, gate); });
             }
 
+// only implement communication with OCSP server if http_listener supports OCSP stapling
+// cf. preprocessor conditions in nmos::make_http_listener_config
+#if !defined(_WIN32) || defined(CPPREST_FORCE_HTTP_LISTENER_ASIO)
+            if (server_secure)
```
garethsb, line 123: Is there a way of disabling the OCSP implementation? Would that be a good idea?

garethsb:
What automated tests are possible for the OCSP functionality?

I know there is a single test case in the NMOS Testing Tool, provided by _testssl.sh_. Does this cover both TLS 1.2 and TLS 1.3? Is the test sufficient in other ways? Should we test with both one and two server certs or with other deployment differences?

How about unit tests?

Development/nmos/node_server.cpp
lo-simon: I thought this was really driven by the client, if the client is asking for it in the handshake we should always try to serve it.

We could add a new `ocsp_enable` in the settings to indicate whether to insert/or not of the `ocsp_behaviour_thread` and default the value to true.

Development/nmos/ocsp_utils.cpp
```diff
@@ -0,0 +1,194 @@
...
+        // construct an OCSP request from the specified list of server certificate chains
+        std::vector<uint8_t> make_ocsp_request(const std::vector<std::string>& certificate_chains)
+        {
+            if (certificate_chains.empty())
+            {
+                throw ocsp_exception("failed to make_ocsp_request: no server certificate chains");
+            }
+
+            // a minimal server certificate chain starts with the server's certificate, followed by the server's issuer certificate.
+            // for now, assuming all the server certificates are issued by the same issuer, let's get the issuer certificate from the 1st server certificate chain.
```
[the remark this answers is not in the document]
lo-simon: I never thought that there could be a difference between TLS 1.2. and TLS 1.3, for the OCSP request, do you have the RFC in hand?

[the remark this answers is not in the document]
lo-simon: I am assuming both certificates are issued from the same issuer. I will need to study how to handle OCSP for the different issuers.

ca -> immediate ca 1 ->certificate 1
ca -> immediate ca 2 ->certificate 2

Development/nmos/ocsp_behaviour.h
garethsb: You're saying all other calls of these two handlers are "protected" by a read lock on the model need? I wonder if that's consistent/necessary/good idea or whether would be more natural to call with it unlocked?

In any case, I suspect the content is wrong about writes being OK...
Ah, I also just spotted an inconsistent space before `&` on next line...

Development/nmos/ocsp_utils.cpp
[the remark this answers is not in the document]
garethsb: Regarding TLS 1 3, RFC 8446 says:

> It also changes how Online Certificate Status Protocol (OCSP) messages are carried and therefore updates [[RFC6066](https://datatracker.ietf.org/doc/html/rfc6066)] and obsoletes [[RFC6961](https://datatracker.ietf.org/doc/html/rfc6961)] as
   described in [Section 4.4.2.1](https://datatracker.ietf.org/doc/html/rfc8446#section-4.4.2.1).

That obsoleted RFC 6961 is the spec for supporting multiple stapled certs in TLS 1.2. Since it sounds like you've already tested that, maybe OpenSSL is implementing that for you? Probably it implements the TLS 1.3 way too. We just need tests 😁

Development/nmos/ocsp_behaviour.cpp
```diff
@@ -0,0 +1,366 @@
...
+                        if (ocsp_uris.size())
+                        {
+                            mode = ocsp_behaviour;
+
+                            // extract the shortest half certificate expiry time from all server certificates
+                            state.next_request = details::half_certificate_expiry_from_now(server_certificate_chains, gate);
+
+                            // construct an OCSP request with the server certificates
+                            state.ocsp_request = details::make_ocsp_request(server_certificate_chains, gate);
+                            if (state.ocsp_request.size())
```
garethsb, line 106:
```suggestion
                            if (!state.ocsp_request.empty())
```

P.S. @lo-simon, what does it mean if it's empty? If that's unexpected, should we log something?

Development/nmos/ocsp_behaviour.h
lo-simon: Good point, I was just trying to keep it consistent, but is it really necessary, maybe not. Then we will have to check all other threads, behaviour is the same too or not!

> Ah, I also just spotted an inconsistent space before & on next line...

where?

Development/nmos/ocsp_utils.cpp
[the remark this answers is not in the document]
lo-simon: At the moment in nmos-cpp settings, we can only allow for 1 root CA.

Development/ssl/ssl_utils.cpp
```diff
@@ -0,0 +1,282 @@
...
+                auto two_digits_to_uint = [&]()
+                {
+                    unsigned n = 10 * (*s++ - '0');
+                    return n + (*s++ - '0');
+                };
+
+                uint32_t year;
+                switch (time->type)
+                {
+                    // https://www.rfc-editor.org/rfc/rfc5280#section-4.1.2.5.1
```
garethsb, line 159:
```suggestion
                // see https://tools.ietf.org/html/rfc5280#section-4.1.2.5.1
```

Development/ssl/ssl_utils.cpp
```diff
@@ -0,0 +1,282 @@
...
+                OPENSSL_free(name);
+
+                // exmaple subject format
+                // e.g. subject=/DC=Example/CN=api.example.com
+                const std::string common_name_prefix{ "CN=" };
+                std::vector<std::string> tokens;
+                boost::split(tokens, subject, boost::is_any_of("/"));
+                auto found_common_name_token = std::find_if(tokens.begin(), tokens.end(), [&common_name_prefix](const std::string& token)
+                    {
+                        return std::string::npos != token.find(common_name_prefix);
```
garethsb, line 32: Shouldn't this check that the token _starts_ with `CN=`, not just _contains_ it?

Development/ssl/ssl_utils.cpp
```diff
@@ -0,0 +1,282 @@
...
+                auto subject_name = X509_get_subject_name(x509);
+                if (!subject_name)
+                {
+                    throw ssl_exception("failed to get subject: X509_get_subject_name failure: " + last_openssl_error());
+                }
+                auto name = X509_NAME_oneline(subject_name, NULL, 0);
+                std::string subject(name);
+                OPENSSL_free(name);
+
+                // exmaple subject format
```
garethsb, line 25:
```suggestion
                // example subject format
```

Development/ssl/ssl_utils.h
```diff
@@ -0,0 +1,46 @@
...
+        struct certificate_info
+        {
+            std::string common_name;
+            std::string issuer_name;
+            time_t not_before;
+            time_t not_after;
+            std::vector<std::string> subject_alternative_names;
+        };
+        // get certificate information, such as expire date, it is represented as the number of seconds from 1970-01-01T0:0:0Z as measured in UTC
+        certificate_info certificate_information(const std::string& certificate);
```
garethsb, line 36: This (or the `details` functions it uses) needs unit tests, I think?

Development/ssl/ssl_utils.h
```diff
@@ -0,0 +1,46 @@
...
+
+        // split certificate chain to list of certificates
+        std::vector<std::string> split_certificate_chain(const std::string& certificate_chain);
```
garethsb, line 39: Does this need unit tests?

Development/ssl/ssl_utils.h
```diff
@@ -0,0 +1,46 @@
...
+
+        // calculate the number of seconds to expire with the given ratio
+        double certificate_expiry_from_now(const std::string& cert, double ratio = 1.0);
```
garethsb, line 42: This is a `certificate` not a `certificate_chain`, yes?

```suggestion
        double certificate_expiry_from_now(const std::string& certificate, double ratio = 1.0);
```

Development/nmos/ocsp_behaviour.cpp
```diff
@@ -0,0 +1,366 @@
...
+
+            // start a background task to continously request certificate status on a given interval
+            return pplx::do_while([=, &model, &ocsp_state, &state, &gate]
+            {
+                auto request_interval = std::chrono::seconds(0);
+                if (state.base_uri == state.client->base_uri())
+                {
+                    auto interval = std::uniform_int_distribution<>(
+                        std::min(ocsp_interval_min, (state.next_request > 0.0 ? (int)state.next_request : ocsp_interval_min)),
+                        std::min(ocsp_interval_max, (state.next_request > 0.0 ? (int)state.next_request : ocsp_interval_max)))(state.engine);
```
garethsb, line 212: I find this slightly easier to read. What do you think?

```suggestion
                        state.next_request > 0.0 ? std::min((int)state.next_request, ocsp_interval_min) : ocsp_interval_min,
                        state.next_request > 0.0 ? std::min((int)state.next_request, ocsp_interval_max) : ocsp_interval_max)(state.engine);
```

Development/nmos/ocsp_behaviour.cpp
lo-simon: I think we can actually remove the 

line 97
`if (ocsp_uris.size())`
and line106
`if (state.ocsp_request.size())`
as `get_ocsp_uris` and `make_ocsp_request` have already throw for empty
which should be logged in the `try...catch` block.

And that will make the code clear too.

Good point what should we do if there is no OCSP server found in the certificates, then there is no point to have the `ocsp_behaviour_thread` running unless certificates can be changed dynamically.

Development/nmos/ocsp_utils.cpp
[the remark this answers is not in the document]
garethsb: I don't think that's true... we allow for 1 ca certificate _file_, but it may contain multiple authorities?

Development/ssl/ssl_utils.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+                uint32_t year;
+                switch (time->type)
+                {
+                    // https://www.rfc-editor.org/rfc/rfc5280#section-4.1.2.5.1
+                case V_ASN1_UTCTIME: // YYMMDDHHMMSSZ
+                    year = two_digits_to_uint();
+                    year += year < 50 ? 2000 : 1900;
+                    break;
+                    // https://www.rfc-editor.org/rfc/rfc5280#section-4.1.2.5.2
```
garethsb, line 164:
```suggestion
                // https://tools.ietf.org/html/rfc5280#section-4.1.2.5.2
```

Development/nmos/ocsp_utils.cpp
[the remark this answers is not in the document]
lo-simon: What am I thinking, the Root CA is for verifying the other server's certificate, which we are making calls to. Ignore my previous message.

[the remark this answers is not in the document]
garethsb: Ah, that's true. The server certs are likely to be rooted at (one of) the same CAs, 
but maybe that's beside the point like you say.

Also... I had forgotten that although we can be configured with multiple server certs with different issuers, actually we can only send one of those chains to a client... which one should be selected by OpenSSL based on what algorithms the client says it supports, etc. See e.g. https://tools.ietf.org/html/rfc8446#section-4.4.2.2

So that leads to interesting question... don't we need to know which one, to decide which one to fetch and return the OCSP response for?! Is there some way that OpenSSL tells us?!

Development/nmos/node_server.cpp
lo-simon: Yes, we can move the `ocsp_behaviour_thread` to the node implementation and when an OCSP response is received at the `ocsp_behaviour_thread`, then it executes the on_receive_ocsp_response callback on nmos to cache the OCSP response.

Callback as

```
namespace nmos
{
    namespace experimental
    {
        struct ocsp_state;
    }

    // callback to supply OCSP response in DER format
    // this callback is executed when receiving the OCSP response
    // this callback should not throw exceptions
    typedef std::function<void(const std::vector<uint8_t>&)> ocsp_response_handler;

    // construct callback to receive OCSP response
    ocsp_response_handler make_ocsp_response_handler(nmos::experimental::ocsp_state& ocsp_state, slog::base_gate& gate);
}
```

garethsb: I was thinking sorta like this...

```C++
namespace nmos
{
    namespace experimental
    {
        struct ocsp_state;
    }

    // callback to return OCSP response in DER format
    // this callback is executed for each new TLS connection
    // this callback should not throw exceptions
    typedef std::function<std::vector<uint8_t>()> ocsp_response_handler;

    // construct callback to return OCSP response
    ocsp_response_handler make_ocsp_response_handler(nmos::experimental::ocsp_state& ocsp_state, slog::base_gate& gate);
}
```

Development/nmos/ocsp_behaviour.h
```diff
@@ -0,0 +1,25 @@
...
+    void ocsp_behaviour_thread(nmos::model& model, nmos::experimental::ocsp_state& ocsp_state, load_ca_certificates_handler load_ca_certificates, load_server_certificates_handler load_server_certificates, slog::base_gate & gate);
```
garethsb, line 22:
```suggestion
    void ocsp_behaviour_thread(nmos::model& model, nmos::experimental::ocsp_state& ocsp_state, load_ca_certificates_handler load_ca_certificates, load_server_certificates_handler load_server_certificates, slog::base_gate& gate);
```

Development/nmos/ocsp_utils.cpp
```diff
@@ -102,6 +109,20 @@ namespace nmos
                 return nmos::experimental::set_ocsp_response(ssl, nmos::with_read_lock(ocsp_state.mutex, [&] { return ocsp_state.ocsp_response; })) ? SSL_TLSEXT_ERR_OK : SSL_TLSEXT_ERR_NOACK;
             }
 #endif
+
+            // get the certificate at position index in the server certificate chain
+            std::string get_certificate_at(const std::string& certificate_chain, size_t index)
```
garethsb, line 114: I don't like this function because it hides an expensive operation.

Development/nmos/ocsp_utils.cpp
```diff
@@ -16,58 +16,65 @@ namespace nmos
 
         namespace details
         {
-            // construct OCSP request using the issuer certificate and the list of server certificates
+            typedef std::map<std::string, std::vector<std::string>> issuer_certificate_vs_server_certifiactes_map;
```
garethsb, line 19: Typo "certifiactes" appears all over the place.

Development/nmos/ocsp_utils.cpp
```diff
@@ -16,58 +16,65 @@ namespace nmos
...
+                        }
+                        X509_ptr server_x509(PEM_read_bio_X509_AUX(server_bio.get(), NULL, NULL, NULL), &X509_free);
+                        if (!server_x509)
+                        {
+                            throw ocsp_exception("failed to make_ocsp_request while loading server certificate: PEM_read_bio_X509_AUX failure: " + ssl::experimental::last_openssl_error());
+                        }
+
+                        auto const cert_id_md = EVP_sha1();
+
+                        // creates and returns a new OCSP_CERTID structure using message digest dgst for certificate subject with issuer
```
garethsb, line 68: Copied sentence from man page didn't quite make sense to me...

```suggestion
                        // creates and returns a new OCSP_CERTID structure using SHA1 message digest for the server certificate with the issuer
```

Development/nmos/ocsp_utils.cpp
lo-simon: Fixed

Development/ssl/ssl_utils.cpp
```diff
@@ -0,0 +1,324 @@
...
+                const auto relative_distinguished_name_sequence = get_relative_distinguished_name_sequence(x509_name);
+
+                std::vector<std::string> relative_distinguished_names;
+                boost::split(relative_distinguished_names, relative_distinguished_name_sequence, boost::is_any_of(","));
+
+                return relative_distinguished_names;
+            }
+
+            // get the Attribute Type And Values from the given X509 Distinguished Name
+            // see https://www.rfc-editor.org/rfc/rfc2253#section-2.2
```
garethsb, line 47:
```suggestion
            // see https://tools.ietf.org/html/rfc2253#section-2.2
```

Development/ssl/ssl_utils.cpp
```diff
@@ -0,0 +1,324 @@
...
+                int bytes_read;
+                while ((bytes_read = BIO_gets(bio.get(), &buffer[0], sizeof(buffer))) > 0)
+                {
+                    result.append(buffer);
+                }
+                return result;
+            }
+
+            // get the Relative Distinguished Names from the given X509 Distinguished Name
+            // see https://www.rfc-editor.org/rfc/rfc2253#section-2.1
```
garethsb, line 35:
```suggestion
            // see https://tools.ietf.org/html/rfc2253#section-2.1
```

Development/ssl/ssl_utils.cpp
```diff
@@ -0,0 +1,324 @@
...
+#include <openssl/pem.h>
+
+namespace ssl
+{
+    namespace experimental
+    {
+        namespace details
+        {
+            // get the Relative Distinguished Name Sequence in the format specified in RFC2253
+            // see https://www.rfc-editor.org/rfc/rfc2253#section-2
```
garethsb, line 14:
```suggestion
            // see https://tools.ietf.org/html/rfc2253#section-2
```

Development/ssl/ssl_utils.cpp
```diff
@@ -0,0 +1,324 @@
...
+#include <openssl/err.h>
+#include <openssl/pem.h>
+
+namespace ssl
+{
+    namespace experimental
+    {
+        namespace details
+        {
+            // get the Relative Distinguished Name Sequence in the format specified in RFC2253
```
garethsb, line 13:
```suggestion
            // get the Relative Distinguished Name Sequence in the format specified in RFC 2253
```

Development/ssl/ssl_utils.cpp
```diff
@@ -0,0 +1,324 @@
...
+
+            // get the Attribute Value from the given X509 Distinguished Name with the Attribyte Type
+            // see https://www.rfc-editor.org/rfc/rfc2253#section-2.3
+            std::string get_attribute_value(X509_NAME* x509_name, const std::string& attribute_type)
+            {
+                const auto attribute_type_and_values = get_attribute_type_and_values(x509_name);
+
+                auto found = std::find_if(attribute_type_and_values.begin(), attribute_type_and_values.end(), [&attribute_type](const std::string& attribute_type_and_value)
+                {
+                    return std::string::npos != attribute_type_and_value.find(attribute_type + "=");
```
garethsb, line 75: `find` anywhere, or should it be at the start??

Development/nmos-cpp-node/node_implementation.cpp
```diff
@@ -1236,3 +1237,11 @@ nmos::experimental::node_implementation make_node_implementation(nmos::node_mode
         .on_validate_channelmapping_output_map(make_node_implementation_map_validator()) // may be omitted if not required
         .on_channelmapping_activated(make_node_implementation_channelmapping_activation_handler(gate));
 }
+
+// This constructs all the callbacks including OCSP stapling used to integrate the example device-specific underlying implementation
+// into the server instance for the NMOS Node.
+nmos::experimental::node_implementation make_node_implementation(nmos::node_model& model, nmos::experimental::ocsp_state& ocsp_state, slog::base_gate& gate)
+{
+    return make_node_implementation(model, gate)
+        .on_get_ocsp_response(nmos::make_ocsp_response_handler(ocsp_state, gate));
```
garethsb, line 1246: How about just doing this in _main.cpp_, to really show how separable OCSP is from the rest of the core NMOS bits? Could be a nice model for when we come to add Auth/Cert Prov in the same way?

Especially since we're adding `nmos::ocsp_behaviour_thread` in _main.cpp_...

Development/ssl/ssl_utils.h
```diff
@@ -27,7 +27,7 @@ namespace ssl
 
         struct certificate_info
         {
-            std::string common_name;
+            std::string subject_common_name;
             std::string issuer_name;
```
garethsb, line 31: Hmm, should this member be `issuer_common_name` too then?

garethsb (commented):
Very quick review of last two commits!

Development/nmos/server_utils.cpp
```diff
@@ -76,6 +81,13 @@ namespace nmos
                     {
                         ctx.use_tmp_dh(boost::asio::buffer(dh_param.data(), dh_param.size()));
                     }
+
+                    // set up server certificate status callback when client includes a certificate status request extension in the TLS handshake
+                    if (get_ocsp_response)
+                    {
+                        *ocsp_response = get_ocsp_response();
```
garethsb, line 88: Will this get moved into the variable?

Development/ssl/ssl_utils.cpp
```diff
@@ -0,0 +1,324 @@
...
+            }
+
+            // get the Relative Distinguished Names from the given X509 Distinguished Name
+            // see https://www.rfc-editor.org/rfc/rfc2253#section-2.1
+            inline std::vector<std::string> get_relative_distinguished_names(X509_NAME* x509_name)
+            {
+                const auto relative_distinguished_name_sequence = get_relative_distinguished_name_sequence(x509_name);
+
+                std::vector<std::string> relative_distinguished_names;
+                boost::split(relative_distinguished_names, relative_distinguished_name_sequence, boost::is_any_of(","));
```
garethsb, line 41: I think this needs to be more complicated to handle escaped characters, e.g. `\,`, see https://tools.ietf.org/html/rfc2253#section-5

Development/nmos/server_utils.cpp
lo-simon: yes

Development/cmake/NmosCppTest.cmake
```diff
@@ -110,7 +119,7 @@ source_group("mdns\\test\\Header Files" FILES ${NMOS_CPP_TEST_MDNS_TEST_HEADERS}
 source_group("nmos\\test\\Header Files" FILES ${NMOS_CPP_TEST_NMOS_TEST_HEADERS})
 source_group("pplx\\test\\Header Files" FILES ${NMOS_CPP_TEST_PPLX_TEST_HEADERS})
 source_group("rql\\test\\Header Files" FILES ${NMOS_CPP_TEST_RQL_TEST_HEADERS})
-source_group("sdp\\test\\Header Files" FILES ${NMOS_CPP_TEST_SDP_TEST_HEADERS})
+source_group("ssl\\test\\Header Files" FILES ${NMOS_CPP_TEST_SSL_TEST_HEADERS})
```
garethsb, line 122: Why get rid of the _sdp\test\Header Files_ group?

Development/ssl/test/ssl_test.cpp
```diff
@@ -0,0 +1,39 @@
...
+    {
+        std::string field;
+        std::string value;
+    };
+
+    const test tests[] =
+    {
+        { "C", "UK"},
+        { "O", "Disorganized Organization"},
+        { "CN", "Joe Bloggs"}
```
garethsb, line 22: In addition to these, I think you need some test values that contain characters that need escaping, at least comma, equal, plus, new line and backslash.

```suggestion
        { "C", "UK" },
        { "O", "Disorganized Organization" },
        { "CN", "Joe Bloggs" }
```

Development/nmos/ocsp_response_handler.h
```diff
@@ -0,0 +1,29 @@
...
+    {
+        struct ocsp_state;
+    }
+
+    // callback to return OCSP response in DER format
+    // this callback is executed for each new TLS connection
+    // this callback should not throw exceptions
+    typedef std::function<std::vector<uint8_t>()> ocsp_response_handler;
+
+    // construct callback to receive OCSP response
```
garethsb, line 25: Did you mean retrieve rather than receive?

```suggestion
    // construct callback to retrieve OCSP response
```

(Comment should also go in the cpp for consistency.)

Development/nmos/ocsp_behaviour.cpp
```diff
@@ -0,0 +1,362 @@
...
+        web::http::client::http_client_config make_ocsp_client_config(const nmos::settings& settings, load_ca_certificates_handler load_ca_certificates, slog::base_gate& gate)
+        {
+            auto config = nmos::make_http_client_config(settings, std::move(load_ca_certificates), gate);
+            config.set_timeout(std::chrono::seconds(nmos::experimental::fields::ocsp_request_max(settings)));
+            return config;
+        }
+
+        struct ocsp_service_exception {};
+
+        // make an asynchronously POST request on the OCSP server to get certificate status
```
garethsb, line 153:
```suggestion
        // make an asynchronous POST request on the OCSP server to get certificate status
```

Development/ssl/ssl_utils.h
```diff
@@ -0,0 +1,54 @@
...
+            time_t not_after;
+            std::vector<std::string> subject_alternative_names;
+        };
+        // get certificate information, such as expire date, it is represented as the number of seconds from 1970-01-01T0:0:0Z as measured in UTC
+        certificate_info get_certificate_info(const std::string& certificate);
+
+        // split certificate chain to a list of certificates
+        std::vector<std::string> split_certificate_chain(const std::string& certificate_chain);
+
+        // calculate the number of seconds to expire with the given ratio
```
garethsb, line 49:
```suggestion
        // calculate the specified proportion of the number of seconds until expiry of the specified certificate
```

Development/ssl/ssl_utils.h
```diff
@@ -0,0 +1,54 @@
...
+        double certificate_expiry_from_now(const std::string& certificate, double ratio = 1.0);
```
garethsb, line 50:
```suggestion
        double certificate_expiry_from_now(const std::string& certificate, double proportion = 1.0);
```

Development/ssl/ssl_utils.cpp
```diff
@@ -0,0 +1,325 @@
...
+                    certificates.push_back(certificate_chain.substr(start, end - start + end_delimiter.length()));
+                    start = end + end_delimiter.length();
+                }
+
+            } while (std::string::npos != start && std::string::npos != end);
+
+            return certificates;
+        }
+
+        // calculate the number of seconds until expiry with the given ratio
```
garethsb, line 316:
```suggestion
        // calculate the specified proportion of the number of seconds until expiry of the specified certificate
```

Development/ssl/ssl_utils.cpp
```diff
@@ -0,0 +1,325 @@
...
+        double certificate_expiry_from_now(const std::string& certificate, double ratio)
```
garethsb, line 317:
```suggestion
        double certificate_expiry_from_now(const std::string& certificate, double proportion)
```

Development/ssl/ssl_utils.cpp
```diff
@@ -0,0 +1,325 @@
...
+        {
+            const auto certificate_info = get_certificate_info(certificate);
+            const auto now = time(NULL);
+            const auto from_now = difftime(certificate_info.not_after, now);
+            return (from_now > 0.0 ? from_now * ratio : 0.0);
```
garethsb, line 322:
```suggestion
            return (from_now > 0.0 ? from_now * proportion : 0.0);
```

Development/ssl/ssl_utils.h
```diff
@@ -0,0 +1,54 @@
...
+        typedef std::unique_ptr<ASN1_TIME, decltype(&ASN1_STRING_free)> ASN1_TIME_ptr;
+
+        // get last openssl error
+        std::string last_openssl_error();
+
+        struct certificate_info
+        {
+            std::string subject_common_name;
+            std::string issuer_common_name;
+            time_t not_before;
```
garethsb, line 39:
```suggestion
            // not_before and not_after are the start and end of the certificate's validity
            // represented as the number of seconds in UTC since 1970-01-01T0:0:0Z
            time_t not_before;
```

Development/ssl/ssl_utils.h
```diff
@@ -0,0 +1,54 @@
...
+            time_t not_after;
+            std::vector<std::string> subject_alternative_names;
+        };
+        // get certificate information, such as expire date, it is represented as the number of seconds from 1970-01-01T0:0:0Z as measured in UTC
```
garethsb, line 43:
```suggestion

        // get certificate information, such as subject, issuer and validity
```

Development/ssl/ssl_utils.cpp
```diff
@@ -0,0 +1,325 @@
...
+
+        // get last openssl error
+        std::string last_openssl_error()
+        {
+            char buffer[1024] = { 0 };
+            ERR_error_string_n(ERR_get_error(), buffer, sizeof(buffer));
+            return buffer;
+        }
+
+        // get certificate information, such as expire date, it is represented as the number of seconds from 1970-01-01T0:0:0Z as measured in UTC
```
garethsb, line 231:
```suggestion
        // get certificate information, such as subject, issuer and validity
```

garethsb (commented):
Just comment tweaks

Development/ssl/ssl_utils.h
garethsb: Actually, now this returns `double` there is little point in the ratio/proportion parameter, we could just do `* 0.5` where we need to? Simpler all round...

Development/nmos/ocsp_behaviour.cpp
```diff
@@ -0,0 +1,362 @@
...
+                // let's select the next available OCSP server
+                state.ocsp_service_error = true;
+
+                model.notify();
+            });
+        }
+
+        double half_certificate_expiry_from_now(const std::vector<utility::string_t>& certificate_chains, slog::base_gate& gate)
+        {
+            double expiry_time = -1.0;
```
garethsb, line 257: Wouldn't this all be much simpler if you initialized this, and `next_request` to `(std::numeric_limits<double>::max)()`?

Development/ssl/ssl_utils.h
lo-simon: Okay, will checkin next.

Development/nmos/ocsp_behaviour.cpp
lo-simon: Yes, done

Development/nmos/ocsp_behaviour.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+            // start a background task to continously request certificate status on a given interval
+            return pplx::do_while([=, &model, &ocsp_state, &state, &gate]
+            {
+                auto request_interval = std::chrono::seconds(0);
+                if (state.base_uri == state.client->base_uri())
+                {
+                    auto interval = std::uniform_int_distribution<>(
+                        state.next_request > 0.0 ? std::min((int)state.next_request, ocsp_interval_min) : ocsp_interval_min,
+                        state.next_request > 0.0 ? std::min((int)state.next_request, ocsp_interval_max) : ocsp_interval_max)(state.engine);
```
garethsb, line 208: Could you get rid of the check outside the `std::min` calls here too, if you initialized `next_request` to `(std::numeric_limits<double>::max)()` rather than `0.0`?

BTW, to avoid issues if people don't want `NOMINMAX` on Windows, we usually write `(std::min)(...)`.

Development/nmos/ocsp_behaviour.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+        double half_certificate_expiry_from_now(const std::vector<utility::string_t>& certificate_chains, slog::base_gate& gate)
+        {
+            double expiry_time = (std::numeric_limits<double>::max)();
+            try
+            {
+                // get the shortest expiry time from all the certificates
+                for (const auto& certificate_chain : certificate_chains)
+                {
+                    const auto expiry_time_ = ssl::experimental::certificate_expiry_from_now(utility::us2s(certificate_chain)) * 0.5;
```
garethsb, line 263: `certificate_expiry_from_now` seems to assume (using `get_certificate_info`) that it is passed a single certificate, whereas here you pass it a certificate chain. Is that intentional? Is it only the validity of the first (server?) certificate that matters?

garethsb: I figured, so if that's what you wanted, maybe just add a comment here?

Development/nmos/ocsp_behaviour.cpp
lo-simon: Is that mean all the `std::min` and `std::max` should be written as 

e.g. in certificate_handlers.cpp
```
const auto size = std::min(private_key_files.size(), certificate_chain_files.size());
```
should be 
```
const auto size = (std::min)(private_key_files.size(), certificate_chain_files.size());
```

garethsb: Yep, looks like that one crept through review...

Development/ssl/ssl_utils.cpp
lo-simon: Fixed in the latest check-in

Development/nmos/ocsp_behaviour.cpp
```diff
@@ -0,0 +1,365 @@
...
+        {
+            auto config = nmos::make_http_client_config(settings, std::move(load_ca_certificates), gate);
+            config.set_timeout(std::chrono::seconds(nmos::experimental::fields::ocsp_request_max(settings)));
+            return config;
+        }
+
+        struct ocsp_service_exception {};
+
+        // make an asynchronous POST request on the OCSP server to get certificate status
+        // see https://specs.amwa.tv/bcp-003-03/releases/v1.0.0/docs/1.0._Certificate_Provisioning.html#certificate-request
```
garethsb, line 154: I don't think this is the right reference?
If it needs a reference at all, maybe this one?
```suggestion
        // see https://specs.amwa.tv/bcp-003-01/releases/v1.0.1/docs/Secure_Communication.html#certificate-management-client
```

Development/ssl/ssl_utils.cpp
```diff
@@ -0,0 +1,216 @@
...
+                    }
+                    else
+                    {
+                        // hmm, not supporting other type of subject alt name
+                    }
+                }
+                return subject_alternative_names;
+            }
+
+            // convert ANS.1 time to POSIX (UTC)
```
garethsb, line 55:
```suggestion
            // convert ASN.1 time to POSIX (UTC)
```

Development/ssl/ssl_utils.cpp
```diff
@@ -0,0 +1,216 @@
...
+                    tm.tm_year -= 1900;
+                    break;
+                // https://tools.ietf.org/html/rfc5280#section-4.1.2.5.2
+                case V_ASN1_GENERALIZEDTIME: // YYYYMMDDHHMMSSZ
+                    tm.tm_year = 100 * two_digits_to_uint();
+                    tm.tm_year += two_digits_to_uint();
+                    tm.tm_year -= 1900;
+                    break;
+                default:
+                    throw ssl_exception("failed to convert ASN1_TIME to UTC: invalid ANS.1 time type");
```
garethsb, line 98:
```suggestion
                    throw ssl_exception("failed to convert ASN1_TIME to UTC: invalid ASN.1 time type");
```

Development/ssl/ssl_utils.cpp
```diff
@@ -0,0 +1,216 @@
...
+                }
+                tm.tm_mon = two_digits_to_uint() - 1;
+                tm.tm_mday = two_digits_to_uint();
+                tm.tm_hour = two_digits_to_uint();
+                tm.tm_min = two_digits_to_uint();
+                tm.tm_sec = two_digits_to_uint();
+                if (*s != 'Z') { throw ssl_exception("failed to convert ASN1_TIME to UTC: invalid ANS.1 time format"); }
```
garethsb, line 105:
```suggestion
                if (*s != 'Z') { throw ssl_exception("failed to convert ASN1_TIME to UTC: invalid ASN.1 time format"); }
```

garethsb (changes requested):
Suggestions to fix typos

PR #277, opened 2022-10-10
OCSP patch 1

garethsb:
Tweak preprocessor conditions for code that are only used with the Boost.ASIO implementation of http_client, http_listener, websocket_client, websocket_listener


lo-simon (approved):
LGTM

PR #280, opened 2022-10-18
video/jxsv

garethsb:
JPEG XS support, per [[Work In Progress] BCP-006-01](https://specs.amwa.tv/bcp-006-01/branches/v1.0-dev).

I propose waiting until we add e.g. `video/H264` or similar before moving the common things out of _nmos/video_jxsv.h_ into common headers.

~The few missing/new ST 2110 format-specific parameters to add, and updates for ST 2110-xx:2022, can be done as a separate PR; I'm fed up waiting for SMPTE/IEEE.~ (Done.)

### DIFF
--- a/Development/cmake/NmosCppLibraries.cmake
+++ b/Development/cmake/NmosCppLibraries.cmake
@@ -799,6 +799,7 @@ set(NMOS_CPP_NMOS_SOURCES
     nmos/settings_api.cpp
     nmos/system_api.cpp
     nmos/system_resources.cpp
+    nmos/video_jxsv.cpp
     )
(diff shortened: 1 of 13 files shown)


garethsb:
@jonathan-r-thorpe, @lo-simon, with 85c2ca0105455e22e2c4d2ab4c53ab52ce51d9dd, I've now updated the SDP generation and parsing to cover all the previously ignored ST 2110 format-specific parameters as well as the updates for 2022 revisions of ST 2110-10/-20/-21/-22/-40. That makes this a much bigger changeset in total, so it may be easier to review commit-by-commit.

Issue #281, opened 2022-10-19
Add a VS2022 runner


cristian-recoseanu:
Would it be possible to add a VS2022 runner now that the option is available in GitHub actions (https://github.blog/changelog/2021-11-16-github-actions-windows-server-2022-with-visual-studio-2022-is-now-generally-available-on-github-hosted-runners/) and VS2022 is more widely available?

Thank you!

lo-simon:
Resolved by #282

PR #282, opened 2022-10-20
Switch one runner to windows-2022 to test VS 2022

garethsb:
#281.


garethsb:
Hm, Windows Server 2022 build succeeded and most tests worked fine. The BCP-003-01 tests showed:
```
* WARNING: Could not find an installation of 'testssl'. Some tests will be disabled.
```
And right at the top of the log:
```
[1;35mFatal error: No IPv4/IPv6 address(es) for "nmos-api.local" available[m
```

garethsb:
Hey, @lo-simon,

I just noticed that after your PR https://github.com/AMWA-TV/nmos-testing/pull/678 bumped the testing tool's requirement for _testssl.sh_ to v3.0.7, the nmos-cpp CI didn't get the same bump so I've just added it to this PR.

Before I did that, I thought we ought to have seen a message in the `testoutput` files like this:

```
 * WARNING: Version of 'testssl' does not match the expected '3.0.7'
```

Sure enough, it's there at the top of the file for the Ubuntu 22.04 runner for example.

But it's not there on the `windows-2019` one, which like the `windows-2022` one, has:

```
* WARNING: Could not find an installation of 'testssl'. Some tests will be disabled.
```

However, the BCP-003-01 test suite on the `windows-2019` runner showed PASSes!

I don't understand...

garethsb:
@lo-simon Seems we are now back where we were with this, ❌ BCP-003-01 can't run on [windows-2022](https://github.com/sony/nmos-cpp/actions/runs/3381451521#summary-9260609848) runner for some reason... 😕

garethsb:
Can this now work with latest nmos-testing without the monkey patch?

lo-simon:
@garethsb, yes it should now work without the monkey patch, can you roll it back, and re-sync your fork? If you don't have time, I can do it, but I will have to cherry-pick your changes and do it on a new PR.

garethsb:
@lo-simon Fingers crossed now https://github.com/AMWA-TV/nmos-testing/pull/832 is merged this simple change from one year ago finally works!!

PR #285, opened 2022-10-27
RFC 2253 parse_distinguished_name

garethsb:


garethsb:
We're going to use `X509_NAME_get_text_by_NID` instead!!

PR #286, opened 2022-10-31
Bump up testssl.sh to v3.0.7

lo-simon:
bump testssl.sh from v3.0.2 to v3.0.7


garethsb:
That ain't gonna do much until you rebuild the install script... see e4b538a6fe771482f9a5c2612690cc47093328f4 in #282

PR #288, opened 2022-11-01
Fix rql::details::rql_exception 

garethsb:
`rql::details::rql_exception` was previously thrown during evaluation when an unimplemented operator was encountered; if Query API subscriptions were created that used an unrecognized call-operator, this could result in an exception during subsequent `nmos::insert_resource_events` and therefore registrations to be rejected with a `501` Not Implemented, and uncaught exceptions when new WebSocket connections were opened for that subscription. This solves this by validating the call-operators immediately after the `rql::parse_query` so that such subscriptions are rejected (appropriately, with `501` Not Implemented), and (belt and braces) returning false, i.e. no match, if an exception is encountered during evaluation.

### DIFF
--- a/Development/nmos/logging_api.cpp
+++ b/Development/nmos/logging_api.cpp
@@ -36,16 +36,24 @@ namespace nmos
             return logging_api;
         }
 
+        static inline rql::extractor make_rql_extractor(const web::json::value& value)
+        {
+            return [&value](web::json::value& results, const web::json::value& key)
+            {
+                return web::json::extract(value.as_object(), results, key.as_string());
+            };
+        }
+
         bool match_logging_rql(const web::json::value& value, const web::json::value& query)
         {
-            return query.is_null() || rql::evaluator
+            try
             {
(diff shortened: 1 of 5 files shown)


garethsb:
Hmm, seems like a very spurious failure ❌ [IS-04-01 test_13](https://github.com/AMWA-TV/nmos-testing/blob/master/nmostesting/suites/IS0401Test.py#L794) on [ubuntu-20.04_avahi_multicast](https://github.com/sony/nmos-cpp/actions/runs/3381423160#summary-9260526412). I can't understand how it could be related to this PR...

lo-simon:
Very spurious failure ❌ [IS-04-01 test_13](https://github.com/AMWA-TV/nmos-testing/blob/master/nmostesting/suites/IS0401Test.py#L794) on [ubuntu-20.04_avahi_multicast](https://github.com/sony/nmos-cpp/actions/runs/3381423160#summary-9260526412), this is nothing related to this PR.

PR #296, opened 2022-11-10
Fix ASN1_TIME_to_time_t using an older version of openssl than 1.1.1

lo-simon:


garethsb:
Presumably this is to support an older version than even 1.0.2 which itself is already out of support, and it arrived in March 2015.

We should think about the fact that even 1.1.1 will be out of support after 7 September 2023...

lo-simon:
Yes, maybe we should start thinking about adding openssl 3.0 support. As long as we are still doing Ubuntu 14.04, we will have to continue to support 1.0.1f.

garethsb:
@lo-simon Did the Ubuntu 14.04 GitHub runner have a build or test step it failed before this PR??

If you hadn't just merged, I would have suggested you change the commit message and PR title to reflect 1.0.1f caused the trouble, not anything pre-1.1.0 as well...

PR #299, opened 2022-11-16
Workaround for Avahi target name with current Conan recipe

### LINKED ISSUE #298: CMake Error when using nmos-cpp conan package on Ubuntu 20.04
When importing nmos-cpp into a project using conan and CMake the following error occurs when configuring the project CMake cache.
```
CMake Error at cmake/NmosCppDependencies.cmake:267 (target_link_libraries):
  The link interface of target "DNSSD" contains:

    Avahi::compat-libdns_sd

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

Call Stack (most recent call first):
  CMakeLists.txt:21 (include)
```
I have been able to fix/work-around the issue by removing the avahi folder in .conan/data and manually installing the earliest revision of Avahi by doing:
`conan install avahi/0.8@#045ad530d68fcb08ffda15df2ad838ed`

The issue seems to be that the find_package(Avahi) (cmake/NmosCppDependencies.cpp:262) is finding the conan installed version of avahi, and it is not using the customised "third_party/cmake/FindAvahi.cmake". It is the custom file that defines the target "Avahi::compat-libdns_sd" and since that file is not being used, the target is not defined.

This docker file [Dockerfile.txt](https://github.com/sony/nmos-cpp/files/10021122/Dockerfile.txt) can be used to reproduce the error using the command "docker build . -t nmos-cpp-conan" (assuming Dockerfile is in current folder and .txt extension removed). Note that both zlib and openssl need to be explicitly defined in the conanfile.txt in order to fix version incompatibilities.

garethsb:
May resolve #298.

`avahi/0.8@#10417ed9b1382e1c2b6916db4247dc40` uses CMakeDeps and therefore the target namespace is `avahi::` not `Avahi::`.

@will-leathers-smith-sony, can you try this out? Unfortunately, this _won't_ be tested by GitHub Actions, since the local [_conanfile.txt_](https://github.com/sony/nmos-cpp/blob/master/Development/conanfile.txt) doesn't require the Conan `avahi` package, instead the GitHub Actions [build-test workflow](https://github.com/sony/nmos-cpp/blob/25ffd0b3a8b5a29f3bdf71fd46d3386da789bebb/.github/workflows/build-test.yml#L250) uses `apt-get` to install it.

### DIFF
--- a/Development/cmake/NmosCppDependencies.cmake
+++ b/Development/cmake/NmosCppDependencies.cmake
@@ -264,7 +264,12 @@ if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
             message(STATUS "Found Avahi version " ${Avahi_VERSION})
         endif()
 
-        target_link_libraries(DNSSD INTERFACE Avahi::compat-libdns_sd)
+        if(TARGET Avahi::compat-libdns_sd)
+            target_link_libraries(DNSSD INTERFACE Avahi::compat-libdns_sd)
+        else()
+            # unfortunately some versions of the Conan 'avahi' recipe may produce a lowercase namespace
+            target_link_libraries(DNSSD INTERFACE avahi::compat-libdns_sd)
+        endif()
     else()
         find_package(DNSSD REQUIRED)
         if(NOT DNSSD_VERSION)


will-leathers-smith-sony:
I have tested this by deleting avahi from my local conan cache and deleting the package for nmos-cpp, then applying the fix to the relevant nmos-cpp cmake file inside Conan. Seems good to me.

PR #308, opened 2022-11-30
Use base_model instead of model in the OCSP behaviour_thread

lo-simon:
`base_model` should be used rather than `model` for the OCSP behaviour thread implementation.


garethsb (approved):
LGTM :-)

PR #271, opened 2022-07-11
IS-11 support

N-Nagorny:
Hello everyone!

This PR adds Stream Compatibility Management API (https://github.com/AMWA-TV/is-11/commit/e98d15681ea4005e62828c41dc33f299395b2828) implementation and IS-11 support.

Could you please review it and describe what's needed to change to be merged?

PR #309, opened 2022-11-30
Fix nmos::match_rational_constraint and add support for pattern constraints for string parameter constraints

garethsb:
Spotted these issues while reviewing PR #271, so thank you, @N-Nagorny!


lo-simon:
LGTM

PR #316, opened 2023-02-01
Channel Mapping API inputs/outputs

garethsb:
Tiny fix to make sure we don't add Inputs `input0`, `input1`, etc. to the Channel Mapping API resources if there aren't actually corresponding Receivers in the Node API resources (and same for Outputs vs. Senders). Hadn't spotted this when I added `receivers` and `senders` arrays to the _config.json_.

Other changes are just optimization to avoid inserting/trying to update resources where the Channel Mapping API and/or Events API are disabled via `channelmapping_port` and `events_port` being set to -1.

### DIFF
--- a/Development/nmos-cpp-node/node_implementation.cpp
+++ b/Development/nmos-cpp-node/node_implementation.cpp
@@ -710,9 +710,12 @@ void node_implementation_init(nmos::node_model& model, slog::base_gate& gate)
         }
     }
 
-    // example audio inputs
+    // example channelmapping resources demonstrating a range of input/output capabilities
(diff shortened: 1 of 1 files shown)


garethsb:
@lo-simon Ubuntu 14.04 build tools are broken again... log ends...
```
xz: (stdin): File format not recognized
tar: Child returned status 1
tar: Error is not recoverable: exiting now
Error: Process completed with exit code 2.
```

garethsb:
Now we noticed that `tar xz` is probably related to the "install cmake" step using lukka/get-cmake@v3.24.2, so this could just have been an ephemeral corrupt download... 🤞

PR #317, opened 2023-02-20
SDP HDCP signalling

N-Nagorny:
I explored what's needed to signalize HDCP support in RTP streams and found two useful SDP session attributes: `hkep` described in https://vsf.tv/download/technical_recommendations/VSF_TR-10-5_2022-03-22.pdf (Section 10) and `extmap` described in https://www.digital-cp.com/sites/default/files/HDCP%20Direct%20Adaptation%20Amendment%20Spec%20Rev2_3.pdf (Section 3.6.1).

Could you guys review this PR please and give some feedback if it's useful for the project?

### DIFF
--- a/Development/nmos/sdp_attributes.h
+++ b/Development/nmos/sdp_attributes.h
@@ -1,0 +1,48 @@
+#ifndef NMOS_SDP_ATTRIBUTES_H
+#define NMOS_SDP_ATTRIBUTES_H
+
+#include "sdp/json.h"
+
+namespace nmos
+{
+    namespace sdp_attributes
+    {
+        // RTP Header Extensions
+        // See https://tools.ietf.org/html/rfc5285#section-5
+        struct extmap
+        {
+            uint64_t local_id;
+            sdp::direction direction;
+            utility::string_t uri;
+            utility::string_t ext_attributes;
+
+            extmap() : local_id() {}
+            extmap(uint64_t local_id, const utility::string_t& uri) : local_id(local_id), uri(uri) {}
+            extmap(uint64_t local_id, const sdp::direction& direction, const utility::string_t& uri) : local_id(local_id), direction(direction), uri(uri) {}
+            extmap(uint64_t local_id, const utility::string_t& uri, const utility::string_t& ext_attributes) : local_id(local_id), uri(uri), ext_attributes(ext_attributes) {}
+            extmap(uint64_t local_id, const sdp::direction& direction, const utility::string_t& uri, const utility::string_t& ext_attributes) : local_id(local_id), direction(direction), uri(uri), ext_attributes(ext_attributes) {}
+        };
+
+        web::json::value make_extmap(const extmap& extmap);
+        extmap parse_extmap(const web::json::value& extmap);
+
+        // HDCP Key Exchange Protocol (HKEP) Signalling
+        // See VSF TR-10-5:2022 Internet Protocol Media Experience (IPMX): HDCP Key Exchange Protocol, Section 10
+        // at https://videoservicesforum.com/download/technical_recommendations/VSF_TR-10-5_2022-03-22.pdf
+        struct hkep
+        {
+            uint64_t port;
+            utility::string_t unicast_address;
+            utility::string_t node_id;
+            utility::string_t port_id;
+
+            hkep() : port() {}
+            hkep(uint64_t port, const utility::string_t& unicast_address, const utility::string_t& node_id, const utility::string_t& port_id) : port(port), unicast_address(unicast_address), node_id(node_id), port_id(port_id) {}
+        };
+
+        web::json::value make_hkep(const hkep& hkep);
+        hkep parse_hkep(const web::json::value& hkep);
+    }
+}
+
+#endif
--- a/Development/nmos/sdp_utils.cpp
+++ b/Development/nmos/sdp_utils.cpp
@@ -727,7 +727,7 @@ namespace nmos
         for (const auto& extmap_entry : sdp_params.extmap)
         {
             // a=extmap:<value>["/"<direction>] <URI> <extensionattributes>
-            // See https://www.rfc-editor.org/rfc/rfc5285#section-5
+            // See https://tools.ietf.org/html/rfc5285#section-5
             web::json::push_back(
                 session_attributes, value_of({
                     { sdp::fields::name, sdp::attributes::extmap },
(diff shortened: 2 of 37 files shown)


Development/sdp/sdp_grammar.cpp
```diff
@@ -714,6 +714,54 @@ namespace sdp
...
+                            std::string s;
+                            s += number_converter.format(v.at(sdp::fields::local_id));
+                            if (v.has_field(sdp::fields::direction)) s += "/" + string_converter.format(v.at(sdp::fields::direction));
+                            s += " " + string_converter.format(v.at(sdp::fields::uri));
+                            if (v.has_field(sdp::fields::extensionattributes)) s += " " + string_converter.format(v.at(sdp::fields::extensionattributes));
+                            return s;
+                        },
+                        [](const std::string& s) {
+                            auto v = web::json::value::object(keep_order);
+                            size_t pos = (!s.empty() && ' ' == s.front()) ? 1 : 0;
```
garethsb, line 731: The ABNF doesn't seem to allow SP before the local ID? Are bad implementations common in the wild?
https://www.rfc-editor.org/rfc/rfc5285#section-7

I wonder if the rest of this parsing function can be done more simply, I'll have a play and see!

Development/nmos/sdp_utils.h
```diff
@@ -245,11 +245,41 @@ namespace nmos
...
+            utility::string_t uri;
+            utility::string_t ext_attributes;
+
+            extmap_t() : local_id() {}
+            extmap_t(uint64_t local_id, const utility::string_t& uri): local_id(local_id), uri(uri) {}
+            extmap_t(uint64_t local_id, const sdp::direction& direction, const utility::string_t& uri): local_id(local_id), direction(direction), uri(uri) {}
+            extmap_t(uint64_t local_id, const utility::string_t& uri, const utility::string_t& ext_attributes): local_id(local_id), uri(uri), ext_attributes(ext_attributes) {}
+            extmap_t(uint64_t local_id, const sdp::direction& direction, const utility::string_t& uri, const utility::string_t& ext_attributes): local_id(local_id), direction(direction), uri(uri), ext_attributes(ext_attributes) {}
+        };
+        std::vector<sdp_parameters::extmap_t> extmap;
```
garethsb, line 263: I think `extmap` is a media attribute not a session attribute, right?
And multiple `extmap` attributes may be needed for a media description?
But also we have ST 2022-7, so do we need different set of `extmap` for each leg?

Development/nmos/sdp_utils.h
```diff
@@ -245,11 +245,41 @@ namespace nmos
             {}
         } mediaclk;
 
+        // RTP Header Extensions
+        // See https://www.rfc-editor.org/rfc/rfc5285#section-5
```
garethsb, line 249: ⬇️ **minor nit:** IETF certainly now redirects requests for RFCs to rfc-editor.org, but for consistency for now please could we keep all URLs the same as other references, i.e. https://tools.ietf.org/html/rfc5285#section-5?

Development/nmos/sdp_utils.cpp
```diff
@@ -1121,6 +1159,58 @@ namespace nmos
...
 
+        // RTP Header Extensions
+        // See https://www.rfc-editor.org/rfc/rfc5285#section-5
+        {
+            auto isExtmap = [](const web::json::value& nv)
+            {
+                return sdp::fields::name(nv) == sdp::attributes::extmap;
+            };
+            auto iter = session_attributes.begin();
+            while ((iter = std::find_if(iter, session_attributes.end(), isExtmap)) != session_attributes.end())
```
garethsb, line 1170: Would something like this work? (And similar for `sdp_params.hkep` obvs.)

```
sdp_params.extmap = boost::copy_range<std::vector<sdp_parameters::extmap_t>>(session_attributes | boost::adaptors::filtered([](const web::json::value& nv)
{
    return sdp::fields::name(nv) == sdp::attributes::extmap;
}) | boost::adaptors::transformed([](const web::json::value& nv)
{
    return sdp_parameters::extmap_t(...);
}));

File: Development/nmos/sdp_utils.cpp [hunk 5]
```diff
@@ -723,6 +723,44 @@ namespace nmos
...
+            // a=hkep:<port> <nettype> <addrtype> <unicast-address> <node-id> <port-id>
+            // See VSF TR-10-5 Section 10
+            web::json::push_back(
+                session_attributes, value_of({
+                    { sdp::fields::name, sdp::attributes::hkep },
+                    { sdp::fields::value, value_of({
+                        { sdp::fields::port, hkep_attr.port },
+                        { sdp::fields::network_type, sdp::network_types::internet.name },
+                        { sdp::fields::address_type, details::get_address_type_multicast(origin_address).first.name },
+                        { sdp::fields::unicast_address, origin_address },
```
garethsb, line 756: Is the logic of `nmos::details::get_origin_address()` always appropriate for `hkep`?

Development/sdp/sdp_grammar.cpp
N-Nagorny: Yes, SP is not allowed and I never seen such SDP samples, it's my bad.

Regarding making it simpler, I see that the regex may be turned into just `\D`. As for the optional `direction` syntax element, the only alternative that I see now is reading a string token and searching it in the list of directions but I'm not sure that it's simpler than the current implementation.

Development/sdp/sdp_grammar.cpp
```diff
@@ -714,6 +714,54 @@ namespace sdp
...
+                            return v;
+                        }
+                    }
+                },
+                {
+                    sdp::attributes::hkep,
+                    {
+                        [](const web::json::value& v) {
+                            std::string s;
+                            s += number_converter.format(v.at(sdp::fields::port));
```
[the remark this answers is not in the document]
N-Nagorny: It looks more like SP is not allowed here as well, that would allow us to just reuse `object_converter`.

https://www.rfc-editor.org/rfc/rfc5234#appendix-B.1 says that `HEXDIG` is uppercase. Should we transform the IDs from JSON representation to uppercase while generating SDP representation?

Development/nmos/sdp_utils.h
N-Nagorny: https://www.rfc-editor.org/rfc/rfc8285#section-5 says
> The mapping MAY be provided per media stream (in the media-level section(s) of SDP, i.e., after an "m=" line) or globally for all streams (i.e., before the first "m=" line, at session level). The definitions MUST be either all session level or all media level; it is not permitted to mix the two styles. In addition, as noted above, the IDs used MUST be unique in each media section of the SDP or unique in the session for session-level SDP declarations.

If I correctly understand ST 2022-7, it's about two same RTP streams going different paths (different UDP headers). In this case, RTP headers are the same and it still makes sense to put `extmap` at session level.

Development/nmos/sdp_utils.h
N-Nagorny: Sure, fixed it.

Development/nmos/sdp_utils.cpp
N-Nagorny: Perfect! Thank you, didn't notice it in the original code for some reason.

Development/nmos/sdp_utils.cpp
N-Nagorny: Maybe, it's better to allow apps explicitly set it, you're right.

N-Nagorny:
Regarding keeping `sdp_parameters` the current size, I don't have other ideas except of copying the approach from JSON representation: replacing member variables, related to session attributes, to a single `vector<variant<extmap, hkep>>` or something like this.

Development/sdp/sdp_grammar.cpp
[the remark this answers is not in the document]
garethsb: You're right, `HEXDIG` is uppercase, but TR-10-5 doesn't include ABNF that could state clearly `HEXDIG`, so it's not completely clear.

I assume **NodeId** an NMOS Node `id`? In which case, should that follow NMOS and be lowercase?

How about **PortId**, what spec defines this 5 octet thing?

Maybe these are questions for VSF...

Development/nmos/sdp_utils.h
garethsb: Ah, interesting! OK then...

garethsb (approved):
L**V**GTM (without having used `extmap` and `hkep` before myself!) - apart from the issue of session-level or media-level `extmap` described in https://github.com/sony/nmos-cpp/pull/317#discussion_r1185646191.

Development/nmos/sdp_utils.h
garethsb: Actually... having looked back at examples of `extmap` all the ones I find use the media-level attributes 😕

https://github.com/search?q=org%3AAMWA-TV+extmap&type=code

garethsb:
> Regarding keeping sdp_parameters the current size,

I think it's a nice idea to include the changes to parse and generate `extmap` and `hkep` between SDP string and JSON in _sdp/sdp_grammar.cpp_ etc. so that `sdp::make_session_description()` and `sdp::parse_sesssion_description()` support them by default.

My question is whether these attributes are sufficiently general use to mean that `nmos::sdp_parameters` and the `nmos::make_session_description()` and `nmos::get_session_description_sdp_parameters()` functions should include the higher level generate/parse from typed C++ structs/values to/from JSON, or whether we should leave that to user application code, which can access the JSON model directly, alongside those functions?

Development/nmos/sdp_attributes.h
```diff
@@ -0,0 +1,48 @@
...
+{
+    namespace sdp_attributes
+    {
+        // RTP Header Extensions
+        // See https://tools.ietf.org/html/rfc5285#section-5
+        struct extmap
+        {
+            uint64_t local_id;
+            sdp::direction direction;
+            utility::string_t uri;
```
garethsb, line 16: @N-Nagorny What do you think about making this a `web::uri`? Would that be good or bad? It would mean `make_extmap()` could throw `web::uri_exception`, and I'm not sure that's what we want. (We have lots of "typed string enums" in e.g. `nmos::sdp_parameters` but these don't throw if constructed with a string that isn't a recognized value.)

Development/nmos/sdp_utils.h
garethsb: Resolved by delegating this decision to application code in the latest refactor to add _nmos/sdp_attributes.h_.

garethsb:
@N-Nagorny over to you to see if this meets your requirements :-)

Development/nmos/sdp_attributes.h
```diff
@@ -0,0 +1,48 @@
...
+
+        // HDCP Key Exchange Protocol (HKEP) Signalling
+        // See VSF TR-10-5:2022 Internet Protocol Media Experience (IPMX): HDCP Key Exchange Protocol, Section 10
+        // at https://videoservicesforum.com/download/technical_recommendations/VSF_TR-10-5_2022-03-22.pdf
+        struct hkep
+        {
+            uint64_t port;
+            utility::string_t unicast_address;
+            utility::string_t node_id;
+            utility::string_t port_id;
```
garethsb, line 37: Should `node_id` be represented in this struct as `nmos::id` (actually just a typedef for `utility::string_t` right now, of course)?

Should `port_id` be represented in this struct as `uint64_t` since it is a 40-bit value, even though it's represented as `xx:xx:xx:xx:xx` string in the SDP file?

Development/nmos/sdp_attributes.h
N-Nagorny: I don't see how `web::uri` can help for constructing an entry, it's too permissive. It would be useful for parsing but I don't know if it makes sense if `scheme` usually returns `"urn"` and `path` returns the rest of the string and the other functions return nothing. Let's leave it as `utility::string_t`.

N-Nagorny:
@garethsb, thank you for the review and the rework you've done! I'm happy with the current code and it meets my requirements!

Development/nmos/sdp_attributes.h
garethsb: @N-Nagorny Based on [Alain's response](https://github.com/sony/nmos-cpp/pull/317#discussion_r1210096686), if we wanted to make `node_id` be `nmos::id` we would need to lowercase the received string. At the moment, that would be in your application code, but I'm not sure what's best, since everyone using this `struct hkep` would have to do the same.

PR #324, opened 2023-04-20
Enabling client and server sockets binding on specified network interfaces.

lo-simon:
Currently, there is no support for the client websocket binding, it is due to missing implementation in the cpprestsdk. I have been attempting to patch the cpprestsdk to add this feature, unfortunately, when I tested the client websocket binding, it was still using the available network interface instead the one specified.
 
Even if that work, we have no idea how long will cpprestsdk merge the patch.
 
So for now, we are leaving the client websocket binding out. Users can specify the binding interface via the `client_address` and the `server_address` in the settings JSON.
where
`client_address` - IP address of the network interface to bind client connections for Linux (no websocket support)
`server_address` - IP address of the network interface to bind server connections

### DIFF
--- a/Development/cpprest/host_utils.cpp
+++ b/Development/cpprest/host_utils.cpp
@@ -5,6 +5,7 @@
 #include <boost/asio/ip/host_name.hpp>
 #include <boost/asio/ip/tcp.hpp>
 #include <boost/range/adaptor/filtered.hpp>
+#include <boost/range/algorithm.hpp>
 #include "cpprest/asyncrt_utils.h" // for utility::conversions
 
 #if defined(_WIN32)
@@ -327,6 +328,16 @@ namespace web
                     addresses.push_back(utility::conversions::to_string_t(re.endpoint().address().to_string()));
                 }
                 return addresses; // empty if host_name cannot be resolved
+            }
+
+            // get the associated network interface name from an IP address
+            utility::string_t get_interface_name(const utility::string_t& address, const std::vector<web::hosts::experimental::host_interface>& host_interfaces)
+            {
+                const auto interface = boost::range::find_if(host_interfaces, [&](const web::hosts::experimental::host_interface& interface)
+                {
+                    return interface.addresses.end() != boost::range::find(interface.addresses, address);
+                });
+                return host_interfaces.end() != interface ? interface->name : utility::string_t {};
             }
         }
     }
--- a/Development/cpprest/host_utils.cpp
+++ b/Development/cpprest/host_utils.cpp
@@ -5,7 +5,8 @@
 #include <boost/asio/ip/host_name.hpp>
 #include <boost/asio/ip/tcp.hpp>
 #include <boost/range/adaptor/filtered.hpp>
-#include <boost/range/algorithm.hpp>
+#include <boost/range/algorithm/find.hpp>
+#include <boost/range/algorithm/find_if.hpp>
 #include "cpprest/asyncrt_utils.h" // for utility::conversions
 
 #if defined(_WIN32)
@@ -337,7 +338,7 @@ namespace web
                 {
                     return interface.addresses.end() != boost::range::find(interface.addresses, address);
                 });
-                return host_interfaces.end() != interface ? interface->name : utility::string_t {};
+                return host_interfaces.end() != interface ? interface->name : utility::string_t{};
             }
         }
     }
--- a/Development/cpprest/uri_schemes.h
+++ b/Development/cpprest/uri_schemes.h
@@ -27,8 +27,7 @@ namespace web
     // Check if the URI is secure
     inline bool get_secure(const web::uri& uri)
     {
-        const auto scheme = uri.scheme();
-        return (scheme == uri_schemes::wss || scheme == uri_schemes::https);
+        return uri_schemes::https == uri.scheme() || uri_schemes::wss == uri.scheme();
     }
 }
 
--- a/Development/cpprest/uri_schemes.h
+++ b/Development/cpprest/uri_schemes.h
@@ -1,6 +1,7 @@
 #ifndef CPPREST_URI_SCHEMES_H
 #define CPPREST_URI_SCHEMES_H
 
+#include "cpprest/base_uri.h"
 #include "cpprest/details/basic_types.h"
 
 namespace web
@@ -22,6 +23,13 @@ namespace web
 
     inline utility::string_t http_scheme(bool secure) { return secure ? uri_schemes::https : uri_schemes::http; }
     inline utility::string_t ws_scheme(bool secure) { return secure ? uri_schemes::wss : uri_schemes::ws; }
+
+    // Check if the URI is secure
+    inline bool get_secure(const web::uri& uri)
+    {
+        const auto scheme = uri.scheme();
+        return (scheme == uri_schemes::wss || scheme == uri_schemes::https);
+    }
 }
 
 #endif
--- a/Development/cpprest/uri_schemes.h
+++ b/Development/cpprest/uri_schemes.h
@@ -1,7 +1,6 @@
 #ifndef CPPREST_URI_SCHEMES_H
 #define CPPREST_URI_SCHEMES_H
 
-#include "cpprest/base_uri.h"
 #include "cpprest/details/basic_types.h"
 
 namespace web
@@ -24,10 +23,9 @@ namespace web
     inline utility::string_t http_scheme(bool secure) { return secure ? uri_schemes::https : uri_schemes::http; }
     inline utility::string_t ws_scheme(bool secure) { return secure ? uri_schemes::wss : uri_schemes::ws; }
 
-    // Check if the URI is secure
-    inline bool is_secure(const web::uri& uri)
+    inline bool is_secure_uri_scheme(const utility::string_t& scheme)
     {
-        return uri_schemes::https == uri.scheme() || uri_schemes::wss == uri.scheme();
+        return uri_schemes::https == scheme || uri_schemes::wss == scheme;
     }
 }
 
--- a/Development/cpprest/uri_schemes.h
+++ b/Development/cpprest/uri_schemes.h
@@ -25,7 +25,7 @@ namespace web
     inline utility::string_t ws_scheme(bool secure) { return secure ? uri_schemes::wss : uri_schemes::ws; }
 
     // Check if the URI is secure
-    inline bool get_secure(const web::uri& uri)
+    inline bool is_secure(const web::uri& uri)
     {
         return uri_schemes::https == uri.scheme() || uri_schemes::wss == uri.scheme();
     }
--- a/Development/nmos-cpp-registry/config.json
+++ b/Development/nmos-cpp-registry/config.json
@@ -109,6 +109,13 @@
     //"mdns_port": 3208,
     //"schemas_port": 3208,
 
+    // client_address [registry, node]: IP address of the network interface to bind client connections for Linux
+    // only supprting HTTP/HTTPS client connections, no ws/wss support yet
+    //"client_address": "",
+
+    // server_address [registry, node]: IP address of the network interface to bind server connections
+    //"server_address": "",
+
     // addresses [registry, node]: addresses on which to listen for each API, or empty string for the wildcard address
 
     //"settings_address": "127.0.0.1",
--- a/Development/nmos-cpp-registry/config.json
+++ b/Development/nmos-cpp-registry/config.json
@@ -109,12 +109,17 @@
     //"mdns_port": 3208,
     //"schemas_port": 3208,
 
-    // addresses [registry, node]: addresses on which to listen for each API, or empty string for the wildcard address
+    // addresses [registry, node]: IP addresses on which to listen for each API, or empty string for the wildcard address
+
+    // server_address [registry, node]: if specified, this becomes the default address on which to listen for each API instead of the wildcard address
+    //"server_address": "",
+
+    // addresses [registry, node]: IP addresses on which to listen for specific APIs
 
     //"settings_address": "127.0.0.1",
     //"logging_address": "",
 
-    // addresses [registry]: addresses on which to listen for each API, or empty string for the wildcard address
+    // addresses [registry]: IP addresses on which to listen for specific APIs
 
     //"admin_address": "",
     //"mdns_address": "",
@@ -123,9 +128,6 @@
     // client_address [registry, node]: IP address of the network interface to bind client connections
     // for now, only supporting HTTP/HTTPS client connections on Linux
     //"client_address": "",
-
-    // server_address [registry, node]: IP address of the network interface to bind server connections
-    //"server_address": "",
 
     // query_ws_paging_default/query_ws_paging_limit [registry]: default/maximum number of events per message when using the Query WebSocket API (a client may request a lower limit)
     //"query_ws_paging_default": 10,
--- a/Development/nmos-cpp-registry/config.json
+++ b/Development/nmos-cpp-registry/config.json
@@ -110,7 +110,7 @@
     //"schemas_port": 3208,
 
     // client_address [registry, node]: IP address of the network interface to bind client connections for Linux
-    // only supprting HTTP/HTTPS client connections, no ws/wss support yet
+    // only supporting HTTP/HTTPS client connections, no ws/wss support yet
     //"client_address": "",
 
     // server_address [registry, node]: IP address of the network interface to bind server connections
--- a/Development/nmos-cpp-registry/config.json
+++ b/Development/nmos-cpp-registry/config.json
@@ -109,8 +109,8 @@
     //"mdns_port": 3208,
     //"schemas_port": 3208,
 
-    // client_address [registry, node]: IP address of the network interface to bind client connections for Linux
-    // only supporting HTTP/HTTPS client connections, no ws/wss support yet
+    // client_address [registry, node]: IP address of the network interface to bind client connections
+    // for now, only supporting HTTP/HTTPS client connections on Linux
     //"client_address": "",
 
     // server_address [registry, node]: IP address of the network interface to bind server connections
--- a/Development/nmos-cpp-registry/config.json
+++ b/Development/nmos-cpp-registry/config.json
@@ -109,13 +109,6 @@
     //"mdns_port": 3208,
     //"schemas_port": 3208,
 
-    // client_address [registry, node]: IP address of the network interface to bind client connections
-    // for now, only supporting HTTP/HTTPS client connections on Linux
-    //"client_address": "",
-
-    // server_address [registry, node]: IP address of the network interface to bind server connections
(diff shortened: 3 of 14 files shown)


Development/nmos/client_utils.cpp
```diff
@@ -55,39 +60,165 @@ namespace nmos
                 }
             };
         }
+#endif
```
garethsb, line 63: Missing blank line after

Development/nmos/client_utils.cpp
```diff
@@ -55,39 +60,165 @@ namespace nmos
...
+                    return boost::range::find_if(interfaces, [&](const web::hosts::experimental::host_interface& interface)
+                    {
+                        return interface.addresses.end() != boost::range::find(interface.addresses, address);
+                    });
+                };
+                const auto interface = find_interface();
+                if (interfaces.end() != interface) { interface_name = interface->name; }
+            }
+            return interface_name;
+        }
```
garethsb, line 83: Since this function is basically copied from the nmos-cpp-node example code, probably we should make it fit for both uses and put it in _cpprest/host_utils.(h,cpp)_?

This might work?

```suggestion
        // get the associated network interface name from an IP address
        inline utility::string_t get_interface_name(const utility::string_t& address, const std::vector<web::hosts::experimental::host_interface>& host_interfaces = web::hosts::experimental::host_interfaces())
        {
            const auto interface = boost::range::find_if(interfaces, [&](const web::hosts::experimental::host_interface& interface)
            {
                return interface.addresses.end() != boost::range::find(interface.addresses, address);
            });
            return interfaces.end() != interface ? interface->name : utility::string_t{};
        }
```

Development/nmos/client_utils.cpp
```diff
@@ -55,39 +60,165 @@ namespace nmos
...
+
+#if !defined(_WIN32) || !defined(__cplusplus_winrt) || defined(CPPREST_FORCE_HTTP_CLIENT_ASIO)
+        // bind socket to a specific network interface, only supporting Linux
+#if defined(linux) || defined(__linux) || defined(__linux__)
```
garethsb, line 87: Unnecessary, I believe. We use just `__linux__` in e.g. _cpprest/host_utils.cpp_ and a quick Google found this advice:
http://www.faqs.org/docs/Linux-HOWTO/GCC-HOWTO.html#INDEX.25

Development/nmos/client_utils.cpp
```diff
@@ -55,39 +60,165 @@ namespace nmos
...
+                auto socket = (boost::asio::ip::tcp::socket*)native_handle;
+                if (!socket->is_open())
+                {
+                    // for now, limited to IPv4
+                    socket->open(boost::asio::ip::tcp::v4());
+                }
+                socket_fd = socket->lowest_layer().native_handle();
+            }
+            // SO_BINDTODEVICE not defined in windows & mac
+            return (setsockopt(socket_fd, SOL_SOCKET, SO_BINDTODEVICE, utility::us2s(interface_name).c_str(), interface_name.length()) == 0);
```
garethsb, line 112: Unnecessary parentheses.

I'm pretty sure it works, but passing the pointer from `c_str()` on a temporary as one arg, and the length of the string before it was converted feels a bit uncomfortable. 😄 

```suggestion
            const auto interface_name_ = utility::us2s(interface_name);
            return setsockopt(socket_fd, SOL_SOCKET, SO_BINDTODEVICE, interface_name_.data(), interface_name_.length()) == 0;
```

BTW, does `setsockopt` require that the interface name is null terminated as well as passing the (unterminated) length? I assumed not, so suggested using `.data()` instead of `.c_str()`.

Development/nmos/client_utils.cpp
```diff
@@ -55,39 +60,165 @@ namespace nmos
...
+        }
+#endif
+
+        inline std::function<void(web::http::client::native_handle)> make_client_nativehandle_options(const utility::string_t& client_address, bool secure, slog::base_gate& gate)
+        {
+            // get the associated network interface name from IP address
+            const auto interface_name = get_interface_name(client_address);
```
garethsb, line 119: Is it right to get interface name early, rather than in the callback itself? Probably...

Development/nmos/client_utils.cpp
```diff
@@ -55,39 +60,165 @@ namespace nmos
...
+            if (!client_address.empty() && interface_name.empty())
+            {
+                slog::log<slog::severities::error>(gate, SLOG_FLF) << "No network interface found for " << client_address << " to bind for the HTTP client connection";
+            }
+
+            return [interface_name, secure, &gate](web::http::client::native_handle native_handle)
+            {
+                if (!interface_name.empty())
```
garethsb, line 127: Why not just immediately return an empty or no-op function from `make_client_nativehandle_options` if `client_address` is empty (or handle this outside this function completely)?

Development/nmos/client_utils.cpp
```diff
@@ -55,39 +60,165 @@ namespace nmos
...
+#else
+                    slog::log<slog::severities::error>(gate, SLOG_FLF) << "Bind HTTP client connection to " << interface_name << " not supported";
+#endif
+                }
+            };
+        }
+
+#ifdef CPPRESTSDK_ENABLE_BIND_WEBSOCKET_CLIENT
+        // The current version of the C++ REST SDK 2.10.18 does not provide the callback to enable the custom websocket setting
+        inline std::function<void(web::websockets::client::native_handle)> make_ws_client_nativehandle_options(const utility::string_t& client_address, bool secure, slog::base_gate& gate)
```
garethsb, line 144: Will `web::websockets::client::native_handle` be same underlying type as `web::http::client::native_handle`? I suppose not necessarily...
Just seems a lot of duplication between these functions.

Development/nmos/client_utils.cpp
```diff
@@ -55,39 +60,165 @@ namespace nmos
...
+    // construct client config based on settings and specified secure flag, e.g. using the specified proxy, and the OCSP client
     // with the remaining options defaulted, e.g. request timeout
-    web::http::client::http_client_config make_http_client_config(const nmos::settings& settings, load_ca_certificates_handler load_ca_certificates, slog::base_gate& gate)
+    web::http::client::http_client_config make_http_client_config(const nmos::settings& settings, load_ca_certificates_handler load_ca_certificates, bool secure, slog::base_gate& gate)
     {
         web::http::client::http_client_config config;
         const auto proxy = proxy_uri(settings);
         if (!proxy.is_empty()) config.set_proxy(proxy);
-        config.set_validate_certificates(nmos::experimental::fields::validate_certificates(settings));
+        if (secure) config.set_validate_certificates(nmos::experimental::fields::validate_certificates(settings));
```
garethsb, line 181: Clearly previously we thought it didn't matter to do `set_validate_certificiates` and `set_ssl_context_callback` even for HTTP, they just wouldn't be used. What changed?

Development/nmos/settings.cpp
```diff
@@ -54,8 +55,14 @@ namespace nmos
...
 
+            // if the "server_address" settings is presented, use it for the node's "api endpoints" and "href", and device's "controls href"
+            const auto server_address = experimental::get_server_address(settings);
+            if (!server_address.empty())
+            {
+                web::json::insert(settings, std::make_pair(nmos::fields::host_address, server_address));
+            }
             // if the "host_address" setting was omitted, use the first of the "host_addresses"
-            if (settings.has_field(nmos::fields::host_addresses))
+            else if (settings.has_field(nmos::fields::host_addresses))
```
garethsb, line 65: 💭

garethsb (commented):
First review.

Development/nmos/settings.cpp
```diff
@@ -151,7 +158,12 @@ namespace nmos
         {
             hosts.push_back(get_host_name(settings));
         }
-        if (0 != (mode & nmos::experimental::href_mode_addresses))
+        const auto server_address = experimental::get_server_address(settings);
```
garethsb, line 161: Not sure this is necessary. If the app only wants to advertise one address, then can't that be `host_address`?

Development/nmos/settings.cpp
```diff
@@ -189,6 +201,28 @@ namespace nmos
...
             return bst::nullopt;
         }
+
+        // Get server IP address of the interface to bind server connections
+        utility::string_t get_server_address(const settings& settings)
+        {
+            const auto server_address = nmos::experimental::fields::server_address(settings);
+            if (!server_address.empty())
+            {
+                // verify the server address match one of the interface addresses
```
garethsb, line 211: Why is this verification step necessary? If the app gave us a bogus address, it must be better to report that rather than silently ignore it. But I'd probably just trust the app and use `nmos::experimental::fields::server_address(settings)` directly...

Development/nmos/settings.cpp
garethsb, line 65: Finishing this thought... I think it's probably better to separate the use of `client_address` and `server_address` for binding sockets, from the use of `host_address(es)` for putting in HTTP headers and the JSON data model?

Development/cpprest/uri_schemes.h
```diff
@@ -22,6 +23,13 @@ namespace web
 
     inline utility::string_t http_scheme(bool secure) { return secure ? uri_schemes::https : uri_schemes::http; }
     inline utility::string_t ws_scheme(bool secure) { return secure ? uri_schemes::wss : uri_schemes::ws; }
+
+    // Check if the URI is secure
+    inline bool get_secure(const web::uri& uri)
```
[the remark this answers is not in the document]
lo-simon: I used `get_secure` because cpprestsdk websocketpp\uri.hpp using that name. But I am happy to change it to `is_secure`. btw initially I did use is_secure too until I spotted the name used in cpprestsdk.

Development/nmos/client_utils.cpp
```diff
@@ -55,39 +60,165 @@ namespace nmos
...
+                        slog::log<slog::severities::error>(gate, SLOG_FLF) << "Unable to bind HTTP client connection to " << interface_name << ", bind_to_device reported error : " << strerror_r(errno, error, sizeof(error));
+                    }
+#else
+                    slog::log<slog::severities::error>(gate, SLOG_FLF) << "Bind HTTP client connection to " << interface_name << " not supported";
+#endif
+                }
+            };
+        }
+
+#ifdef CPPRESTSDK_ENABLE_BIND_WEBSOCKET_CLIENT
```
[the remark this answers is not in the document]
lo-simon: My mistake, should both be `CPPRESTSDK_ENABLE_BIND_WEBSOCKET_CLIENT`

Development/nmos-cpp-registry/config.json
```diff
@@ -109,6 +109,13 @@
     //"mdns_port": 3208,
     //"schemas_port": 3208,
 
+    // client_address [registry, node]: IP address of the network interface to bind client connections for Linux
+    // only supporting HTTP/HTTPS client connections, no ws/wss support yet
+    //"client_address": "",
+
+    // server_address [registry, node]: IP address of the network interface to bind server connections
+    //"server_address": "",
```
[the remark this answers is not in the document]
lo-simon: > Do you think the server_address should be the default for settings_address, logging_address, admin_address, mdns_address and schemas_address if they aren't specified? Like http_port is the default for the <api>_port settings...

yes, I do have the same thought too.

Development/nmos/client_utils.cpp
lo-simon: As we have the `secure` variable, I thought it will look better not to call the `set_validate_certificates` if not in secure mode.

garethsb: We discussed that an OCSP client is insecure even when the client_secure setting is true.

We noted that `load_ca_certificates` would only be required when `secure` is false, so that feels a bit odd too.

Development/nmos/ocsp_behaviour.cpp
```diff
@@ -141,9 +142,9 @@ namespace nmos
 
     namespace details
     {
-        web::http::client::http_client_config make_ocsp_client_config(const nmos::settings& settings, load_ca_certificates_handler load_ca_certificates, slog::base_gate& gate)
+        web::http::client::http_client_config make_ocsp_client_config(const nmos::settings& settings, load_ca_certificates_handler load_ca_certificates, bool secure, slog::base_gate& gate)
```
garethsb, line 145: Isn't OCSP always done over HTTP, not HTTPS? Hmm...

This discussion explains why using HTTP is the best practice... https://social.technet.microsoft.com/Forums/office/en-US/c65e1784-39be-4732-a135-bfba7446ad05/should-the-ocsp-responder-service-be-running-http-80-or-https-443-

But the spec says it MAY be done over HTTPS... https://datatracker.ietf.org/doc/html/rfc6960#appendix-A.1

Development/nmos/settings.cpp
garethsb: In the case of a reverse proxy (e.g. for TLS termination) the server_address is an implementation detail, where the host_address(es) are the public IP addresses of the proxy.

Development/nmos/client_utils.cpp
lo-simon: From my understanding, they are the same at least for the Linux implementation. see the diff of cpprestsdk. 

```
 Release/include/cpprest/ws_client.h              | 26 ++++++++++++++++++++++++
 Release/src/websockets/client/ws_client_wspp.cpp | 12 +++++++++++
 2 files changed, 38 insertions(+)

diff --git a/Release/include/cpprest/ws_client.h b/Release/include/cpprest/ws_client.h
index af17bd60..74d8c9c0 100644
--- a/Release/include/cpprest/ws_client.h
+++ b/Release/include/cpprest/ws_client.h
@@ -69,6 +69,8 @@ enum class websocket_close_status
     server_terminate = 1011,
 };
 
+typedef void* native_handle;
+
 /// <summary>
 /// Websocket client configuration class, used to set the possible configuration options
 /// used to create an websocket_client instance.
@@ -179,6 +181,29 @@ public:
     /// caution.</remarks>
     void set_validate_certificates(bool validate_certs) { m_validate_certificates = validate_certs; }
 
+    /// <summary>
+    /// Sets a callback to enable custom setting of platform specific options.
+    /// </summary>
+    /// <remarks>
+    /// The native_handle is the following type depending on the underlying platform:
+    ///         https - boost::asio::ssl::stream<boost::asio::ip::tcp::socket &> *
+    ///         http - boost::asio::ip::tcp::socket *
+    /// </remarks>
+    /// <param name="callback">A user callback allowing for customization of the connect</param>
+    void set_nativehandle_options(const std::function<void(native_handle)>& callback)
+    {
+        m_set_user_nativehandle_options = callback;
+    }
+
+    /// <summary>
+    /// Invokes a user's callback to allow for customization of the connect.
+    /// </summary>
+    /// <param name="handle">A internal implementation handle.</param>
+    void invoke_nativehandle_options(native_handle handle) const
+    {
+        if (m_set_user_nativehandle_options) m_set_user_nativehandle_options(handle);
+    }
+
 #if !defined(_WIN32) || !defined(__cplusplus_winrt)
     /// <summary>
     /// Sets a callback to enable custom setting of the ssl context, at construction time.
@@ -209,6 +234,7 @@ private:
 #if !defined(_WIN32) || !defined(__cplusplus_winrt)
     std::function<void(boost::asio::ssl::context&)> m_ssl_context_callback;
 #endif
+    std::function<void(native_handle)> m_set_user_nativehandle_options;
 };
 
 /// <summary>
diff --git a/Release/src/websockets/client/ws_client_wspp.cpp b/Release/src/websockets/client/ws_client_wspp.cpp
index d7c31c40..825ae465 100644
--- a/Release/src/websockets/client/ws_client_wspp.cpp
+++ b/Release/src/websockets/client/ws_client_wspp.cpp
@@ -262,6 +262,8 @@ public:
                         SSL_set_tlsext_host_name(ssl_stream.native_handle(), const_cast<char*>(server_name.c_str()));
                     }
                 }
+
+                m_config.invoke_nativehandle_options(&ssl_stream);
             });
 
             return connect_impl<websocketpp::config::asio_tls_client>();
@@ -269,6 +271,16 @@ public:
         else
         {
             m_client = std::unique_ptr<websocketpp_client_base>(new websocketpp_client());
+
+            // Options specific to underlying socket.
+            auto& client = m_client->client<websocketpp::config::asio_client>();
+
+            client.set_socket_init_handler([this](websocketpp::connection_hdl,
+                                                  boost::asio::ip::tcp::socket& socket) {
+
+                m_config.invoke_nativehandle_options(&socket);
+            });
+
             return connect_impl<websocketpp::config::asio_client>();
         }
```
     }

Development/cpprest/uri_schemes.h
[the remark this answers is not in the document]
lo-simon: get_secure removed to is_secure

Development/nmos/client_utils.cpp
lo-simon: hmm, do you think the network card could be changed, or set up in the callback?

Development/nmos/client_utils.cpp
lo-simon: Okay, add no-op function in event of an empty interface name

Development/nmos/client_utils.cpp
garethsb: IP address assigned to interfaces could change at least, but we have that challenge in lots of other places so let's not worry...

Development/nmos/client_utils.cpp
lo-simon: Yes, you are right, `setsockopt` does not require the interface name to be null-terminated, as it is controlled by the data length parameter.

Development/nmos/settings.cpp
lo-simon: Yes, we can use the href_mode or the host_address to advertise for one address.

Development/nmos/settings.cpp
lo-simon: Yes, server_address is an implementation detail, where the host_address(es) are the public IP addresses of the proxy.

Development/cpprest/host_utils.cpp
```diff
@@ -5,6 +5,7 @@
 #include <boost/asio/ip/host_name.hpp>
 #include <boost/asio/ip/tcp.hpp>
 #include <boost/range/adaptor/filtered.hpp>
+#include <boost/range/algorithm.hpp>
```
garethsb, line 8: If it wouldn't involve a lot of `#include`s, could you avoid the convenience header, and use the precise ones that are needed?

Maybe this?

```suggestion
#include <boost/range/algorithm/find.hpp>
#include <boost/range/algorithm/find_if.hpp>
```

Development/nmos/client_utils.cpp
garethsb: We discussed that `secure` is currently needed because of the type-erased `native_handle` types, so added a comment about that, and agreed to leave as is.

PR #330, opened 2023-05-09
Create IS-12 websocket server

lo-simon:
Add the basic WebSocket server for IS-12

### DIFF
--- a/Development/nmos/control_protocol_ws_api.cpp
+++ b/Development/nmos/control_protocol_ws_api.cpp
@@ -1,0 +1,84 @@
+#include "nmos/control_protocol_ws_api.h"
+
+#include "nmos/slog.h"
+
+namespace nmos
+{
+    // IS-12 Control Protocol WebSocket API
+
+    web::websockets::experimental::listener::validate_handler make_control_protocol_ws_validate_handler(nmos::node_model& model, slog::base_gate& gate_)
+    {
+        return [&model, &gate_](web::http::http_request req)
+        {
+            nmos::ws_api_gate gate(gate_, req.request_uri());
+
+            // RFC 6750 defines two methods of sending bearer access tokens which are applicable to WebSocket
+            // Clients SHOULD use the "Authorization Request Header Field" method.
+            // Clients MAY use a "URI Query Parameter".
+            // See https://tools.ietf.org/html/rfc6750#section-2
+
+            // For now just return true
+            const auto& ws_ncp_path = req.request_uri().path();
+            slog::log<slog::severities::more_info>(gate, SLOG_FLF) << "Validating websocket connection to: " << ws_ncp_path;
+
+            return true;
+        };
+    }
+
+    web::websockets::experimental::listener::open_handler make_control_protocol_ws_open_handler(nmos::node_model& model, nmos::websockets& websockets, slog::base_gate& gate_)
+    {
+        return [&model, &websockets, &gate_](const web::uri& connection_uri, const web::websockets::experimental::listener::connection_id& connection_id)
+        {
+            nmos::ws_api_gate gate(gate_, connection_uri);
+
+            const auto& ws_ncp_path = connection_uri.path();
+            slog::log<slog::severities::info>(gate, SLOG_FLF) << "Opening websocket connection to: " << ws_ncp_path;
+            {
+                // create a websocket connection resource
+
+                nmos::id id = nmos::make_id();
+                websockets.insert({ id, connection_id });
+
+                slog::log<slog::severities::info>(gate, SLOG_FLF) << "Creating websocket connection: " << id;
+            }
+        };
+    }
+
+    web::websockets::experimental::listener::close_handler make_control_protocol_ws_close_handler(nmos::node_model& model, nmos::websockets& websockets, slog::base_gate& gate_)
+    {
+        return [&model, &websockets, &gate_](const web::uri& connection_uri, const web::websockets::experimental::listener::connection_id& connection_id, web::websockets::websocket_close_status close_status, const utility::string_t& close_reason)
+        {
+            nmos::ws_api_gate gate(gate_, connection_uri);
+
+            const auto& ws_ncp_path = connection_uri.path();
+            slog::log<slog::severities::info>(gate, SLOG_FLF) << "Closing websocket connection to: " << ws_ncp_path << " [" << (int)close_status << ": " << close_reason << "]";
+
+            auto websocket = websockets.right.find(connection_id);
+            if (websockets.right.end() != websocket)
+            {
+                slog::log<slog::severities::info>(gate, SLOG_FLF) << "Deleting websocket connection";
+
+                websockets.right.erase(websocket);
+            }
+        };
+    }
+
+    web::websockets::experimental::listener::message_handler make_control_protocol_ws_message_handler(nmos::node_model& model, nmos::websockets& websockets, slog::base_gate& gate_)
+    {
+        return [&model, &websockets, &gate_](const web::uri& connection_uri, const web::websockets::experimental::listener::connection_id& connection_id, const web::websockets::websocket_incoming_message& msg_)
+        {
+            nmos::ws_api_gate gate(gate_, connection_uri);
+           // theoretically blocking, but in fact not
+            auto msg = msg_.extract_string().get();
+
+            const auto& ws_ncp_path = connection_uri.path();
+            slog::log<slog::severities::too_much_info>(gate, SLOG_FLF) << "Received websocket message: " << msg << " on connection to: " << ws_ncp_path;
+
+            auto websocket = websockets.right.find(connection_id);
+            if (websockets.right.end() != websocket)
+            {
+                // hmm, todo message handling....
+            }
+        };
+    }
+}
--- a/Development/nmos/control_protocol_ws_api.h
+++ b/Development/nmos/control_protocol_ws_api.h
@@ -1,0 +1,33 @@
+#ifndef NMOS_CONTROL_PROTOCOL_WS_API_H
+#define NMOS_CONTROL_PROTOCOL_WS_API_H
+
+#include "nmos/websockets.h"
+
+namespace slog
+{
+    class base_gate;
+}
+
+// Events API websocket implementation
+// See https://specs.amwa.tv/is-07/releases/v1.0.1/docs/5.2._Transport_-_Websocket.html
+namespace nmos
+{
+    struct node_model;
+
+    web::websockets::experimental::listener::validate_handler make_control_protocol_ws_validate_handler(nmos::node_model& model, slog::base_gate& gate);
+    web::websockets::experimental::listener::open_handler make_control_protocol_ws_open_handler(nmos::node_model& model, nmos::websockets& websockets, slog::base_gate& gate);
+    web::websockets::experimental::listener::close_handler make_control_protocol_ws_close_handler(nmos::node_model& model, nmos::websockets& websockets, slog::base_gate& gate);
+    web::websockets::experimental::listener::message_handler make_control_protocol_ws_message_handler(nmos::node_model& model, nmos::websockets& websockets, slog::base_gate& gate);
+
+    inline web::websockets::experimental::listener::websocket_listener_handlers make_control_protocol_ws_api(nmos::node_model& model, nmos::websockets& websockets, slog::base_gate& gate)
+    {
+        return{
+            nmos::make_control_protocol_ws_validate_handler(model, gate),
+            nmos::make_control_protocol_ws_open_handler(model, websockets, gate),
+            nmos::make_control_protocol_ws_close_handler(model, websockets, gate),
+            nmos::make_control_protocol_ws_message_handler(model, websockets, gate)
+        };
+    }
+}
+
+#endif
--- a/Development/nmos/node_resources.cpp
+++ b/Development/nmos/node_resources.cpp
@@ -16,6 +16,7 @@
 #include "nmos/is05_versions.h"
 #include "nmos/is07_versions.h"
 #include "nmos/is08_versions.h"
+#include "nmos/is12_versions.h"
 #include "nmos/media_type.h"
 #include "nmos/resource.h"
 #include "nmos/sdp_utils.h" // for nmos::make_components
@@ -122,6 +123,26 @@ namespace nmos
                     { U("href"), manifest_uri.set_host(host).to_uri().to_string() },
                     { U("type"), type }
                 }));
+            }
+        }
+
+        if (0 <= nmos::fields::control_protocol_ws_port(settings))
+        {
+            for (const auto& version : nmos::is12_versions::from_settings(settings))
+            {
+                auto ncp_uri = web::uri_builder()
+                    .set_scheme(nmos::ws_scheme(settings))
+                    .set_port(nmos::fields::control_protocol_ws_port(settings))
+                    .set_path(U("/x-nmos/ncp/") + make_api_version(version));
+                auto type = U("urn:x-nmos:control:ncp/") + make_api_version(version);
+
+                for (const auto& host : hosts)
+                {
+                    web::json::push_back(data[U("controls")], value_of({
+                        { U("href"), ncp_uri.set_host(host).to_uri().to_string() },
+                        { U("type"), type }
+                    }));
+                }
             }
         }
 
--- a/Development/nmos/node_server.cpp
+++ b/Development/nmos/node_server.cpp
@@ -3,6 +3,7 @@
 #include "cpprest/ws_utils.h"
 #include "nmos/api_utils.h"
 #include "nmos/channelmapping_activation.h"
+#include "nmos/control_protocol_ws_api.h"
 #include "nmos/events_api.h"
 #include "nmos/events_ws_api.h"
 #include "nmos/logging_api.h"
@@ -63,8 +64,15 @@ namespace nmos
             // Configure the Channel Mapping API
             node_server.api_routers[{ {}, nmos::fields::channelmapping_port(node_model.settings) }].mount({}, nmos::make_channelmapping_api(node_model, node_implementation.validate_map, gate));
 
-            auto& events_ws_api = node_server.ws_handlers[{ {}, nmos::fields::events_ws_port(node_model.settings) }];
+            const auto& events_ws_port = nmos::fields::events_ws_port(node_model.settings);
+            auto& events_ws_api = node_server.ws_handlers[{ {}, events_ws_port }];
             events_ws_api.first = nmos::make_events_ws_api(node_model, events_ws_api.second, gate);
+
+            // can't share a port between the events ws and the control protocol ws
+            const auto& control_protocol_ws_port = nmos::fields::control_protocol_ws_port(node_model.settings);
+            if (control_protocol_ws_port == events_ws_port) throw std::runtime_error("Same port used for events and control protocol websockets are not supported");
+            auto& control_protocol_ws_api = node_server.ws_handlers[{ {}, control_protocol_ws_port }];
+            control_protocol_ws_api.first = nmos::make_control_protocol_ws_api(node_model, control_protocol_ws_api.second, gate);
 
             // Set up the listeners for each HTTP API port
 
@@ -84,6 +92,7 @@ namespace nmos
             auto websocket_config = nmos::make_websocket_listener_config(node_model.settings, node_implementation.load_server_certificates, node_implementation.load_dh_param, node_implementation.get_ocsp_response, gate);
             websocket_config.set_log_callback(nmos::make_slog_logging_callback(gate));
 
+            size_t event_ws_pos{ 0 };
             for (auto& ws_handler : node_server.ws_handlers)
             {
                 // if IP address isn't specified for this router, use default server address or wildcard address
@@ -91,9 +100,11 @@ namespace nmos
                 // map the configured client port to the server port on which to listen
                 // hmm, this should probably also take account of the address
                 node_server.ws_listeners.push_back(nmos::make_ws_api_listener(server_secure, host, nmos::experimental::server_port(ws_handler.first.second, node_model.settings), ws_handler.second.first, websocket_config, gate));
+
+                event_ws_pos = (ws_handler.first.second == events_ws_port) ? event_ws_pos : ++event_ws_pos;
             }
 
-            auto& events_ws_listener = node_server.ws_listeners.back();
+            auto& events_ws_listener = node_server.ws_listeners.at(event_ws_pos);
 
             // Set up node operation (including the DNS-SD advertisements)
 
--- a/Development/cmake/NmosCppLibraries.cmake
+++ b/Development/cmake/NmosCppLibraries.cmake
@@ -757,6 +757,7 @@ set(NMOS_CPP_NMOS_SOURCES
     nmos/connection_api.cpp
     nmos/connection_events_activation.cpp
     nmos/connection_resources.cpp
+    nmos/control_protocol_ws_api.cpp
     nmos/did_sdid.cpp
     nmos/events_api.cpp
     nmos/events_resources.cpp
@@ -828,6 +829,7 @@ set(NMOS_CPP_NMOS_HEADERS
     nmos/connection_api.h
     nmos/connection_events_activation.h
     nmos/connection_resources.h
+    nmos/control_protocol_ws_api.h
     nmos/device_type.h
     nmos/did_sdid.h
     nmos/event_type.h
@@ -846,6 +848,7 @@ set(NMOS_CPP_NMOS_HEADERS
     nmos/is07_versions.h
     nmos/is08_versions.h
     nmos/is09_versions.h
+    nmos/is12_versions.h
     nmos/json_fields.h
     nmos/json_schema.h
     nmos/lldp_handler.h
--- a/Development/nmos/is12_versions.h
+++ b/Development/nmos/is12_versions.h
@@ -1,0 +1,26 @@
+#ifndef NMOS_IS12_VERSIONS_H
+#define NMOS_IS12_VERSIONS_H
+
+#include <set>
+#include <boost/range/adaptor/transformed.hpp>
+#include "nmos/api_version.h"
+#include "nmos/settings.h"
+
+namespace nmos
+{
+    namespace is12_versions
+    {
+        const api_version v1_0{ 1, 0 };
+
+        const std::set<api_version> all{ nmos::is12_versions::v1_0 };
+
+        inline std::set<api_version> from_settings(const nmos::settings& settings)
+        {
+            return settings.has_field(nmos::fields::is12_versions)
+                ? boost::copy_range<std::set<api_version>>(nmos::fields::is12_versions(settings) | boost::adaptors::transformed([](const web::json::value& v) { return nmos::parse_api_version(v.as_string()); }))
+                : nmos::is12_versions::all;
+        }
+    }
+}
+
+#endif
--- a/Development/nmos/settings.cpp
+++ b/Development/nmos/settings.cpp
@@ -66,6 +66,8 @@ namespace nmos
                 const auto http_port = nmos::fields::http_port(settings);
                 // can't share a port between an http_listener and a websocket_listener, so use next higher port
                 const auto ws_port = http_port + 1;
+                // can't share a port between the events ws and the control protocol ws
+                const auto ncp_ws_port = ws_port + 1;
                 if (registry) web::json::insert(settings, std::make_pair(nmos::fields::query_port, http_port));
                 if (registry) web::json::insert(settings, std::make_pair(nmos::fields::query_ws_port, ws_port));
                 if (registry) web::json::insert(settings, std::make_pair(nmos::fields::registration_port, http_port));
@@ -81,6 +83,7 @@ namespace nmos
                 if (registry) web::json::insert(settings, std::make_pair(nmos::experimental::fields::admin_port, http_port));
                 if (registry) web::json::insert(settings, std::make_pair(nmos::experimental::fields::mdns_port, http_port));
                 if (registry) web::json::insert(settings, std::make_pair(nmos::experimental::fields::schemas_port, http_port));
+                if (!registry) web::json::insert(settings, std::make_pair(nmos::fields::control_protocol_ws_port, ncp_ws_port));
             }
         }
     }
--- a/Development/nmos/settings.h
+++ b/Development/nmos/settings.h
@@ -101,6 +101,9 @@ namespace nmos
         // is09_versions [registry, node]: used to specify the enabled API versions for a version-locked configuration
         const web::json::field_as_array is09_versions{ U("is09_versions") }; // when omitted, nmos::is09_versions::all is used
 
+        // is12_versions [node]: used to specify the enabled API versions for a version-locked configuration
+        const web::json::field_as_array is12_versions{ U("is12_versions") }; // when omitted, nmos::is12_versions::all is used
+
         // pri [registry, node]: used for the 'pri' TXT record; specifying nmos::service_priorities::no_priority (maximum value) disables advertisement completely
         const web::json::field_as_integer_or pri{ U("pri"), 100 }; // default to highest_development_priority
 
@@ -136,6 +139,7 @@ namespace nmos
         const web::json::field_as_integer_or channelmapping_port{ U("channelmapping_port"), 3215 };
         // system_port [node]: used to construct request URLs for the System API (if not discovered via DNS-SD)
         const web::json::field_as_integer_or system_port{ U("system_port"), 10641 };
+        const web::json::field_as_integer_or control_protocol_ws_port{ U("control_protocol_ws_port"), 3218 };
 
         // listen_backlog [registry, node]: the maximum length of the queue of pending connections, or zero for the implementation default (the implementation may not honour this value)
         const web::json::field_as_integer_or listen_backlog{ U("listen_backlog"), 0 };


garethsb (commented):
Looking good

garethsb (commented):
Looking good

Development/nmos/node_server.cpp
```diff
@@ -63,9 +64,16 @@ namespace nmos
             // Configure the Channel Mapping API
             node_server.api_routers[{ {}, nmos::fields::channelmapping_port(node_model.settings) }].mount({}, nmos::make_channelmapping_api(node_model, node_implementation.validate_map, gate));
 
-            auto& events_ws_api = node_server.ws_handlers[{ {}, nmos::fields::events_ws_port(node_model.settings) }];
+            const auto& events_ws_port = nmos::fields::events_ws_port(node_model.settings);
+            auto& events_ws_api = node_server.ws_handlers[{ {}, events_ws_port }];
             events_ws_api.first = nmos::make_events_ws_api(node_model, events_ws_api.second, gate);
 
+            // can't share a port between the events ws and the control protocol ws
```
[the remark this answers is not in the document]
lo-simon: Yes, definitely. For now, just want to get the ball rolling.

Development/nmos/control_protocol_ws_api.h
```diff
@@ -0,0 +1,33 @@
...
+#define NMOS_CONTROL_PROTOCOL_WS_API_H
+
+#include "nmos/websockets.h"
+
+namespace slog
+{
+    class base_gate;
+}
+
+// Events API websocket implementation
```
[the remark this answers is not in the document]
lo-simon: Thanks, removed

Development/nmos/control_protocol_ws_api.cpp
```diff
@@ -0,0 +1,84 @@
...
+            nmos::ws_api_gate gate(gate_, connection_uri);
+
+            const auto& ws_ncp_path = connection_uri.path();
+            slog::log<slog::severities::info>(gate, SLOG_FLF) << "Opening websocket connection to: " << ws_ncp_path;
+            {
+                // create a websocket connection resource
+
+                nmos::id id = nmos::make_id();
+                websockets.insert({ id, connection_id });
+
```
maweit, line 41: websockets.insert should have be protected by a write lock.

Development/nmos/control_protocol_ws_api.cpp
```diff
@@ -0,0 +1,84 @@
...
+            const auto& ws_ncp_path = connection_uri.path();
+            slog::log<slog::severities::info>(gate, SLOG_FLF) << "Closing websocket connection to: " << ws_ncp_path << " [" << (int)close_status << ": " << close_reason << "]";
+
+            auto websocket = websockets.right.find(connection_id);
+            if (websockets.right.end() != websocket)
+            {
+                slog::log<slog::severities::info>(gate, SLOG_FLF) << "Deleting websocket connection";
+
+                websockets.right.erase(websocket);
```
maweit, line 61: Same as in the open method: there should be a write lock for the `erase` method

Development/nmos/control_protocol_ws_api.cpp
```diff
@@ -0,0 +1,84 @@
...
+        return [&model, &websockets, &gate_](const web::uri& connection_uri, const web::websockets::experimental::listener::connection_id& connection_id, const web::websockets::websocket_incoming_message& msg_)
+        {
+            nmos::ws_api_gate gate(gate_, connection_uri);
+           // theoretically blocking, but in fact not
+            auto msg = msg_.extract_string().get();
+
+            const auto& ws_ncp_path = connection_uri.path();
+            slog::log<slog::severities::too_much_info>(gate, SLOG_FLF) << "Received websocket message: " << msg << " on connection to: " << ws_ncp_path;
+
+            auto websocket = websockets.right.find(connection_id);
```
maweit, line 77: Likely needs a read lock.

Development/nmos/node_server.cpp
```diff
@@ -84,16 +92,19 @@ namespace nmos
...
+            size_t event_ws_pos{ 0 };
             for (auto& ws_handler : node_server.ws_handlers)
             {
                 // if IP address isn't specified for this router, use default server address or wildcard address
                 const auto& host = !ws_handler.first.first.empty() ? ws_handler.first.first : !server_address.empty() ? server_address : web::websockets::experimental::listener::host_wildcard;
                 // map the configured client port to the server port on which to listen
                 // hmm, this should probably also take account of the address
                 node_server.ws_listeners.push_back(nmos::make_ws_api_listener(server_secure, host, nmos::experimental::server_port(ws_handler.first.second, node_model.settings), ws_handler.second.first, websocket_config, gate));
+
+                event_ws_pos = (ws_handler.first.second == events_ws_port) ? event_ws_pos : ++event_ws_pos;
```
maweit, line 104: Don't thinks this works: if the first `ws_handler` is of event-type, 0 is assigned to `event_ws_pos`. If the second is not the `event_ws_pos`  is  incremented and points to the handler which is not an event-type. But in line 107 this is assigned to `events_ws_listener`. Maybe i understood something wrong?

Development/nmos/control_protocol_ws_api.cpp
garethsb: Good spot, I agree.

Compare e.g. https://github.com/sony/nmos-cpp/blob/694539a37798a711b38c26386168680b0af99dcb/Development/nmos/events_ws_api.cpp#L58

Development/nmos/control_protocol_ws_api.cpp
garethsb: Agree

Development/nmos/control_protocol_ws_api.cpp
garethsb: Agree

Development/nmos-cpp-node/node_implementation.cpp
```diff
@@ -894,6 +895,17 @@ void node_implementation_init(nmos::node_model& model, slog::base_gate& gate)
         auto channelmapping_output = nmos::make_channelmapping_output(id, name, description, source_id, channel_labels, routable_inputs);
         if (!insert_resource_after(delay_millis, model.channelmapping_resources, std::move(channelmapping_output), gate)) throw node_implementation_init_exception();
     }
+
+    // example root block
+    auto root_block = nmos::make_root_block();
+    // example device manager
+    auto device_manager = nmos::make_device_manager(2, root_block, model.settings);
```
maweit, line 902: Is a constant better here than a "2"?

Development/nmos-cpp-node/node_implementation.cpp
```diff
@@ -894,6 +895,17 @@ void node_implementation_init(nmos::node_model& model, slog::base_gate& gate)
...
+    if (!insert_resource_after(delay_millis, model.control_protocol_resources, std::move(device_manager), gate)) throw node_implementation_init_exception();
+    // example class manager
+    auto class_manager = nmos::make_class_manager(3, root_block);
```
maweit, line 905: Same as for line 902?

Development/nmos-cpp-node/node_implementation.cpp
lo-simon: This is the `oid` for the device manager instance, which can be between `2-65535`, as the root block can only be `1`

Development/nmos-cpp-node/node_implementation.cpp
lo-simon: This is the `oid` for the device manager instance, which can be between `2-65535`, as the root block can only be `1`. and `2` has been used for the device manager instance.

Development/nmos-cpp-node/node_implementation.cpp
jonathan-r-thorpe: In the short term, perhaps assign the value to a temporary variable called `oid`, or `device_manager_oid` to mitigate the "magicness" of the raw number?

In the long term we need to work out a strategy for assigning oids to objects. For instance a function that returns a monotonically increasing integer value for automatically generating the oids.

Development/nmos/control_protocol_handlers.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+    get_control_protocol_methods_handler make_get_control_protocol_methods_handler(experimental::control_protocol_state& control_protocol_state, slog::base_gate& gate)
+    {
+        return [&]()
+        {
+            slog::log<slog::severities::too_much_info>(gate, SLOG_FLF) << "Retrieve all method handlers from cache";
+
+            std::map<nmos::nc_class_id, experimental::methods> methods;
+
+            auto lock = control_protocol_state.read_lock();
```
maweit, line 71: Shouldn't this be a write-lock? `methods` gets overwritten

Development/nmos/control_protocol_methods.cpp
```diff
@@ -0,0 +1,550 @@
...
+            value nc_block_member_descriptor;
+
+            for (const auto& role : path.as_array())
+            {
+                // look for the role in members
+                if (resource->data.has_field(nmos::fields::nc::members))
+                {
+                    auto& members = nmos::fields::nc::members(resource->data);
+                    auto member_found = std::find_if(members.begin(), members.end(), [&](const web::json::value& nc_block_member_descriptor)
+                        {
```
maweit, line 357: Indentation different then elsewhere

Development/nmos/control_protocol_methods.cpp
```diff
@@ -0,0 +1,550 @@
...
+
+                    while (!inherited_class_id.empty())
+                    {
+                        const auto& inherited_control_class = get_control_protocol_class(inherited_class_id);
+                        {
+                            for (const auto& property : inherited_control_class.properties.as_array()) { web::json::push_back(properties, property); }
+                            for (const auto& method : inherited_control_class.methods.as_array()) { web::json::push_back(methods, method); }
+                            for (const auto& event : inherited_control_class.events.as_array()) { web::json::push_back(events, event); }
+                        }
+                        inherited_class_id.pop_back();
```
maweit, line 480: This is not clear to me: if `pop_back` is used this assumes that `inherited_class_id` has a specific order. But `get_control_protocol_class()` as implemented in control_protocol_handlers.cpp:11-24  uses a `find`-method which could deliver any position.

Development/nmos/control_protocol_methods.cpp
```diff
@@ -0,0 +1,550 @@
...
+                }
+
+                auto& data = resource->data.at(nmos::fields::nc::name(property));
+
+                if (!data.is_null() && data.as_array().size() > (size_t)index)
+                {
+                    return make_control_protocol_response(handle, { nc_method_status::ok }, data.at(index));
+                }
+
+                // out of bound
```
maweit, line 106: Maybe "out of bounds" can be misleading: if `data.is_null()` returns true this also an error but not out of bounds?
See also lines 152 and 245.

Development/nmos/node_resources.cpp
```diff
@@ -125,6 +126,26 @@ namespace nmos
             }
         }
 
+        if (0 <= nmos::fields::control_protocol_ws_port(settings))
```
maweit, line 129: Is port=0 really allowed?

Development/nmos/node_resources.cpp
```diff
@@ -125,6 +126,26 @@ namespace nmos
...
+        {
+            for (const auto& version : nmos::is12_versions::from_settings(settings))
+            {
+                auto ncp_uri = web::uri_builder()
+                    .set_scheme(nmos::ws_scheme(settings))
+                    .set_port(nmos::fields::control_protocol_ws_port(settings))
+                    .set_path(U("/x-nmos/ncp/") + make_api_version(version));
+                auto type = U("urn:x-nmos:control:ncp/") + make_api_version(version);
```
maweit, line 137: make_api_version called twice. Maybe saving it in one call is better?

Development/nmos/control_protocol_methods.cpp
lo-simon: Thanks for spotting it.

Development/nmos/control_protocol_methods.cpp
lo-simon: class_id is The idea here is that we walk the class_id

lo-simon: The `inherited_class_id` reflects the ancestry of the class being identified.

> A class id field is either a definition index or an authority key. A definition index is an ordinal that starts at 1 for every inheritance level of the control model class tree for example [ 1, 1, 3, 5].
https://specs.amwa.tv/ms-05-02/branches/v1.0-dev/docs/Framework.html#ncclassid

The idea is that if the `include_inherited` is required, we need to walk the `inherited_class_id` backwards to obtain all the inherited classes. At each pop_back we identify the next inherited `class_id`s, by using it to obtain the inherited `control_class` from the `control_protocol_state`'s control_classes **map** (via `get_control_protocol_class`) to build up the `properties, methods, and events`.

Development/nmos/control_protocol_methods.cpp
lo-simon: Thanks, well spotted.

Development/nmos/node_resources.cpp
lo-simon: Yes, but this is using the same style as other uri builds within the make_device function. This call is only being executed once in the node initialisation.

Development/nmos/control_protocol_handlers.cpp
lo-simon: I think the read lock is okay to build up the methods map.

Development/nmos-cpp-node/node_implementation.cpp
```diff
@@ -894,6 +900,289 @@ void node_implementation_init(nmos::node_model& model, slog::base_gate& gate)
...
+        if (!insert_resource_after(delay_millis, model.control_protocol_resources, std::move(right_gain), gate)) throw node_implementation_init_exception();
+        if (!insert_resource_after(delay_millis, model.control_protocol_resources, std::move(master_gain), gate)) throw node_implementation_init_exception();
+        if (!insert_resource_after(delay_millis, model.control_protocol_resources, std::move(channel_gain), gate)) throw node_implementation_init_exception();
+        if (!insert_resource_after(delay_millis, model.control_protocol_resources, std::move(stereo_gain), gate)) throw node_implementation_init_exception();
+        if (!insert_resource_after(delay_millis, model.control_protocol_resources, std::move(device_manager), gate)) throw node_implementation_init_exception();
+        if (!insert_resource_after(delay_millis, model.control_protocol_resources, std::move(class_manager), gate)) throw node_implementation_init_exception();
+        if (!insert_resource_after(delay_millis, model.control_protocol_resources, std::move(root_block), gate)) throw node_implementation_init_exception();
+    }
```
maweit, line 1185: I find it confusing that i construct a tree with push_backs'  and then also have to insert them one by on into the model. My expectation had been that you can just insert the complete root block tree into the model. Sounds reasonable?

Development/nmos/control_protocol_methods.cpp
```diff
@@ -0,0 +1,568 @@
...
+                {
+                    return make_control_protocol_message_response(handle, { nc_method_status::parameter_error });
+                }
+
+                const auto notification = make_control_protocol_notification(nmos::fields::nc::oid(resource->data), nc_object_property_changed_event_id, { parse_nc_property_id(property_id), nc_property_change_type::type::value_changed, val });
+                const auto notification_event = make_control_protocol_notification(web::json::value_of({ notification }));
+
+                modify_resource(resources, resource->id, [&](nmos::resource& resource)
+                {
+                    resource.data[nmos::fields::nc::name(property)] = val;
```
maweit, line 68: Can this segfault? `val` is use in the constructor of  `nc_property_changed_event_data` (line 63) which uses a std::move for the value. So `val` might actually be undefined later.

Development/nmos/control_protocol_methods.cpp
```diff
@@ -0,0 +1,568 @@
...
+                }
+
+                if (data.as_array().size() > (size_t)index)
+                {
+                    const auto notification = make_control_protocol_notification(nmos::fields::nc::oid(resource->data), nc_object_property_changed_event_id, { parse_nc_property_id(property_id), nc_property_change_type::type::sequence_item_changed, val, nc_id(index) });
+                    const auto notification_event = make_control_protocol_notification(web::json::value_of({ notification }));
+
+                    modify_resource(resources, resource->id, [&](nmos::resource& resource)
+                    {
+                        resource.data[nmos::fields::nc::name(property)][index] = val;
```
maweit, line 154: Same question as for line 68: if `val` gets std:move'd can this segfault?

Development/nmos/control_protocol_methods.cpp
lo-simon: Thank you @maweit for spotting it, I will correct them all.

Development/nmos/control_protocol_methods.cpp
lo-simon: Thank you @maweit for spotting it, I will correct them all.

Development/nmos-cpp-node/node_implementation.cpp
lo-simon: Good idea, will see what I can do.

Development/nmos-cpp-node/node_implementation.cpp
```diff
@@ -894,6 +920,374 @@ void node_implementation_init(nmos::node_model& model, slog::base_gate& gate)
...
+                // create "Example sequence numeric property" with level 1: property constraints, See https://specs.amwa.tv/ms-05-02/branches/v1.0.x/docs/Constraints.html
+                // use nmos::details::make_nc_parameter_constraints_number to create sequence property constraints
+                nmos::experimental::make_control_class_property(U("Example number sequence property"), { 3, 13 }, number_sequence, U("NcUint64"), false, false, true, false, make_number_example_argument_constraints()),
+                nmos::experimental::make_control_class_property(U("Example object sequence property"), { 3, 14 }, object_sequence, U("ExampleDataType"), false, false, true)
+            };
+
+            auto example_method_with_no_args = [](nmos::resources& resources, const nmos::resource& resource, int32_t handle, const web::json::value& arguments, bool is_deprecated, nmos::get_control_protocol_class_handler get_control_protocol_class, nmos::get_control_protocol_datatype_handler get_control_protocol_datatype, nmos::control_protocol_property_changed_handler, slog::base_gate& gate)
```
maweit, line 1011: The prototype is not consistent: all types follows a name except for nmos::control_protocol_property_changed_handler. Is there a reason for it?

Development/nmos-cpp-node/node_implementation.cpp
```diff
@@ -894,6 +920,374 @@ void node_implementation_init(nmos::node_model& model, slog::base_gate& gate)
...
+            {
+                // note, model mutex is already locked by the outter function, so access to control_protocol_resources is OK...
+
+                slog::log<slog::severities::more_info>(gate, SLOG_FLF) << "Executing the example method with no arguments";
+
+                return nmos::make_control_protocol_message_response(handle, { is_deprecated ? nmos::nc_method_status::method_deprecated : nmos::nc_method_status::ok });
+            };
+            auto example_method_with_simple_args = [](nmos::resources& resources, const nmos::resource& resource, int32_t handle, const web::json::value& arguments, bool is_deprecated, nmos::get_control_protocol_class_handler get_control_protocol_class, nmos::get_control_protocol_datatype_handler get_control_protocol_datatype, nmos::control_protocol_property_changed_handler, slog::base_gate& gate)
```
maweit, line 1019: Same as above and more follow below.

Development/nmos/control_protocol_utils.cpp
```diff
@@ -0,0 +1,648 @@
...
+                }
+            }
+        }
+    }
+
+    // push a control protocol resource into other control protocol NcBlock resource
+    void push_back(control_protocol_resource& nc_block_resource, const control_protocol_resource& resource)
+    {
+        using web::json::value;
+
```
maweit, line 554: Should there be  a write lock?

Development/nmos/control_protocol_utils.cpp
```diff
@@ -0,0 +1,648 @@
...
+        if (resources.end() == found || !found->has_data()) return false;
+
+        auto pre = found->data;
+
+        // "If an exception is thrown by some user-provided operation, then the element pointed to by position is erased."
+        // This seems too surprising, despite the fact that it means that a modification may have been partially completed,
+        // so capture and rethrow.
+        // See https://www.boost.org/doc/libs/1_68_0/libs/multi_index/doc/reference/ord_indices.html#modify
+        std::exception_ptr modifier_exception;
+
```
maweit, line 579: Should there also be a write lock?

Development/nmos/control_protocol_utils.cpp
lo-simon: This function is only used by the `node_implementation_init`, which has already had the write lock protecting it.

Will add a note in the function to remind the user to set the write lock in the outer function.

Development/nmos/control_protocol_utils.cpp
lo-simon: The write lock is already in the outer function to protect it.

Will add a note in the function to remind the user to set the write lock in the outer function.

Development/nmos-cpp-node/node_implementation.cpp
lo-simon: Clean up the non-standard class's method by removing those un-used `control_protocol_property_changed`, `get_control_protocol_class_` and  `get_control_protocol_datatype` handlers.

Development/nmos-cpp-node/config.json
```diff
@@ -367,5 +372,19 @@
     // If the Resource Server fails to verify a token using all public keys available it MUST reject the token."
     //"service_unavailable_retry_after": 5,
 
+    // manufacturer_name [node]: the manufacturer name of the NcDeviceManager used for NMOS Control Protocol
+    // See https://specs.amwa.tv/ms-05-02/branches/v1.0.x/docs/Framework.html#ncdevicemanager
+    //"manufacturer_name": "",
+
+    // product_name/product_key/product_revision_level [node]: the product description of the NcDeviceManager used for NMOS Control Protocol
```
jonathan-r-thorpe, line 379:
```suggestion
    // product_name/product_key/product_revision_level [node]: the product description reported by the NcDeviceManager when using the NMOS Control Protocol
```

Development/nmos-cpp-node/config.json
```diff
@@ -367,5 +372,19 @@
...
+    // See https://specs.amwa.tv/ms-05-02/branches/v1.0.x/docs/Framework.html#ncproduct
+    //"product_name": "",
+    //"product_key": "",
+    //"product_revision_level": "",
+
+    // serial_number [node]: the serial number of the NcDeviceManager used for NMOS Control Protocol
```
jonathan-r-thorpe, line 385:
```suggestion
    // serial_number [node]: the serial number reported by the NcDeviceManager when using the NMOS Control Protocol
```

Development/nmos/control_protocol_state.cpp
```diff
@@ -0,0 +1,360 @@
...
+
+namespace nmos
+{
+    namespace experimental
+    {
+        namespace details
+        {
+            // create control class
+            //   where
+            //     properties: vector of NcPropertyDescriptor can be constructed using make_control_class_property
```
jonathan-r-thorpe, line 14: Comment feels ambiguous; is it an NcPropertyDescriptor or _vector_ of NcPropertyDescriptor that is created by make_control_class_property?
```suggestion
            //     properties: vector of NcPropertyDescriptor where NcPropertyDescriptor can be constructed using make_control_class_property
```

Development/nmos/control_protocol_state.cpp
```diff
@@ -0,0 +1,360 @@
...
+            //     methods:    vector of NcMethodDescriptor vs assoicated method handler where NcMethodDescriptor can be constructed using make_nc_method_descriptor
+            //     events:     vector of NcEventDescriptor can be constructed using make_nc_event_descriptor
```
jonathan-r-thorpe, line 16: Similar ambiguity to above I think.
```suggestion
            //     events:     vector of NcEventDescriptor where NcEventDescriptor can be constructed using make_nc_event_descriptor
```

Development/nmos/control_protocol_state.cpp
```diff
@@ -0,0 +1,360 @@
...
+            control_class make_control_class(const utility::string_t& description, const nc_class_id& class_id, const nc_name& name, const web::json::value& fixed_role, const std::vector<web::json::value>& properties_, const std::vector<nmos::experimental::method>& methods_, const std::vector<web::json::value>& events_)
```
jonathan-r-thorpe, line 17: Confused by two method signatures  (here and line 34) that are identical apart from the data type of the fixed role. Perhaps a comment to explain when the fixed role should be a JSON value and when a string.

Development/nmos/control_protocol_state.cpp
```diff
@@ -0,0 +1,360 @@
...
+                for (const auto& property : properties_) { web::json::push_back(properties, property); }
+                web::json::value events = value::array();
+                for (const auto& event : events_) { web::json::push_back(events, event); }
+
+                return { description, class_id, name, fixed_role, properties, methods_, events };
+            }
+        }
+        // create control class with fixed role
+        //   where
+        //     properties: vector of NcPropertyDescriptor which can be constructed using make_control_class_property
```
jonathan-r-thorpe, line 31:
```suggestion
        //     properties: vector of NcPropertyDescriptor where NcPropertyDescriptor can be constructed using make_control_class_property
```

Development/nmos/control_protocol_state.cpp
```diff
@@ -0,0 +1,360 @@
...
+        //     methods:    vector of NcMethodDescriptor which can be constructed using make_nc_method_descriptor and the assoicated method handler
```
jonathan-r-thorpe, line 32:
```suggestion
        //     methods:    vector of NcMethodDescriptor where NcMethodDescriptor  can be constructed using make_nc_method_descriptor and the assoicated method handler
```

Development/nmos/control_protocol_state.cpp
```diff
@@ -0,0 +1,360 @@
...
+        //     events:     vector of NcEventDescriptor can be constructed using make_nc_event_descriptor
```
jonathan-r-thorpe, line 33:
```suggestion
        //     events:     vector of NcEventDescriptor where NcEventDescriptor can be constructed using make_nc_event_descriptor
```

Development/nmos/control_protocol_state.cpp
```diff
@@ -0,0 +1,360 @@
...
+        control_class make_control_class(const utility::string_t& description, const nc_class_id& class_id, const nc_name& name, const utility::string_t& fixed_role, const std::vector<web::json::value>& properties, const std::vector<nmos::experimental::method>& methods, const std::vector<web::json::value>& events)
+        {
+            using web::json::value;
+
+            return details::make_control_class(description, class_id, name, value::string(fixed_role), properties, methods, events);
+        }
+        // create control class with no fixed role
+        //   where
+        //     properties: vector of NcPropertyDescriptor which can be constructed using make_control_class_property
+        //     methods:    vector of NcMethodDescriptor which can be constructed using make_nc_method_descriptor and the assoicated method handler
```
jonathan-r-thorpe, line 43:
```suggestion
        //     methods:    vector of NcMethodDescriptor where NcMethodDescriptor can be constructed using make_nc_method_descriptor and the assoicated method handler
```

Development/nmos/control_protocol_state.cpp
```diff
@@ -0,0 +1,360 @@
...
+        //     events:     vector of NcEventDescriptor can be constructed using make_nc_event_descriptor
```
jonathan-r-thorpe, line 44:
```suggestion
        //     events:     vector of NcEventDescriptor where NcEventDescriptor can be constructed using make_nc_event_descriptor
```

Development/nmos/control_protocol_state.cpp
```diff
@@ -0,0 +1,360 @@
...
+            {
+                using web::json::value;
+
+                value parameters = value::array();
+                for (const auto& parameter : parameters_) { web::json::push_back(parameters, parameter); }
+
+                return nmos::details::make_nc_method_descriptor(description, id, name, result_datatype, parameters, is_deprecated);
+            }
+        }
+        // create control class method
```
jonathan-r-thorpe, line 76:
```suggestion
        // create standard control class method
```

Development/nmos/control_protocol_state.cpp
```diff
@@ -0,0 +1,360 @@
...
+        method make_control_class_method(const utility::string_t& description, const nc_method_id& id, const nc_name& name, const utility::string_t& result_datatype, const std::vector<web::json::value>& parameters, bool is_deprecated, standard_method_handler method_handler)
+        {
+            return make_control_class_standard_method(details::make_control_class_method(description, id, name, result_datatype, parameters, is_deprecated), method_handler);
+        }
+        method make_control_class_method(const utility::string_t& description, const nc_method_id& id, const nc_name& name, const utility::string_t& result_datatype, const std::vector<web::json::value>& parameters, bool is_deprecated, non_standard_method_handler method_handler)
```
jonathan-r-thorpe, line 81:
```suggestion
        // create non-standard control class method
        method make_control_class_method(const utility::string_t& description, const nc_method_id& id, const nc_name& name, const utility::string_t& result_datatype, const std::vector<web::json::value>& parameters, bool is_deprecated, non_standard_method_handler method_handler)
```

Development/nmos/control_protocol_state.cpp
jonathan-r-thorpe: Ignore this comment - I realised that one function delegates to the other (DOH!)

Development/nmos/control_protocol_state.cpp
```diff
@@ -0,0 +1,360 @@
...
+        //     properties: vector of NcPropertyDescriptor which can be constructed using make_control_class_property
+        //     methods:    vector of NcMethodDescriptor which can be constructed using make_nc_method_descriptor and the assoicated method handler
+        //     events:     vector of NcEventDescriptor can be constructed using make_nc_event_descriptor
+        control_class make_control_class(const utility::string_t& description, const nc_class_id& class_id, const nc_name& name, const utility::string_t& fixed_role, const std::vector<web::json::value>& properties, const std::vector<nmos::experimental::method>& methods, const std::vector<web::json::value>& events)
+        {
+            using web::json::value;
+
+            return details::make_control_class(description, class_id, name, value::string(fixed_role), properties, methods, events);
+        }
+        // create control class with no fixed role
```
jonathan-r-thorpe, line 40:
```suggestion
        // create control class without fixed role
```

Development/nmos/control_protocol_state.cpp
```diff
@@ -0,0 +1,360 @@
...
+                if (!data.is_null())
+                {
+                    return std::vector<web::json::value>(data.as_array().begin(), data.as_array().end());
+                }
+                return std::vector<web::json::value>{};
+            };
+
+            auto to_methods_vector = [](const web::json::value& nc_method_descriptors, const std::map<nmos::nc_method_id, const standard_method_handler>& method_handlers)
+            {
+                // NcMethodDescriptor vs method
```
jonathan-r-thorpe, line 107: vs?
```suggestion
                // NcMethodDescriptor
```

Development/nmos/control_protocol_state.cpp
```diff
@@ -0,0 +1,360 @@
...
+                    }
+                }
+                return methods;
+            };
+
+            // setup the core control classes
+            control_classes =
+            {
+                // Control class models
+                // See https://specs.amwa.tv/ms-05-02/branches/v1.0.x/models/classes/#control-class-models-for-branch-v10-dev
```
jonathan-r-thorpe, line 124:
```suggestion
                // See https://specs.amwa.tv/ms-05-02/branches/v1.0.x/models/classes/
```

Development/nmos/control_protocol_state.cpp
```diff
@@ -0,0 +1,360 @@
...
+                {
+                    for (const auto& nc_method_descriptor : nc_method_descriptors.as_array())
+                    {
+                        methods.push_back(make_control_class_standard_method(nc_method_descriptor, method_handlers.at(nmos::details::parse_nc_method_id(nmos::fields::nc::id(nc_method_descriptor)))));
+                    }
+                }
+                return methods;
+            };
+
+            // setup the core control classes
```
jonathan-r-thorpe, line 120:
```suggestion
            // setup the standard control classes
```

Development/nmos/control_protocol_state.cpp
```diff
@@ -0,0 +1,360 @@
...
+                    // NcClassManager methods
+                    to_methods_vector(make_nc_class_manager_methods(),
+                    {
+                        // link NcClassManager method_ids with method functions
+                        { nc_class_manager_get_control_class_method_id, get_control_class },
+                        { nc_class_manager_get_datatype_method_id, get_datatype }
+                    }),
+                    // NcClassManager events
+                    to_vector(make_nc_class_manager_events())) },
+                // identification beacon model
```
jonathan-r-thorpe, line 196:
```suggestion
                // Identification feature set
```

Development/nmos/control_protocol_state.cpp
```diff
@@ -0,0 +1,360 @@
...
+                // See https://specs.amwa.tv/nmos-control-feature-sets/branches/main/identification/#ncidentbeacon
+                // NcIdentBeacon
+                { nc_ident_beacon_class_id, make_control_class(U("NcIdentBeacon class descriptor"), nc_ident_beacon_class_id, U("NcIdentBeacon"),
+                    // NcIdentBeacon properties
+                    to_vector(make_nc_ident_beacon_properties()),
+                    // NcIdentBeacon methods
+                    to_methods_vector(make_nc_ident_beacon_methods(), {}),
+                    // NcIdentBeacon events
+                    to_vector(make_nc_ident_beacon_events())) },
+                // Monitoring
```
jonathan-r-thorpe, line 206:
```suggestion
                // Monitoring feature set
```

Development/nmos/control_protocol_state.cpp
```diff
@@ -0,0 +1,360 @@
...
+                    to_methods_vector(make_nc_class_manager_methods(),
+                    {
+                        // link NcClassManager method_ids with method functions
+                        { nc_class_manager_get_control_class_method_id, get_control_class },
+                        { nc_class_manager_get_datatype_method_id, get_datatype }
+                    }),
+                    // NcClassManager events
+                    to_vector(make_nc_class_manager_events())) },
+                // identification beacon model
+                // See https://specs.amwa.tv/nmos-control-feature-sets/branches/main/identification/#ncidentbeacon
```
jonathan-r-thorpe, line 197:
```suggestion
                // See https://specs.amwa.tv/nmos-control-feature-sets/branches/main/identification/
```

Development/nmos/control_protocol_state.cpp
```diff
@@ -0,0 +1,360 @@
...
+                { nc_receiver_monitor_protected_class_id, make_control_class(U("NcReceiverMonitorProtected class descriptor"), nc_receiver_monitor_protected_class_id, U("NcReceiverMonitorProtected"),
+                    // NcReceiverMonitorProtected properties
+                    to_vector(make_nc_receiver_monitor_protected_properties()),
+                    // NcReceiverMonitorProtected methods
+                    to_methods_vector(make_nc_receiver_monitor_protected_methods(), {}),
+                    // NcReceiverMonitorProtected events
+                    to_vector(make_nc_receiver_monitor_protected_events())) }
+            };
+
+            // setup the core datatypes
```
jonathan-r-thorpe, line 226:
```suggestion
            // setup the standard datatypes
```

Development/nmos/control_protocol_state.cpp
```diff
@@ -0,0 +1,360 @@
...
+            datatypes =
+            {
+                // Datatype models
+                // See https://specs.amwa.tv/ms-05-02/branches/v1.0.x/models/datatypes/#datatype-models-for-branch-v10-dev
```
jonathan-r-thorpe, line 230:
```suggestion
                // See https://specs.amwa.tv/ms-05-02/branches/v1.0.x/models/datatypes/
```

Development/nmos/control_protocol_state.cpp
```diff
@@ -0,0 +1,360 @@
...
+                { U("NcRolePath"), {make_nc_role_path_datatype()} },
+                { U("NcParameterConstraintsNumber"), {make_nc_parameter_constraints_number_datatype()} },
+                { U("NcParameterConstraintsString"), {make_nc_parameter_constraints_string_datatype()} },
+                { U("NcTimeInterval"), {make_nc_time_interval_datatype()} },
+                { U("NcTouchpointNmos"), {make_nc_touchpoint_nmos_datatype()} },
+                { U("NcTouchpointNmosChannelMapping"), {make_nc_touchpoint_nmos_channel_mapping_datatype()} },
+                { U("NcTouchpointResource"), {make_nc_touchpoint_resource_datatype()} },
+                { U("NcTouchpointResourceNmos"), {make_nc_touchpoint_resource_nmos_datatype()} },
+                { U("NcTouchpointResourceNmosChannelMapping"), {make_nc_touchpoint_resource_nmos_channel_mapping_datatype()} },
+                // Monitoring
```
jonathan-r-thorpe, line 299:
```suggestion
                // Monitoring feature set
```

Development/nmos/control_protocol_state.cpp
```diff
@@ -0,0 +1,360 @@
...
+                // See https://specs.amwa.tv/nmos-control-feature-sets/branches/main/monitoring/#datatypes
+                { U("NcConnectionStatus"), {make_nc_connection_status_datatype()} },
+                { U("NcPayloadStatus"), {make_nc_payload_status_datatype()} }
+            };
+        }
+
+        // insert control class, false if class already presented
```
jonathan-r-thorpe, line 306:
```suggestion
        // insert control class, false if class already inserted
```

Development/nmos/control_protocol_state.cpp
```diff
@@ -0,0 +1,360 @@
...
+
+            if (control_classes.end() != control_classes.find(class_id))
+            {
+                control_classes.erase(class_id);
+                return true;
+            }
+            return false;
+        }
+
+        // insert datatype, false if datatype already presented
```
jonathan-r-thorpe, line 332:
```suggestion
        // insert datatype, false if datatype already inserted
```

Development/nmos/control_protocol_resource.h
```diff
@@ -0,0 +1,425 @@
...
+{
+    namespace experimental
+    {
+        struct control_protocol_state;
+    }
+
+    namespace details
+    {
+        // See https://specs.amwa.tv/ms-05-02/branches/v1.0.x/docs/Framework.html#ncelementid
+        //web::json::value make_nc_element_id(uint16_t level, uint16_t index);
```
jonathan-r-thorpe, line 44: Remove?

Development/nmos/control_protocol_resource.h
```diff
@@ -0,0 +1,425 @@
...
+        web::json::value make_nc_element_id(const nc_element_id& element_id);
+        nc_element_id parse_nc_element_id(const web::json::value& element_id);
+
+        // See https://specs.amwa.tv/ms-05-02/branches/v1.0.x/docs/Framework.html#nceventid
+        //web::json::value make_nc_event_id(uint16_t level, uint16_t index);
```
jonathan-r-thorpe, line 49: Remove commented out signature?

Development/nmos/control_protocol_resource.h
```diff
@@ -0,0 +1,425 @@
...
+    web::json::value make_nc_device_manager_events();
+    // See https://specs.amwa.tv/ms-05-02/branches/v1.0.x/docs/Framework.html#ncclassmanager
+    web::json::value make_nc_class_manager_properties();
+    web::json::value make_nc_class_manager_methods();
+    web::json::value make_nc_class_manager_events();
+    // See https://specs.amwa.tv/nmos-control-feature-sets/branches/main/monitoring/#ncreceivermonitor
+    web::json::value make_nc_receiver_monitor_properties();
+    web::json::value make_nc_receiver_monitor_methods();
+    web::json::value make_nc_receiver_monitor_events();
+    // See https://specs.amwa.tv/nmos-control-feature-sets/branches/main/monitoring/#ncreceivermonitorprotected
```
jonathan-r-thorpe, line 267:
```suggestion
    
    // Monitoring feature set control classes
    // See https://specs.amwa.tv/nmos-control-feature-sets/branches/main/monitoring/#ncreceivermonitorprotected
```

Development/nmos/control_protocol_resource.h
```diff
@@ -0,0 +1,425 @@
...
+    web::json::value make_nc_receiver_monitor_protected_properties();
+    web::json::value make_nc_receiver_monitor_protected_methods();
+    web::json::value make_nc_receiver_monitor_protected_events();
+    // See https://specs.amwa.tv/nmos-control-feature-sets/branches/main/identification/#ncidentbeacon
```
jonathan-r-thorpe, line 271:
```suggestion

    // Identification feature set control classes
    // See https://specs.amwa.tv/nmos-control-feature-sets/branches/main/identification/#ncidentbeacon
```

Development/nmos/control_protocol_resource.h
```diff
@@ -0,0 +1,425 @@
...
+    // See https://specs.amwa.tv/ms-05-02/branches/v1.0.x/models/datatypes/NcTouchpointResourceNmosChannelMapping.html
+    web::json::value make_nc_touchpoint_resource_nmos_channel_mapping_datatype();
+    // See // See https://specs.amwa.tv/ms-05-02/branches/v1.0.x/models/datatypes/NcUri.html
+    web::json::value make_nc_uri_datatype();
+    // See https://specs.amwa.tv/ms-05-02/branches/v1.0.x/models/datatypes/NcUuid.html
+    web::json::value make_nc_uuid_datatype();
+    // See https://specs.amwa.tv/ms-05-02/branches/v1.0.x/models/datatypes/NcVersionCode.html
+    web::json::value make_nc_version_code_datatype();
+
+    // Monitoring datatypes
```
jonathan-r-thorpe, line 416:
```suggestion
    // Monitoring feature set datatypes
```

Development/nmos/control_protocol_resource.cpp
```diff
@@ -0,0 +1,2056 @@
...
+
+    // See https://specs.amwa.tv/ms-05-02/branches/v1.0.x/models/classes/1.3.2.html
+    web::json::value make_nc_class_manager_class()
+    {
+        using web::json::value;
+
+        return details::make_nc_class_descriptor(U("NcClassManager class descriptor"), nc_class_manager_class_id, U("NcClassManager"), U("ClassManager"), make_nc_class_manager_properties(), make_nc_class_manager_methods(), make_nc_class_manager_events());
+    }
+
+    // See https://specs.amwa.tv/nmos-control-feature-sets/branches/main/identification/#ncidentbeacon
```
jonathan-r-thorpe, line 1287:
```suggestion

    // Identification feature set control classes
    // See https://specs.amwa.tv/nmos-control-feature-sets/branches/main/identification/#ncidentbeacon
```

Development/nmos/control_protocol_resource.cpp
```diff
@@ -0,0 +1,2056 @@
...
+    web::json::value make_nc_ident_beacon_class()
+    {
+        using web::json::value;
+
+        return details::make_nc_class_descriptor(U("NcIdentBeacon class descriptor"), nc_ident_beacon_class_id, U("NcIdentBeacon"), make_nc_ident_beacon_properties(), make_nc_ident_beacon_methods(), make_nc_ident_beacon_events());
+    }
+
+    // See https://specs.amwa.tv/nmos-control-feature-sets/branches/main/monitoring/#ncreceivermonitor
```
jonathan-r-thorpe, line 1295:
```suggestion

    // Monitoring feature set control classes
    // See https://specs.amwa.tv/nmos-control-feature-sets/branches/main/monitoring/#ncreceivermonitor
```

Development/nmos/control_protocol_resource.cpp
```diff
@@ -0,0 +1,2056 @@
...
+
+    // See https://specs.amwa.tv/nmos-control-feature-sets/branches/main/monitoring/#ncreceivermonitorprotected
+    web::json::value make_nc_receiver_monitor_protected_class()
+    {
+        using web::json::value;
+
+        return details::make_nc_class_descriptor(U("NcReceiverMonitorProtected class descriptor"), nc_receiver_monitor_protected_class_id, U("NcReceiverMonitorProtected"), make_nc_receiver_monitor_protected_properties(), make_nc_receiver_monitor_protected_methods(), make_nc_receiver_monitor_protected_events());
+    }
+
+    // See https://specs.amwa.tv/ms-05-02/branches/v1.0.x/docs/Framework.html#primitives
```
jonathan-r-thorpe, line 1311:
```suggestion

    // Primitive datatypes
    // See https://specs.amwa.tv/ms-05-02/branches/v1.0.x/docs/Framework.html#primitives
```

Development/nmos/control_protocol_resources.cpp
jonathan-r-thorpe, line 1: For completeness should we not have, in addition to the helpers already here:
- make_receiver_monitor_protected
- make_ident_beacon
?

Development/nmos/control_protocol_state.cpp
lo-simon: "NcMethodDescriptor, method handler" structure array

jonathan-r-thorpe: "NcMethodDescriptor method handler array" perhaps?

Development/nmos-cpp-node/node_implementation.cpp
```diff
@@ -894,6 +920,374 @@ void node_implementation_init(nmos::node_model& model, slog::base_gate& gate)
...
+            value::null(),
+            // specify the level 2: runtime constraints, see https://specs.amwa.tv/ms-05-02/branches/v1.0.x/docs/Constraints.html
+            // use of make_nc_property_constraints_string and make_nc_property_constraints_number to create runtime constraints
+            value_of({
+                { nmos::details::make_nc_property_constraints_string({3, 2}, 5, U("^[a-z]+$")) },
+                { nmos::details::make_nc_property_constraints_number({3, 3}, 10, 100, 2) }
+            }),
+            example_enum::Undefined,
+            U("test"),
+            3,
```
jonathan-r-thorpe, line 1240: This is a bug - the numberProperty is initialized to 3, but the runtime constraints specify a step of 2.  When the "invasive" tests run it stores the original value (3) and attempts to reset to the original value at the end of the test, which fails as it violates the runtime constraints.
```suggestion
            4,
```

Development/nmos/control_protocol_typedefs.h
```diff
@@ -0,0 +1,386 @@
...
+    // See https://specs.amwa.tv/ms-05-02/branches/v1.0.x/docs/Framework.html#ncmethodstatus
+    namespace nc_method_status
+    {
+        enum status
+        {
+            ok = 200,                       // Method call was successful
+            property_deprecated = 298,      // Method call was successful but targeted property is deprecated
+            method_deprecated = 299,        // Method call was successful but method is deprecated
+            bad_command_format = 400,       // Badly-formed command
+            unathorized = 401,              // Client is not authorized
```
jonathan-r-thorpe, line 35: Typo
```suggestion
            unauthorized = 401,              // Client is not authorized
```

Development/nmos/control_protocol_typedefs.h
```diff
@@ -0,0 +1,386 @@
...
+        friend bool operator==(const nc_element_id& lhs, const nc_element_id& rhs) { return lhs.tied() == rhs.tied(); }
+        friend bool operator!=(const nc_element_id& lhs, const nc_element_id& rhs) { return !(lhs == rhs); }
+        friend bool operator<(const nc_element_id& lhs, const nc_element_id& rhs) { return lhs.tied() < rhs.tied(); }
+    };
+
+    // NcEventId
+    // See https://specs.amwa.tv/ms-05-02/branches/v1.0.x/docs/Framework.html#nceventid
+    typedef nc_element_id nc_event_id;
+    // NcEventIds for NcObject
+    // SEe https://specs.amwa.tv/ms-05-02/branches/v1.0.x/docs/Framework.html#ncobject
```
jonathan-r-thorpe, line 149:
```suggestion
    // See https://specs.amwa.tv/ms-05-02/branches/v1.0.x/docs/Framework.html#ncobject
```

Development/nmos/control_protocol_ws_api.cpp
```diff
@@ -0,0 +1,537 @@
...
+                                    auto resource = nmos::find_resource(resources, utility::s2us(std::to_string(oid)));
+                                    if (resources.end() != resource)
+                                    {
+                                        const auto& class_id = nmos::details::parse_nc_class_id(nmos::fields::nc::class_id(resource->data));
+
+                                        // find the relevent method handler to execute
+                                        auto method = get_control_protocol_method(class_id, method_id);
+                                        auto& nc_method_descriptor = std::get<0>(method);
+                                        auto& standard_method = std::get<1>(method);
+                                        auto& non_standard_method = std::get<2>(method);
```
jonathan-r-thorpe, line 265: Is it worth describing the structure of the `method` variable here?
By inspection is seems that the first element is the NcMethodDescriptor, and the next two elements are the method handlers - is this structure described/explained elsewhere?

Development/nmos/control_protocol_handlers.cpp
```diff
@@ -0,0 +1,100 @@
...
+            if (resources.end() != found && nc_receiver_monitor_class_id == details::parse_nc_class_id(nmos::fields::nc::class_id(found->data)))
+            {
+                // update receiver-monitor's connectionStatus propertry
+
+                const auto active = nmos::fields::master_enable(nmos::fields::endpoint_active(connection_resource.data));
+                const web::json::value value = active ? nc_connection_status::connected : nc_connection_status::disconnected;
+
+                // hmm, maybe updating connectionStatusMessage, payloadStatus, and payloadStatusMessage too
+
+                const auto propertry_changed_event = make_propertry_changed_event(nmos::fields::nc::oid(found->data),
```
jonathan-r-thorpe, line 87: Typo
```suggestion
                const auto property_changed_event = make_propertry_changed_event(nmos::fields::nc::oid(found->data),
```

Development/nmos/control_protocol_methods.cpp
```diff
@@ -0,0 +1,628 @@
...
+                {
+                    resource.data[nmos::fields::nc::name(property)] = val;
+
+                    // do notification that the specified property has changed
+                    if (property_changed)
+                    {
+                        property_changed(resource, nmos::fields::nc::name(property), -1);
+                    }
+
+                }, make_propertry_changed_event(nmos::fields::nc::oid(resource.data), { { property_id_, nc_property_change_type::type::value_changed, val } }));
```
jonathan-r-thorpe, line 79: Typo
```suggestion
                }, make_property_changed_event(nmos::fields::nc::oid(resource.data), { { property_id_, nc_property_change_type::type::value_changed, val } }));
```

Development/nmos/control_protocol_methods.cpp
```diff
@@ -0,0 +1,628 @@
...
+                    {
+                        resource.data[nmos::fields::nc::name(property)][index] = val;
+
+                        // do notification that the specified property has changed
+                        if (property_changed)
+                        {
+                            property_changed(resource, nmos::fields::nc::name(property), index);
+                        }
+
+                    }, make_propertry_changed_event(nmos::fields::nc::oid(resource.data), { { property_id_, nc_property_change_type::type::sequence_item_changed, val, nc_id(index) } }));
```
jonathan-r-thorpe, line 187: Typo:
```suggestion
                    }, make_propertry_changed_event(nmos::fields::nc::oid(resource.data), { { property_id_, nc_property_change_type::type::sequence_item_changed, val, nc_id(index) } }));
```

Development/nmos/control_protocol_methods.cpp
```diff
@@ -0,0 +1,628 @@
...
+                    if (data.is_null()) { sequence = value::array(); }
+                    web::json::push_back(sequence, val);
+
+                    // do notification that the specified property has changed
+                    if (property_changed)
+                    {
+                        property_changed(resource, nmos::fields::nc::name(property), (int)sequence.as_array().size()-1);
+                    }
+
+                }, make_propertry_changed_event(nmos::fields::nc::oid(resource.data), { { property_id_, nc_property_change_type::type::sequence_item_added, val, sequence_item_index } }));
```
jonathan-r-thorpe, line 263:
```suggestion
                }, make_propertry_changed_event(nmos::fields::nc::oid(resource.data), { { property_id_, nc_property_change_type::type::sequence_item_added, val, sequence_item_index } }));
```

Development/nmos/control_protocol_methods.cpp
```diff
@@ -0,0 +1,628 @@
...
+            }
+
+            if (data.as_array().size() > (size_t)index)
+            {
+                modify_control_protocol_resource(resources, resource.id, [&](nmos::resource& resource)
+                {
+                    auto& sequence = resource.data[nmos::fields::nc::name(property)].as_array();
+                    sequence.erase(index);
+
+                }, make_propertry_changed_event(nmos::fields::nc::oid(resource.data), { { details::parse_nc_property_id(property_id), nc_property_change_type::type::sequence_item_removed, nc_id(index) } }));
```
jonathan-r-thorpe, line 312: Typo:
```suggestion
                }, make_property_changed_event(nmos::fields::nc::oid(resource.data), { { details::parse_nc_property_id(property_id), nc_property_change_type::type::sequence_item_removed, nc_id(index) } }));
```

Development/nmos/control_protocol_resource.h
```diff
@@ -0,0 +1,427 @@
...
+    // notification
+    // See https://specs.amwa.tv/ms-05-01/branches/v1.0.x/docs/Core_Mechanisms.html#notification-messages
+    // See https://specs.amwa.tv/is-12/branches/v1.0.x/docs/Protocol_messaging.html#notification-message-type
+    web::json::value make_control_protocol_notification(nc_oid oid, const nc_event_id& event_id, const nc_property_changed_event_data& property_changed_event_data);
+    web::json::value make_control_protocol_notification_message(const web::json::value& notifications);
+
+    // property changed notification event
+    // See https://specs.amwa.tv/ms-05-01/branches/v1.0.x/docs/Core_Mechanisms.html#the-propertychanged-event
+    // See https://specs.amwa.tv/ms-05-02/branches/v1.0.x/docs/NcObject.html#propertychanged-event
+    web::json::value make_propertry_changed_event(nc_oid oid, const std::vector<nc_property_changed_event_data>& property_changed_event_data_list);
```
jonathan-r-thorpe, line 208: Typo:
```suggestion
    web::json::value make_property_changed_event(nc_oid oid, const std::vector<nc_property_changed_event_data>& property_changed_event_data_list);
```

Development/nmos-cpp-node/node_implementation.cpp
```diff
@@ -956,6 +1351,33 @@ void node_implementation_run(nmos::node_model& model, slog::base_gate& gate)
...
+                auto& resources = model.control_protocol_resources;
+
+                auto found = nmos::find_resource_if(resources, nmos::types::nc_worker, [&temperature_sensor_control_class_id](const nmos::resource& resource)
+                {
+                    return temperature_sensor_control_class_id == nmos::details::parse_nc_class_id(nmos::fields::nc::class_id(resource.data));
+                });
+
+                if (resources.end() != found)
+                {
+                    const auto propertry_changed_event = nmos::make_propertry_changed_event(nmos::fields::nc::oid(found->data),
```
jonathan-r-thorpe, line 1368: Typo:
```suggestion
                    const auto property_changed_event = nmos::make_property_changed_event(nmos::fields::nc::oid(found->data),
```

Development/nmos/control_protocol_resources.cpp
lo-simon: well done by @jonathan-r-thorpe

Development/nmos/control_protocol_ws_api.cpp
lo-simon: described in control_protocol_handlers.h 

> // method definition (NcMethodDescriptor vs method handler)
> typedef std::tuple<web::json::value, standard_method_handler, non_standard_method_handler> method;

jonathan-r-thorpe: Perhaps put exactly that wording as a a comment?
```suggestion
                                        // find the relevent method handler to execute
                                        // method tuple definition described in control_protocol_handlers.h
                                        auto method = get_control_protocol_method(class_id, method_id);
                                        auto& nc_method_descriptor = std::get<0>(method);
                                        auto& standard_method = std::get<1>(method);
                                        auto& non_standard_method = std::get<2>(method);
```

Development/nmos/json_fields.h
```diff
@@ -230,6 +230,114 @@ namespace nmos
         const web::json::field_as_string hostname{ U("hostname") }; // hostname, ipv4 or ipv6
         const web::json::field_as_integer port{ U("port") }; // 1..65535
 
+        // IS-12 Control Protocol
+        namespace nc
```
jonathan-r-thorpe, line 234: Hmmmm,  Not sure what the right thing to do here is with the namespace name:

`ncp` is what's used in the IS-12 spec to describe the protocol.  However, this namespace also includes the MS-05 field definitions.

So do we state that the `nc` namespace covers IS-12, MS-05 and in future IS-14?
Or do we segment the namespace further?

I would lean towards stating that `nc` covers both IS-12 and MS-05 perhaps...
```suggestion
        // IS-12 Control Protocol and MS-05 model definitions
        namespace nc
```

Development/third_party/is-12/README.md
jonathan-r-thorpe, line 1: This file is an artifact from the IS-12 repo.  It's the boilerplate README.md that should have been removed and is therefore redundant and doesn't need to be imported here.

Development/nmos/control_protocol_utils.h
```diff
@@ -0,0 +1,84 @@
...
+    struct control_protocol_resource;
+
+    struct control_protocol_exception : std::runtime_error
+    {
+        control_protocol_exception(const std::string& message) : std::runtime_error(message) {}
+    };
+
+    namespace details
+    {
+        // get the runtime property constraints of a given property_id
```
jonathan-r-thorpe, line 18:
```suggestion
        // get the runtime property constraints of a given property
```

Development/nmos-cpp-node/node_implementation.cpp
```diff
@@ -894,6 +920,374 @@ void node_implementation_init(nmos::node_model& model, slog::base_gate& gate)
...
+    {
+        // example to create a non-standard Gain control class
+        const auto gain_control_class_id = nmos::make_nc_class_id(nmos::nc_worker_class_id, 0, { 1 });
+        const web::json::field_as_number gain_value{ U("gainValue") };
+        {
+            // Gain control class properties
+            std::vector<web::json::value> gain_control_properties = { nmos::experimental::make_control_class_property(U("Gain value"), { 3, 1 }, gain_value, U("NcFloat32")) };
+
+            // create Gain control class
+            auto gain_control_class = nmos::experimental::make_control_class(U("Gain control class descriptor"), gain_control_class_id, U("GainControl"), gain_control_properties);
```
jonathan-r-thorpe, line 938: I'm getting confused between classes that create the class definitions and those that create the class instances.

In line 938 we have a reference to `make_control_class` that is creating a class descriptor (i.e. class definition) and in line 948 we have `make_nc_worker` that is creating a class instance of worker. 

Perhaps to avoid confusion the `make_control_class` should be `make_control_class_desc`?  This would relate back to the definition of functions in `control_protocol_state`

Development/nmos/control_protocol_state.cpp
```diff
@@ -0,0 +1,361 @@
...
+
+        // create control class method parameter
+        web::json::value make_control_class_method_parameter(const utility::string_t& description, const nc_name& name, const utility::string_t& type_name, bool is_nullable, bool is_sequence, const web::json::value& constraints)
+        {
+            return nmos::details::make_nc_parameter_descriptor(description, name, type_name, is_nullable, is_sequence, constraints);
+        }
+
+        namespace details
+        {
+            web::json::value make_control_class_method(const utility::string_t& description, const nc_method_id& id, const nc_name& name, const utility::string_t& result_datatype, const std::vector<web::json::value>& parameters_, bool is_deprecated)
```
jonathan-r-thorpe, line 66: similar to the above comment.  If we are returning a descriptor maybe this function should be called `make_control_class_method_descriptor`?

Development/nmos/control_protocol_state.cpp
```diff
@@ -0,0 +1,361 @@
...
+                return nmos::details::make_nc_method_descriptor(description, id, name, result_datatype, parameters, is_deprecated);
+            }
+        }
+        // create standard control class method
+        method make_control_class_method(const utility::string_t& description, const nc_method_id& id, const nc_name& name, const utility::string_t& result_datatype, const std::vector<web::json::value>& parameters, bool is_deprecated, standard_method_handler method_handler)
+        {
+            return make_control_class_standard_method(details::make_control_class_method(description, id, name, result_datatype, parameters, is_deprecated), method_handler);
+        }
+        // create non-standard control class method
+        method make_control_class_method(const utility::string_t& description, const nc_method_id& id, const nc_name& name, const utility::string_t& result_datatype, const std::vector<web::json::value>& parameters, bool is_deprecated, non_standard_method_handler method_handler)
```
jonathan-r-thorpe, line 82: Same as above

Development/nmos/control_protocol_state.cpp
```diff
@@ -0,0 +1,361 @@
...
+                    for (const auto& nc_method_descriptor : nc_method_descriptors.as_array())
+                    {
+                        methods.push_back(make_control_class_standard_method(nc_method_descriptor, method_handlers.at(nmos::details::parse_nc_method_id(nmos::fields::nc::id(nc_method_descriptor)))));
+                    }
+                }
+                return methods;
+            };
+
+            // setup the standard control classes
+            control_classes =
```
jonathan-r-thorpe, line 122: `control_class_descriptors`??

Development/nmos/control_protocol_state.cpp
```diff
@@ -0,0 +1,361 @@
...
+                { U("NcTouchpointResourceNmos"), {make_nc_touchpoint_resource_nmos_datatype()} },
+                { U("NcTouchpointResourceNmosChannelMapping"), {make_nc_touchpoint_resource_nmos_channel_mapping_datatype()} },
+                // Monitoring feature set
+                // See https://specs.amwa.tv/nmos-control-feature-sets/branches/main/monitoring/#datatypes
+                { U("NcConnectionStatus"), {make_nc_connection_status_datatype()} },
+                { U("NcPayloadStatus"), {make_nc_payload_status_datatype()} }
+            };
+        }
+
+        // insert control class, false if class already inserted
```
jonathan-r-thorpe, line 307:
```suggestion
        // insert control class descriptor, false if class already inserted
```

Development/nmos/control_protocol_state.h
```diff
@@ -0,0 +1,99 @@
...
+#include "nmos/control_protocol_typedefs.h"
+#include "nmos/mutex.h"
+
+namespace slog { class base_gate; }
+
+namespace nmos
+{
+    namespace experimental
+    {
+        struct control_class // NcClassDescriptor
```
jonathan-r-thorpe, line 16: To avoid confusion between class definitions and instances 

```suggestion
        struct control_class_descriptor // NcClassDescriptor
```

Development/nmos/control_protocol_state.h
```diff
@@ -0,0 +1,99 @@
...
+        {
+            utility::string_t description;
+            nmos::nc_class_id class_id;
+            nmos::nc_name name;
+            web::json::value fixed_role;
+
+            web::json::value properties = web::json::value::array(); // NcPropertyDescriptor array
+            std::vector<method> methods; // NcMethodDescriptor method handler array
+            web::json::value events = web::json::value::array();  // NcEventDescriptor array
```
jonathan-r-thorpe, line 25: Similar here, these arrays are for the descriptors not the instances:
```suggestion
            web::json::value property_descriptors = web::json::value::array(); // NcPropertyDescriptor array
            std::vector<method> method_descriptors; // NcMethodDescriptor method handler array
            web::json::value event_descriptors = web::json::value::array();  // NcEventDescriptor array
```

Development/nmos/control_protocol_state.h
```diff
@@ -0,0 +1,99 @@
...
+                , class_id(std::move(class_id))
+                , name(std::move(name))
+                , fixed_role(std::move(fixed_role))
+                , properties(std::move(properties))
+                , methods(std::move(methods))
+                , events(std::move(events))
+            {}
+        };
+
+        struct datatype // NcDatatypeDescriptorEnum/NcDatatypeDescriptorPrimitive/NcDatatypeDescriptorStruct/NcDatatypeDescriptorTypeDef
```
jonathan-r-thorpe, line 42:
```suggestion
        struct datatype_descriptor // NcDatatypeDescriptorEnum/NcDatatypeDescriptorPrimitive/NcDatatypeDescriptorStruct/NcDatatypeDescriptorTypeDef
```

Development/nmos/control_protocol_state.cpp
```diff
@@ -0,0 +1,361 @@
...
+    namespace experimental
+    {
+        namespace details
+        {
+            // create control class
+            //   where
+            //     properties: vector of NcPropertyDescriptor where NcPropertyDescriptor can be constructed using make_control_class_property
+            //     methods:    vector of NcMethodDescriptor vs assoicated method handler where NcMethodDescriptor can be constructed using make_nc_method_descriptor
+            //     events:     vector of NcEventDescriptor where NcEventDescriptor can be constructed using make_nc_event_descriptor
+            control_class make_control_class(const utility::string_t& description, const nc_class_id& class_id, const nc_name& name, const web::json::value& fixed_role, const std::vector<web::json::value>& properties_, const std::vector<nmos::experimental::method>& methods_, const std::vector<web::json::value>& events_)
```
jonathan-r-thorpe, line 17: Related to comments in `node_implementation.cpp` I think we need to distinguish between the class definitions (descriptors) and the class instances.  It's not clear whether a function called `make_control_class` is creating a definition or an instance.

My suggested solution would be either:
1. maintain `make_control_class` here and append `_instance` to the functions that produce instances of classes e.g. `make_nc_worker_instance`
2. call this `make_control_class_descriptor` to indicate it's creating a definition (descriptor) and keep the instance names as they are
I would favour 2 I think.

Development/nmos/control_protocol_state.cpp
jonathan-r-thorpe: This is along a theme of naming descriptors as descriptors so they don't get confused with class instances

Development/nmos-cpp-node/node_implementation.cpp
```diff
@@ -894,6 +920,374 @@ void node_implementation_init(nmos::node_model& model, slog::base_gate& gate)
...
+            using web::json::value_of;
+
+            return value_of({
+                { enum_property, enum_property_ },
+                { string_property, string_property_ },
+                { number_property, number_property_ },
+                { boolean_property, boolean_property_ }
+            });
+        };
+        // helper function to create Example control
```
jonathan-r-thorpe, line 1113:
```suggestion
        // helper function to create Example control instance
```

Development/nmos-cpp-node/node_implementation.cpp
```diff
@@ -894,6 +920,374 @@ void node_implementation_init(nmos::node_model& model, slog::base_gate& gate)
...
+            // Gain control class properties
+            std::vector<web::json::value> gain_control_properties = { nmos::experimental::make_control_class_property(U("Gain value"), { 3, 1 }, gain_value, U("NcFloat32")) };
+
+            // create Gain control class
+            auto gain_control_class = nmos::experimental::make_control_class(U("Gain control class descriptor"), gain_control_class_id, U("GainControl"), gain_control_properties);
+
+            // insert Gain control class to global state, which will be used by the control_protocol_ws_message_handler to process incoming ws message
+            control_protocol_state.insert(gain_control_class);
+        }
+        // helper function to create Gain control
```
jonathan-r-thorpe, line 943:
```suggestion
        // helper function to create Gain control instance
```

Development/nmos/node_server.cpp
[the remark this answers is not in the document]
jonathan-r-thorpe: Shall we defer resolving this until a future iteration? If yes we can resolve this here and log an issue for a shared port.

Development/third_party/is-12/README.md
lo-simon: Updated with the corrected information

Development/nmos/control_protocol_utils.cpp
```diff
@@ -0,0 +1,652 @@
...
+        web::json::value get_runtime_property_constraints(const nc_property_id& property_id, const web::json::value& runtime_property_constraints)
+        {
+            using web::json::value;
+
+            if (!runtime_property_constraints.is_null())
+            {
+                auto& runtime_prop_constraints = runtime_property_constraints.as_array();
+                auto found_constraints = std::find_if(runtime_prop_constraints.begin(), runtime_prop_constraints.end(), [&property_id](const web::json::value& constraints)
+                    {
+                        //return nmos::fields::nc::id(property) == nmos::fields::nc::property_id(constraints);
```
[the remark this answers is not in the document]
lo-simon: done

Development/nmos/control_protocol_utils.h
lo-simon: It is matching to other function's comments too

Development/nmos-cpp-node/node_implementation.cpp
lo-simon: done

Development/nmos/control_protocol_state.cpp
lo-simon: done

Development/nmos/control_protocol_state.cpp
```diff
@@ -0,0 +1,361 @@
...
+                using web::json::value;
+
+                value parameters = value::array();
+                for (const auto& parameter : parameters_) { web::json::push_back(parameters, parameter); }
+
+                return nmos::details::make_nc_method_descriptor(description, id, name, result_datatype, parameters, is_deprecated);
+            }
+        }
+        // create standard control class method
+        method make_control_class_method(const utility::string_t& description, const nc_method_id& id, const nc_name& name, const utility::string_t& result_datatype, const std::vector<web::json::value>& parameters, bool is_deprecated, standard_method_handler method_handler)
```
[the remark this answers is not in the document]
lo-simon: done

Development/nmos/control_protocol_state.cpp
lo-simon: done

Development/nmos/control_protocol_state.cpp
lo-simon: done

Development/nmos/control_protocol_state.cpp
lo-simon: done

Development/nmos/control_protocol_state.h
lo-simon: done

Development/nmos/control_protocol_state.h
lo-simon: done

Development/nmos/control_protocol_state.h
lo-simon: done

Development/nmos/control_protocol_state.cpp
lo-simon: done as no. 2

Development/nmos-cpp-node/node_implementation.cpp
lo-simon: done

Development/nmos/control_protocol_handlers.h
```diff
@@ -0,0 +1,82 @@
...
+namespace nmos
+{
+    namespace experimental
+    {
+        struct control_protocol_state;
+        struct control_class_descriptor;
+        struct datatype_descriptor;
+    }
+
+    // callback to retrieve a specific control protocol classs descriptor
```
jonathan-r-thorpe, line 22: Typo
```suggestion
    // callback to retrieve a specific control protocol class descriptor
```

Development/nmos/node_server.cpp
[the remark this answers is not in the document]
jonathan-r-thorpe: Will resolve and create issue

Development/nmos-cpp-node/node_implementation.cpp
jonathan-r-thorpe: Will leave this to implementers to decide on strategy for managing oids

Development/nmos/test/control_protocol_test.cpp
```diff
@@ -0,0 +1,1464 @@
...
+    // datatype constraints validation
+    BST_REQUIRE_NO_THROW(nmos::details::constraints_validation(value::string(U("1a")), value::null(), value::null(), with_constraints_string_constraints_validation_params));
+    BST_REQUIRE_THROW(nmos::details::constraints_validation(value::string(U("1a2")), value::null(), value::null(), with_constraints_string_constraints_validation_params), nmos::control_protocol_exception);
+    BST_REQUIRE_THROW(nmos::details::constraints_validation(value::string(U("1*")), value::null(), value::null(), with_constraints_string_constraints_validation_params), nmos::control_protocol_exception);
+    const nmos::details::datatype_constraints_validation_parameters no_constraints_string_constraints_validation_params{ no_constraints_string_datatype, nmos::make_get_control_protocol_datatype_descriptor_handler(control_protocol_state) };
+    BST_REQUIRE_NO_THROW(nmos::details::constraints_validation(value::string(U("1234567890-abcde-!\"�$%^&*()_+=")), value::null(), value::null(), no_constraints_string_constraints_validation_params));
```
garethsb, line 818: @lo-simon, @jonathan-r-thorpe, strings with the '£' character in are causing warnings:

```
illegal character encoding in string literal [-Winvalid-source-encoding]
```

The `£` character doesn't appear in 7-bit ASCII. It's U+00A3 in Unicode, so it's fine in a `wchar_t` string literal on Windows, and actually also in Windows-1252 codepage strings... but in UTF-8 it would be the two octets `C2 A3`.

Do you need to test this character specifically?

jonathan-r-thorpe: I don't think we need to specifically test this character. Let's just remove the '£/�' from the string.

PR #331, opened 2023-05-12
Implementation of IS-13 NMOS Annotation API

garethsb:
⚠️ Ink isn't dry on the [IS-13 spec](https://specs.amwa.tv/is-13), but here's an implementation.


garethsb:
Current `annotation_patch_merger` has three arguments:

1. `const nmos::resource& resource` - the IS-04 resource that is being `PATCH`ed
2. `web::json::value& value` - contains the annotation properties from that resource, modifiable by the callback
3. `const web::json::value& patch` - contains the body of the `PATCH` request, which is guaranteed valid per the schema

The behaviour is that the callback is expected to update the `value` with the `patch`, and then the calling function updates the `resource` from the updated value.

Passing read-only `resource` and a writable `value` ensures that the callback cannot update other properties than `label`, `description` and `tags`, but that it has access to the full resource to determine semantic validity.

However, the implementation is still a bit complicated. The callback might want to proceed to...

* reject the `PATCH` request completely if it attempts to change to read-only properties e.g. because the whole resource is read-only (Sources and Flows in dynamic Nodes), or because certain `tags` are read-only such as BCP-002-02 Asset Distinguishing Information or vendor-specific tags
* reject the `PATCH` request completely if the values are unacceptable for this particular resource, such as being too large for persistent storage or not meeting format for particular tag
* merge the `patch`
* apply default values if the `PATCH` request removed/reset values that cannot be omitted per spec or particular resource
* reject the `PATCH` request completely if the final values are unacceptable

garethsb:
As @jonathan-r-thorpe suggested, `nmos::details::is_read_only_tag` is now an overridable predicate, and default values can also be specified declaratively. This is demonstrated by _nmos-cpp-node/node_implementation.cpp_ for the Node and Device tags specified via _config.json_ `node_tags` and `device_tags`.

Development/nmos/annotation_api.h
```diff
@@ -0,0 +1,53 @@
...
+
+#include "cpprest/api_router.h"
+
+namespace slog
+{
+    class base_gate;
+}
+
+// Annotation API implementation
+// See https://specs.amwa.tv/is-13/branches/v1.0-dev/APIs/ReadWriteNodeAPI.html
```
garethsb, line 12:
```suggestion
// See https://specs.amwa.tv/is-13/branches/v1.0-dev/APIs/AnnotationAPI.html
```

garethsb:
@jonathan-r-thorpe, @lo-simon, I've just attempted to merge **master** into this **rwnode** branch, so let's keep an eye on the tests :-)

garethsb:
Rebased onto current `master` as a single commit and force-pushed `garethsb:rwnode`.

garethsb:
This implementation has been run against the revised IS-13-01 suite in AMWA-TV/nmos-testing#914 (all `test_00`–`test_18` passed, with Node/Device BCP-002-02 asset tags and Sender/Receiver BCP-002-01 group hint present).

nmos-cpp CI clones AMWA-TV/nmos-testing `master` and runs a fixed list of suites, so IS-13-01 cannot be added there until #914 is merged. We'll wire it into CI at that point.

PR #336, opened 2023-06-13
Strip leading zeroes from SDP

maweit:
There are IS-05 compatible devices out in the field which supply SDP files where e.g. the session-id comes  with
leading zeroes (looks like: o=- 00098979940036). This patch strips our leading zeroes from session-id and session-version.

### DIFF
--- a/Development/sdp/sdp_grammar.cpp
+++ b/Development/sdp/sdp_grammar.cpp
@@ -44,6 +44,14 @@ namespace sdp
             return v.as_number(), web::json::visit(web::json::basic_ostream_visitor<char>(os), v), os.str();
         }
         inline web::json::value s2jn(const std::string& s) { auto v = web::json::value::parse(utility::s2us(s)); return v.as_number(), v; }
+        inline web::json::value
+        s2jn_no_leading_zero(const std::string& s)
+        {
+            // strip leading zeroes as they are not valid json and parse will throw,
+            // but they are allowed for sdp fields like session-id
+            auto str = regex_replace(s, std::regex("^(-?)(0+)(0|[1-9])", std::regex::extended), "$1$3");
+            auto v = web::json::value::parse(utility::s2us(str)); return v.as_number(), v;
+        }
 
         // find the first delimiter in str, beginning at pos, and return the substring from pos to the delimiter (or end)
         // set pos to the end of the delimiter
@@ -70,6 +78,7 @@ namespace sdp
         const converter string_converter{ js2s, s2js };
 
         const converter number_converter{ jn2s, s2jn };
+        const converter js_number_converter{ jn2s, s2jn_no_leading_zero };
 
         // <key>[<separator><value>]
         converter key_value_converter(char separator, const std::pair<utility::string_t, converter>& key_converter, const std::pair<utility::string_t, converter>& value_converter)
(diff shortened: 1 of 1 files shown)


garethsb:
Just paraphrasing comments from off-line thread:

If we look at the RFC 4566 ABNF grammar, there are several instances of `1*DIGIT`, which therefore could have leading zeros, and there are a few instances of e.g. `POS-DIGIT *DIGIT` which don’t allow leading zeros.
 
`1*DIGIT` is used for
*	`v=` `<proto-version>` (though luckily that should always be zero)
*	`o=` `<sess-id>` (the example that motivated this PR, called a “numeric string” in the spec)
*	`o=` `<sess-version>`
*	`b=` `<bandwidth>` (interesting, why on earth does that need to allow leading zeros!!)
*	`r=`/`z=` `<typed-time>` (also no good reason it allows leading zeros, as far as I can see; also unused in SDP files that we’re interested in)
*	m= <port> (again, no good reason I can see that this allows leading zeros!!)
 
For all of these except sess-id and sess-version, it doesn’t seem important to me to maintain knowledge of the leading zeros, even though we couldn’t roundtrip the exact SDP text format from our JSON representation.
 
But since sess-id is a “numeric string” would it matter if we couldn’t output leading zeros? I.e. if we just fix the parsing to accept/ignore leading zeros, if you parsed and regenerated the example SDP file, the leading zeros will be gone.
The only way to get roundtrip is to keep the values that allow leading zeros as strings, I think…
But that changes the JSON data model.
...
If we accept the inability to roundtrip. I don't think it's quite as simple as stripping leading zeros, because the fix should probably be applied in number_converter, which must also be able to handle floating point, and simple integer '0',  so we need to avoid stripping a final zero before end of string or before decimal point, maybe? Test cases needed!

garethsb:
To be clear, at the moment, leading zeros are rejected with an `sdp_exception`, so we definitely need to do _something_!

Development/sdp/sdp_grammar.cpp
```diff
@@ -44,6 +44,14 @@ namespace sdp
             return v.as_number(), web::json::visit(web::json::basic_ostream_visitor<char>(os), v), os.str();
         }
         inline web::json::value s2jn(const std::string& s) { auto v = web::json::value::parse(utility::s2us(s)); return v.as_number(), v; }
+        inline web::json::value
+        s2jn_no_leading_zero(const std::string& s)
+        {
+            // strip leading zeroes as they are not valid json and parse will throw,
+            // but they are allowed for sdp fields like session-id
+            auto str = regex_replace(s, std::regex("^(-?)(0+)(0|[1-9])", std::regex::extended), "$1$3");
```
garethsb, line 52: Is the extended POSIX grammar required for this? I would have thought the default (modified ECMAScript) grammar would be sufficient?

garethsb:
https://grammarhow.com/zeroes-or-zeros/ 😄

garethsb:
In terms of implementation, I think having a separate converter for `1*DIGIT` than for `POS-DIGIT *DIGIT` as @maweit proposes makes sense.

I think `js_number_converter` isn't the right name for the former. `number_with_leading_zeros_converter` is too long-winded... how about `digits_converter`?

Then we need to decide whether to also use it for the other instances of `1*DIGIT` that I identified, e.g. `<proto-version>`, `<bandwidth>`, `<port>` and so on...

lo-simon:
Fixed in #340

PR #337, opened 2023-06-19
Add "service_name_prefix" to settings

N-Nagorny:
Hi @lo-simon, @jonathan-r-thorpe.

I like the default DNS-SD service name, it saves time for record resolving but the static `nmos-cpp` prefix make me a bit upset. Could you please review this PR making this prefix customizable?

### DIFF
--- a/Development/nmos/settings.h
+++ b/Development/nmos/settings.h
@@ -114,7 +114,7 @@ namespace nmos
         const web::json::field_as_integer_or discovery_backoff_max{ U("discovery_backoff_max"), 30 };
         const web::json::field_with_default<double> discovery_backoff_factor{ U("discovery_backoff_factor"), 1.5 };
 
-        // service_name_prefix [registry, node]: used as the name prefix of the advertised services (e.g. "nmos-cpp_node" or "nmos-cpp_query")
+        // service_name_prefix [registry, node]: used as a prefix in the advertised service names ("<prefix>_<api>_<host>:<port>", e.g. "nmos-cpp_node_127-0-0-1:3212")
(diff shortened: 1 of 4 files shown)


Development/nmos/settings.h
```diff
@@ -114,6 +114,9 @@ namespace nmos
         const web::json::field_as_integer_or discovery_backoff_max{ U("discovery_backoff_max"), 30 };
         const web::json::field_with_default<double> discovery_backoff_factor{ U("discovery_backoff_factor"), 1.5 };
 
+        // service_name_prefix [registry, node]: used as the name prefix of the advertised services (e.g. "nmos-cpp_node" or "nmos-cpp_query")
```
garethsb, line 117: The examples "nmos-cpp_node" and "nmos-cpp_query" are neither examples of sensible prefixes nor examples of the full service name that would be created, but examples of the `service_base_name` which is an implementation detail so could be a bit confusing for someone who doesn't look closely at the code?

Not sure if we want to show the full service name? If so, maybe:

```suggestion
        // service_name_prefix [registry, node]: used as a prefix in the advertised service names ("<prefix>_<api>_<host>:<port>", e.g. "nmos-cpp_node_nmos-api-local:3212")
```

garethsb (commented):
Basically looks like a good idea to me

Development/nmos/settings.h
N-Nagorny: Thank you, you're right. Fixed it at https://github.com/sony/nmos-cpp/pull/337/commits/89908c5ddfeca3fcbe6791e4bddbdeeaec8f31b0

lo-simon (approved):
LGTM

PR #343, opened 2023-09-07
Fix downgrade support

lo-simon:
Fix IS-04 (v1.3) downgrade resource to register to v1.2 Registry.

There are some parameters which have been extended in nmos-parameter-registers for the latest nmos version.

The problem was identified when a downgraded v1.3 video flow with `transfer_characteristic` set to `UNSPECIFIED` failed to register to the v1.2 Registry. It is because v1.2 video flow schema used in the Registry does not allow `transfer_characteristic` to be `UNSPECIFIED`. This PR sets the default values for the downgraded resources when an invalid value is used or removes it if the parameter is optional.

See https://specs.amwa.tv/nmos-parameter-registers/branches/main/flow-attributes/#transfer-characteristic

### DIFF
--- a/Development/nmos/api_downgrade.cpp
+++ b/Development/nmos/api_downgrade.cpp
@@ -1033,12 +1033,12 @@ namespace nmos
     struct property
     {
         utility::string_t name;
-        web::json::value default;
+        web::json::value default_value;
         bool required;  // is required property
 
-        property(utility::string_t name, web::json::value default = web::json::value::null(), bool required = true)
+        property(utility::string_t name, web::json::value default_value = web::json::value::null(), bool required = true)
             : name(std::move(name))
-            , default(std::move(default))
+            , default_value(std::move(default_value))
             , required(std::move(required))
         {}
     };
@@ -1292,7 +1292,7 @@ namespace nmos
                             // if property required, i.e. obligatory
                             if (property.required)
                             {
-                                result[property.name] = property.default;
+                                result[property.name] = property.default_value;
                             }
                             else
                             {
--- a/Development/nmos/api_downgrade.cpp
+++ b/Development/nmos/api_downgrade.cpp
@@ -1,11 +1,973 @@
 #include "nmos/api_downgrade.h"
 
 #include <map>
+#include "cpprest/json_validator.h"
+#include "cpprest/basic_utils.h"
 #include "nmos/is04_versions.h"
 #include "nmos/resource.h"
+#include "nmos/type.h"
 
 namespace nmos
 {
+    namespace details
+    {
+        static web::json::value make_schema(const char* schema)
+        {
+            return web::json::value::parse(utility::s2us(schema));
+        }
+
+        static web::uri make_schema_id(const api_version& version, const nmos::type& type, const utility::string_t& property)
+        {
+            return U("/downgrade") + make_api_version(version) + U("-") + type.name + U("-") + property;
+        }
+
+        // property schemas are extracted from the resource schemas
+        const std::map<web::uri, web::json::value> property_schemas =
+        {
+            // node
+            // services
+            { make_schema_id(is04_versions::v1_0, nmos::types::node, U("services")), make_schema(R"({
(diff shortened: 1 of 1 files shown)


garethsb:
Since this is a significant change to code which is used by both Node and Registry, I recommend unit tests with very good coverage are added.

Development/nmos/api_downgrade.cpp
```diff
@@ -1,11 +1,973 @@
...
+                          "type": "string",
+                          "description": "URL to reach a service running on the Node",
+                          "format": "uri"
+                        },
+                        "type": {
+                          "type": "string",
+                          "description": "URN identifying the type of service",
+                          "format": "uri"
+                        },
+                        "authorization": {
```
garethsb, line 77: Is it necessary to remove `authorization` when downgrading before v1.3? The v1.2 schema doesn't consider it invalid as far as I know. Same applies to many other cases.

Development/nmos/api_downgrade.cpp
```diff
@@ -68,69 +1030,219 @@ namespace nmos
...
-                    { nmos::is04_versions::v1_1, { U("description"), U("tags"), U("api"), U("clocks") } },
-                    { nmos::is04_versions::v1_2, { U("interfaces") } }
+                    { nmos::is04_versions::v1_0, {
+                        { U("id") },
+                        { U("version") },
+                        { U("label") },
+                        { U("href") },
+                        { U("hostname") },
+                        { U("caps") },
+                        { U("services"), value::array() }
```
garethsb, line 1063: This doesn't seem like right strategy to me. I don't think we should remove `authorization` when considering v1.3 node resource and exposing it via Node API v1.0 or registering via Registration API v1.0. If someone has good reason that we should do that, I don't think we should ever remove all services and replace with empty array... 😕

Development/nmos/api_downgrade.cpp
```diff
@@ -1,11 +1,973 @@
...
+            if (property_schemas.end() == found)
+            {
+                throw web::json::json_exception((_XPLATSTR("schema not found for ") + id.to_string()).c_str());
+            }
+
+            return found->second;
+        }
+
+        // list of property validators to identify the associated property has been changed since the previous version
+        static const web::json::experimental::json_validator& property_changed_validator()
```
garethsb, line 918: Minor nit: `property_changed_validator` sounds like something that is used to perform validation every time a property is changed (e.g. as a result of `nmos::modify_resource`). Not sure what is the best name, though!

Development/nmos/api_downgrade.cpp
```diff
@@ -1,11 +1,973 @@
...
+        {
+            return web::json::value::parse(utility::s2us(schema));
+        }
+
+        static web::uri make_schema_id(const api_version& version, const nmos::type& type, const utility::string_t& property)
+        {
+            return U("/downgrade") + make_api_version(version) + U("-") + type.name + U("-") + property;
+        }
+
+        // property schemas are extracted from the resource schemas
```
garethsb, line 24: Since all of these property schemas are already included in the full schemas embedded in nmos-cpp, could this code be refactored to use a json value path to extract the property value schema from those? Not sure...

garethsb (commented):
A few comments

Development/nmos/api_downgrade.cpp
lo-simon: If we agree with this PR, this can be followed up.

Development/nmos/api_downgrade.cpp
lo-simon: maybe just named it as `property_validator`, using the word changed because I wanted to specify that the validator is only used for those properties which have schema changed between versions.

Development/nmos/api_downgrade.cpp
lo-simon: as long as the JSON validation is successful, the `authorization` parameter will not be removed.

lo-simon:
> Since this is a significant change to code which is used by both Node and Registry, I recommend unit tests with very good coverage are added.

Yes totally agree

PR #347, opened 2023-10-04
Rollback make-badges to include Ubuntu 14.04

lo-simon:


lo-simon:
Redo rollback make-badges to include Ubuntu 14.04 in a new PR.

PR #350, opened 2023-10-23
Open getaddrinfo for public use

lo-simon:
Open the `getaddrinfo` for public use. It is a beneficial function which can look up IP address from the hostname.

### DIFF
--- a/Development/cmake/NmosCppTest.cmake
+++ b/Development/cmake/NmosCppTest.cmake
@@ -134,11 +134,6 @@ target_include_directories(nmos-cpp-test PRIVATE
     ${CMAKE_CURRENT_SOURCE_DIR}
     ${CMAKE_CURRENT_SOURCE_DIR}/third_party
     )
-if(${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
-    target_include_directories(nmos-cpp-test PRIVATE
-        ${CMAKE_CURRENT_SOURCE_DIR}/third_party/mDNSResponder/mDNSShared
-		)
-endif()
 
 include(Catch)
 
--- a/Development/cmake/NmosCppTest.cmake
+++ b/Development/cmake/NmosCppTest.cmake
@@ -134,6 +134,11 @@ target_include_directories(nmos-cpp-test PRIVATE
     ${CMAKE_CURRENT_SOURCE_DIR}
     ${CMAKE_CURRENT_SOURCE_DIR}/third_party
     )
+if(${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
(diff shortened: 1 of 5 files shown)


lo-simon:
Thanks, @garethsb, we want it to be opened, so it is not for test code. I was thinking the same, but not sure whether to bring it into `service_discovery` or some other class. Okay let's do it in `service_discovery`, will do an update shortly.

garethsb:
@lo-simon depending on use case, would `host_addresses` be sufficient?

https://github.com/sony/nmos-cpp/blob/f7bd2f7317358123cb0685ef121ed3606c632952/Development/cpprest/host_utils.cpp#L313

lo-simon:
> @lo-simon depending on use case, would `host_addresses` be sufficient?
> 
> https://github.com/sony/nmos-cpp/blob/f7bd2f7317358123cb0685ef121ed3606c632952/Development/cpprest/host_utils.cpp#L313

Using the `host_addresses` will not be making the hostname resolution via the DNS server specified in resolv.conf configured on the device. Therefore opening up the `getaddrinfo` is necessary.

garethsb:
Hmm, that is surprising,  I thought it should use the system resolver that would use _resolv.conf_ if configured correctly.

garethsb:
As discussed earlier today, it would be good to have a clear, demonstrated, use case, where using DNSServiceGetAddrInfo produces the required result but POSIX `getaddrinfo` (which does use _nnswitch.conf_ and _resolv.conf_ on Linux) via Boost.Asio via our `host_addresses` function doesn't.

If that is the case, this revised PR looks nice. Just a couple of questions:
* can the DNSServiceGetAddrInfo callback be called multiple times to return more than one address, i.e. should the result be vector-of-`address_result` like the results of resolve and browse are vectors of their respective types?
* is it OK to require internet access in the test code (resolving Google DNS to 8.8.8.8)?

lo-simon:
> can the DNSServiceGetAddrInfo callback be called multiple times to return more than one address, i.e. should the result be vector-of-address_result like the results of resolve and browse are vectors of their respective types?

The function wrapped the `mdns_details::getaddrinfo`, which executes the `DNSServiceRefDeallocate` after receiving the 1st `addrinfo` to prevent any more DNSServiceGetAddrInfo callback. But the wrapped function should also be returning a vector of results as in case no mDNSResponder is used, `host_addresses` will be used which will be returning a list of ipaddresses.

> is it OK to require internet access in the test code (resolving Google DNS to 8.8.8.8)?

As it is already passing the `testDnsGetAddrInfo` unit test, the runner must be able to access the internet.

PR #351, opened 2023-10-24
Fix SSN for ST2110-40:2023 to be ST2110-40:2021

garethsb:
Although this revision should have been published in 2022 and the SSN should have been updated to match, neither of those things actually happened...

https://ieeexplore.ieee.org/document/10064369


jonathan-r-thorpe (approved):
LGTM

PR #366, opened 2024-01-31
Bump dependencies

lo-simon:
Bump `boost` to 1.83.0 and `cpprestsdk` to 2.10.19

### DIFF
--- a/Development/nmos/settings.h
+++ b/Development/nmos/settings.h
@@ -107,7 +107,7 @@ namespace nmos
         // pri [registry, node]: used for the 'pri' TXT record; specifying nmos::service_priorities::no_priority (maximum value) disables advertisement completely
         const web::json::field_as_integer_or pri{ U("pri"), 100 }; // default to highest_development_priority
 
-        // highest_pri, lowest_pri [registry, node]: used to specify the (inclusive) range of suitable 'pri' values of discovered APIs, to avoid development and live systems colliding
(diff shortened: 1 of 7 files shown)


Development/nmos/settings.h
```diff
@@ -107,7 +107,7 @@ namespace nmos
         // pri [registry, node]: used for the 'pri' TXT record; specifying nmos::service_priorities::no_priority (maximum value) disables advertisement completely
         const web::json::field_as_integer_or pri{ U("pri"), 100 }; // default to highest_development_priority
 
-        // highest_pri, lowest_pri [registry, node]: used to specify the (inclusive) range of suitable 'pri' values of discovered APIs, to avoid development and live systems colliding
+        // highest_pri, lowest_pri [node]: used to specify the (inclusive) range of suitable 'pri' values of discovered APIs, to avoid development and live systems colliding
```
jonathan-r-thorpe, line 110: Was this an existing error in the comment that's been corrected?

lo-simon: Yes, the comment in it is corrected.

When IS-10 was first instructed, the `higher_pri` and `lowest_pri` were also used for Authorization API priority, but while under development, very often the Registration API can be using a different pri than the Authorization API, as the Authorization API does not provide by the Registry. So therefore in settings, there are a new `authorization_highest_pri` and `authorization_lowest_pri` for the Authorization API priority.

jonathan-r-thorpe (approved):
LGTM

PR #368, opened 2024-02-01
Fix advertiser service

lo-simon:
Thank you to `Niitsu Keita-san` from Sony, who has pointed out that the newly created client (the _DNSServiceRef_t*) will not be reflected to the outside function. In this case, the client will not be released when the advertiser is destructed, i.e., causing a memory leak.


jonathan-r-thorpe (approved):
LGTM

PR #370, opened 2024-02-14
Markdown and whitespace tweaks

garethsb:
* Fix broken link

* Document dependency on jwt-cpp

* Simplify jwt-cpp/README.md to just point at upstream, like other _third_party_ subdirectories

* fix tabs to spaces

* fix more tabs to spaces


lo-simon (approved):
LGTM

PR #373, opened 2024-02-16
Conan 2

garethsb:
* Conan 1.X is no longer supported
* `NMOS_CPP_USE_CONAN` cache variable is no more, but there are two new `NMOS_CPP_USE_SUPPLIED_xxx` options to use the supplied json-schema-validator and jwt-cpp if you're not using Conan
* To use Conan as the dependency provider, `CMAKE_PROJECT_TOP_LEVEL_INCLUDES` must be used explicitly, which is documented in the Getting Started guide (see [Dependency Providers](https://cmake.org/cmake/help/v3.24/command/cmake_language.html#dependency-providers) for the rationale)
  * The version from the tip of the develop2 branch ([f6464d1/conan_provider.cmake](https://github.com/conan-io/cmake-conan/blob/f6464d1e13ef7a47c569f5061f9607ea63339d39/conan_provider.cmake)) is added to the repo
* A conan lockfile is created and output in the log in CI, which will help us track which precise versions have been used for each build
* Bump CMake to 3.28.3, apart from on ubuntu-14.04 (where it didn't work)


garethsb:
Noticed while I was updating CI and documentation that at the moment, the Dependencies doc says:

> Install a [DNS Service Discovery](https://github.com/sony/nmos-cpp/blob/master/Documents/Dependencies.md#dns-service-discovery) implementation, since this isn't currently handled by Conan

That isn't true - hasn't been true for a long time now. However, we haven't changed this because the required library is different on Windows (mDNSResponder) than on Linux (Avahi or mDNSResponder), and that can't be expressed in a simple _conanfile.txt_. It would be good to improve this, as it would simplify the build even further. We could adopt the CCI _conanfile.py_ which expresses the necessary logic.

README.md
lo-simon, line 1: > An [nmos-cpp Conan package](https://conan.io/center/nmos-cpp) is published at Conan Center Index.

I think the link for the [nmos-cpp Conan package] should be https://conan.io/center/recipes/nmos-cpp instead of https://conan.io/center/nmos-cpp

lo-simon (approved):
LGTM

garethsb:
Not sure whether we want to note anywhere what a user will experience if they are not ready for the migration, which requires:
* Conan 2.0.5 or higher (not Conan 1.X)
* CMake 3.24 or higher
* new `cmake` command

If they try to use previous `cmake` command without `-DCMAKE_PROJECT_TOP_LEVEL_INCLUDES`, or if they do not have CMake 3.24 or higher (e.g. CMake 3.20 supplied with Visual Studio 2019) so the `CMAKE_PROJECT_TOP_LEVEL_INCLUDES` will just be ignored, they'll get this error from `find_package(Boost)`.

```
cmake .. -DCMAKE_CONFIGURATION_TYPES="Debug;Release"
-- Building for: Visual Studio 16 2019
...
CMake Error at C:/Program Files/CMake/share/cmake-3.28/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find Boost (missing: Boost_INCLUDE_DIR system date_time regex
  thread) (Required is at least version "1.54.0")
Call Stack (most recent call first):
  C:/Program Files/CMake/share/cmake-3.28/Modules/FindPackageHandleStandardArgs.cmake:600 (_FPHSA_FAILURE_MESSAGE)
  C:/Program Files/CMake/share/cmake-3.28/Modules/FindBoost.cmake:2393 (find_package_handle_standard_args)
  cmake/NmosCppDependencies.cmake:19 (find_package)
  CMakeLists.txt:21 (include)


-- Configuring incomplete, errors occurred!
```

If they have conan 1.X but do use the new `cmake` command, they'll get an error message telling them they need Conan 2.0.5 or higher.

```
cmake .. -DCMAKE_CONFIGURATION_TYPES="Debug;Release" -DCMAKE_PROJECT_TOP_LEVEL_INCLUDES="third_party/cmake/conan_provider.cmake"
-- Selecting Windows SDK version 10.0.19041.0 to target Windows 10.0.19045.
WARN: Migration: Updating settings.yml
CMake Error at third_party/cmake/conan_provider.cmake:487 (message):
  CMake-Conan: Conan version must be 2.0.5 or later
Call Stack (most recent call first):
  third_party/cmake/conan_provider.cmake:516 (conan_version_check)
  cmake/NmosCppDependencies.cmake:19 (find_package)
  CMakeLists.txt:21 (include)


-- Configuring incomplete, errors occurred!
```

lo-simon:
How about under the [Recent Activity](https://github.com/sony/nmos-cpp#recent-activity), we extend and bold the wording `Update to Conan 2; Conan 1.X is no longer supported` with a MUST and otherwise you will get error.

garethsb:
I've started investigating how to use Conan 2 with earlier versions of CMake, by manually running `conan install` and passing a generated preset/toolchain to CMake.

Very approximately (need to sort out paths because this is doing an in-source build)...

```sh
conan install .. -g CMakeToolchain -s:a build_type=Release --build=missing
conan install .. -g CMakeToolchain -s:a build_type=Debug --build=missing
cmake .. --preset conan-default -DCMAKE_CONFIGURATION_TYPES="Debug;Release"
```

PR #374, opened 2024-02-19
Discovery mode revisited

### LINKED ISSUE #357: Provide an HTTP Host header from DNS-SD name resolution for Registration API and System API discovery
nmos-cpp uses a DNS-SD library (Bonjour/mDNSResponder or Avahi) to discover the hostname of the Registry.

By default it also uses the DNS-SD library to then resolve the hostname to an IP address, and makes HTTP requests using that IP address.
This ensured success with mDNS .local names on Windows, and on Linux systems where the Name Service Switch was misconfigured.
However, it means that the HTTP requests to the Registry do not get a Host header. That causes trouble for Gateways/Proxies.

Since June 2022 (https://github.com/sony/nmos-cpp/pull/263) nmos-cpp supports a discovery mode where it delegates name resolution to the HTTP library (which will rely on the system resolver).
Using this would fix the problem... if you didn't care about the cases I mentioned above...

Probably ideally, nmos-cpp would implement name resolution using DNS-SD, but ensure that a `Host` header is included even so. That'll mean passing both the hostname and the IP Address back from `nmos::details::get_resolved_hosts` and ultimately out of `nmos::resolve_service` to be used by the node behaviour thread and node system behaviour thread... All assuming that the underlying HTTP library will allow us to specify a URL with an IP address but include a `Host` header!

The DNS-SD resolution mechanism falls back to `getaddrinfo`, so using this approach should work in all cases and we could apply it also in HTTPS mode, and therefore remove the `discovery_mode` setting again.

garethsb:
POC of replacing `discovery_mode` setting with sneakily stashing `Host` header in the `web::uri` values returned by `nmos::details::resolve_service` and using this in a `web::http::http_pipeline_stage`.

Resolves #357

Using the user info in this way is a bit icky. But the other option is to make lots of places that pass `web::uri` around pass a pair of the URI with the `Host` header. This way has the side effect of making the values in the settings look something like this whether doing HTTP or HTTPS:

```
"registration_services": [
  "http://reg0.example.com@192.0.2.0:80/x-nmos/registration/v1.3",
  "http://reg0.example.com@198.51.100.0:80/x-nmos/registration/v1.3",
  "http://reg0.example.com@203.0.113.0:80/x-nmos/registration/v1.3",
  "http://reg1.example.com@192.0.2.1:80/x-nmos/registration/v1.3",
  "http://reg1.example.com@198.51.100.1:80/x-nmos/registration/v1.3",
  "http://reg1.example.com@203.0.113.1:80/x-nmos/registration/v1.3",
  ...
]
```

If anyone is relying on that property e.g. to populate UI, that would be an unexpected change. :-/

(However, it does also allow the existing `registry_address` setting to be populated in the same way, and that just works too.)

### DIFF
--- a/Development/nmos-cpp-node/config.json
+++ b/Development/nmos-cpp-node/config.json
@@ -235,9 +235,6 @@
     // proxy_port [registry, node]: forward proxy port
     //"proxy_port": 8080,
 
-    // discovery_mode [node]: whether the discovered host name (1) or resolved addresses (2) are used to construct request URLs for Registration APIs or System APIs
-    //"discovery_mode": 1,
-
     // href_mode [registry, node]: whether the host name (1), addresses (2) or both (3) are used to construct response headers, and host and URL fields in the data model
     //"href_mode": 1,
 
--- a/Development/nmos/authorization_operation.cpp
+++ b/Development/nmos/authorization_operation.cpp
@@ -1200,7 +1200,7 @@ namespace nmos
                     const auto service = top_authorization_service(model.settings);
 
                     const auto auth_uri = service.second;
-                    client.reset(new web::http::client::http_client(auth_uri, make_authorization_http_client_config(model.settings, load_ca_certificates, gate)));
+                    client = nmos::details::make_http_client(auth_uri, make_authorization_http_client_config(model.settings, load_ca_certificates, gate));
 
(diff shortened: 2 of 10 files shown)


lo-simon (approved):
LGTM

PR #376, opened 2024-03-01
Fix openssl3 deprecated functions

lo-simon:


Development/nmos/jwk_utils.cpp
```diff
@@ -107,65 +120,45 @@ namespace nmos
...
-                auto x = utility::conversions::from_base64url(base64_x);
-                auto y = utility::conversions::from_base64url(base64_y);
-
-                BIGNUM_ptr x_coordinate(BN_bin2bn(x.data(), (int)x.size(), NULL), &BN_free);
-                BIGNUM_ptr y_coordinate(BN_bin2bn(y.data(), (int)y.size(), NULL), &BN_free);
+                OSSL_PARAM_ptr params(OSSL_PARAM_BLD_to_param(param_bld.get()), &OSSL_PARAM_free);
+                EVP_PKEY_CTX_ptr ctx(EVP_PKEY_CTX_new_from_name(NULL, "RSA", NULL), &EVP_PKEY_CTX_free);
 
-                if (EC_KEY_set_public_key_affine_coordinates(ec_key.get(), x_coordinate.get(), y_coordinate.get()))
+                EVP_PKEY* pkey = NULL;
```
garethsb, line 147: Would it be possible to use a `EVP_PKEY_ptr` here and avoid the explicit `EVP_PKEY_free` calls below?

Development/nmos/jwk_utils.cpp
```diff
@@ -174,101 +167,48 @@ namespace nmos
...
-                    // using crv, x & y to convert Json Web Key to EC Public Key
-                    return jwk_to_public_key(jwk.at(U("crv")).as_string(), jwk.at(U("x")).as_string(), jwk.at(U("y")).as_string()); // may throw jwk_exception
-                }
-                else
-                {
-                    throw jwk_exception("jwk contains invalid 'kty': " + utility::us2s(jwk.serialize()));
                 }
-            }
+                return utility::string_t{};
+            };
```
garethsb, line 189: `extra ‘;’ [-Wpedantic]`

```suggestion
            }
```

Development/nmos/jwk_utils.cpp
```diff
@@ -277,111 +217,186 @@ namespace nmos
                     { U("use"), pubkey_use.name }
                 });
             }
```
garethsb, line 219: This close-brace needs to be outside the #if-#else-#endif

```suggestion
```

Development/nmos/jwk_utils.cpp
```diff
@@ -277,111 +217,186 @@ namespace nmos
...
+                    { U("kid"), keyid },
+                    { U("kty"), U("RSA") },
+                    { U("n"), base64_n },
+                    { U("e"), base64_e },
+                    { U("alg"), alg.name },
+                    { U("use"), pubkey_use.name }
+                });
             }
+#endif
+        }
```
garethsb, line 249:
```suggestion
#endif
            }
        }
```

garethsb: Actually, would it be OK... nice... to move the `construct jwk` return-statement outside the #if-#else-#endif as well?

Development/nmos/jwk_utils.cpp
```diff
@@ -277,111 +217,186 @@ namespace nmos
...
-                }
-                return{};
+            if (!private_key)
+            {
+                throw jwk_exception("extract public key error: failed to create EVP_PKEY-RSA from BIO private key");
             }
 
-            // find the RSA private key from private key list
-            utility::string_t found_rsa_key(const std::vector<utility::string_t>& private_keys)
+            RSA_ptr rsa(EVP_PKEY_get1_RSA(private_key.get()), &RSA_free);
```
garethsb, line 271: In this function is there no difference in control flow, only the difference between these two snippets?

OpenSSL <3.0
```C++
            RSA_ptr rsa(EVP_PKEY_get1_RSA(private_key.get()), &RSA_free);
            if (!rsa)
            {
                throw jwk_exception("extract public key error: failed to load RSA key from private key");
            }
            if (bio && PEM_write_bio_RSA_PUBKEY(bio.get(), rsa.get()))
```
OpenSSL >=3.0
```C++
            if (bio && PEM_write_bio_PUBKEY(bio.get(), private_key.get()))
```

Development/boost/asio/ssl/use_tmp_ecdh.hpp
```diff
@@ -40,16 +45,19 @@ struct evp_pkey_cleanup
     ~evp_pkey_cleanup() { if (p) ::EVP_PKEY_free(p); }
 };
 
+#if OPENSSL_VERSION_NUMBER < 0x30000000L
 struct ec_key_cleanup
```
garethsb, line 49: @lo-simon I can't remember, was this cleanup-struct idiom adopted from the Boost.ASIO code?
It serves the same purpose as your usage of `std::unique_ptr` in _nmos/jwk_utils.cpp_, is that right?

garethsb:
@lo-simon Is this tested in CI with both OpenSSL 3.0+ and with OpenSSL 1.1.1?

Development/nmos/jwk_utils.cpp
lo-simon: Yes I had the same thought, but couldn't think how it could be done

Development/nmos/jwk_utils.cpp
lo-simon: yes, sure

Development/nmos/jwk_utils.cpp
lo-simon: Yes, sure, but I thought we wanted the function to be #if around the whole function body rather than just in part of the function.

Development/nmos/jwk_utils.cpp
lo-simon: I am happy to change it.

Development/nmos/jwk_utils.cpp
lo-simon: I am happy to change it.

Development/boost/asio/ssl/use_tmp_ecdh.hpp
lo-simon: yes, it is; see the following example extracted from asio\ssl\impl\contentext.ipp

```
struct context::bio_cleanup
{
  BIO* p;
  ~bio_cleanup() { if (p) ::BIO_free(p); }
};

struct context::x509_cleanup
{
  X509* p;
  ~x509_cleanup() { if (p) ::X509_free(p); }
};

struct context::evp_pkey_cleanup
{
  EVP_PKEY* p;
  ~evp_pkey_cleanup() { if (p) ::EVP_PKEY_free(p); }
};

#if (OPENSSL_VERSION_NUMBER < 0x30000000L)
struct context::rsa_cleanup
{
  RSA* p;
  ~rsa_cleanup() { if (p) ::RSA_free(p); }
};

struct context::dh_cleanup
{
  DH* p;
  ~dh_cleanup() { if (p) ::DH_free(p); }
};
#endif // (OPENSSL_VERSION_NUMBER < 0x30000000L)
```

lo-simon:
> @lo-simon Is this tested in CI with both OpenSSL 3.0+ and with OpenSSL 1.1.1?

We kept Ubuntu14.04 to use OpenSSL 1.0.1f, so CI will test < 3.0

Development/nmos/jwk_utils.cpp
lo-simon: Done using the boost.ASIO cleanup-struct idiom

Development/nmos/jwk_utils.cpp
lo-simon: Fixed

garethsb (approved):

PR #386, opened 2024-04-09
Add Installation Guide for nmos-cpp Using Conan

johanjino:
Documentation to install nmos-cpp using Conan package manager. The guide includes instructions for installing Conan, detecting profiles, installing nmos-cpp, and running the applications on both Windows and Linux platforms.

### DIFF
--- a/Documents/Installation-with-Conan.md
+++ b/Documents/Installation-with-Conan.md
@@ -1,0 +1,72 @@
+## Install nmos-cpp Using Conan
+
+The following steps describe how to install and set up nmos-cpp using the Conan package manager.
+For many platforms, a binary package is available from Conan Center Index so it isn't necessary to build nmos-cpp or any of its dependencies.
+
+1. Install conan:
+   ```sh
+   pip install --upgrade conan~=2.0.5
+   ```
+
+   `pip` is the package installer for Python. Install Python 3 if necessary.
+
+   If the python Scripts directory is not on the PATH you will get a warning like:
+   > WARNING: The script conan.exe is installed in 'C:\Users\\%USERNAME%\\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\local-packages\Python312\Scripts' which is not on PATH.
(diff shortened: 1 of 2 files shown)


jonathan-r-thorpe:
Hi @johanjino - this looks good to me - well done :) I'll put any specific comments in line with the documents.

Documents/Installation-with-Conan.md
```diff
@@ -0,0 +1,72 @@
...
+   pip install --upgrade conan~=2.0.5
+   ```
+
+   `pip` is the package installer for Python. Install Python 3 if necessary.
+
+   If the python Scripts directory is not on the PATH you will get a warning like:
+   > WARNING: The script conan.exe is installed in 'C:\Users\\%USERNAME%\\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\local-packages\Python312\Scripts' which is not on PATH.
+   > Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
+
+   On Windows, you can use _System Properties \> Environment Variables..._ to permanently add the directory to the user PATH. Restart the Command Prompt and run `conan --help` to ensure Conan is found.
```
jonathan-r-thorpe, line 17: I think this all refers to Windows only so perhaps prefix the whole section with "On Windows, ..."?
```suggestion
   On Windows, if the python Scripts directory is not on the PATH you will get a warning like:
   > WARNING: The script conan.exe is installed in 'C:\Users\\%USERNAME%\\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\local-packages\Python312\Scripts' which is not on PATH.
   > Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.

   Also on Windows, you can use _System Properties \> Environment Variables..._ to permanently add the directory to the user PATH. Restart the Command Prompt and run `conan --help` to ensure Conan is found.
```

garethsb: That specific example message is from Windows, but the warning can also occur on Linux.

PR #387, opened 2024-04-12
IS-14 support

lo-simon:
Add IS-14 configuration API (not complete yet)


Development/nmos/configuration_api.cpp
```diff
@@ -0,0 +1,654 @@
...
+    }
+
+    namespace details
+    {
+        namespace fields
+        {
+            const web::json::field_as_string_or level{ U("level"), {} };
+            const web::json::field_as_string_or index{ U("index"), {} };
+            const web::json::field_as_string_or describe{ U("describe"), {} };
+        }
```
maweit, line 66: Can be removed, i think. The code below uses `nmos::fields::nc::...`

lo-simon: Thanks, @maweit; well spotted; update the code next.

PR #402, opened 2024-07-30
Workaround GCC 10&11 optimiser compilation failure

alanb-sony:
GCC 10 and 11 seem to have an issue compiling the `std::lexicographical_compare` statements in `json_ops.h`. The issue seems to be that the operators used inside `std::lexicographical_compare` aren't defined before the call to `std::lexicographical_compare`.

This only seems to cause a problem for GCC 10 and 11, with c++20 and optimisations enabled. Clang always rejects the code.

A reduced example of the problem: https://godbolt.org/z/9ra6hn7c9

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -646,7 +646,7 @@ jobs:
         # ubuntu-14.04 ca-certificates are out of date
         git config --global http.sslVerify false
         # build and install openssl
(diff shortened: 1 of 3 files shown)


jonathan-r-thorpe (approved):
LGTM

lo-simon (approved):
LGTM

PR #413, opened 2024-10-17
IS-12: Fix datatype constraints check for NcDatatypeDescriptorStruct

jonathan-r-thorpe:
- Fixed checking of `isSequence` for Struct fields
- Fixed checking of `isNullable` for Struct fields
- Removed redundant iteration of sequences (this is done in the calling function)
- Always checks datatype constraints for Struct field (not just when there aren't any `NcParameterConstraints`)
- Improved `testConstraints` unit test to include testing of sequences and nullable Struct fields


lo-simon (approved):
LGTM

PR #415, opened 2024-11-01
IS-12: Ensure NcMethodResultError returned on error (not NcMethodResult)

jonathan-r-thorpe:
Consistently log  and return NcMethodResultError on error


lo-simon (approved):
LGTM

PR #418, opened 2024-11-21
Fix slog warning

lo-simon:
Use char* for slog::log to prevent warning.

### DIFF
--- a/Development/nmos/control_protocol_methods.cpp
+++ b/Development/nmos/control_protocol_methods.cpp
@@ -30,7 +30,7 @@ namespace nmos
         // unknown property
         utility::stringstream_t ss;
         ss << U("unknown property: ") << property_id.serialize() << U(" to do Get");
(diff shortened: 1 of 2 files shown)


garethsb:
Were got able to work out why this overload doesn't work in this case?

https://github.com/sony/nmos-cpp/blob/f54971298c47a633969e9e9adac824b56fc08da7/Development/nmos/slog.h#L15

lo-simon:
> Were you able to work out why this overload doesn't work in this case?
> 
> https://github.com/sony/nmos-cpp/blob/f54971298c47a633969e9e9adac824b56fc08da7/Development/nmos/slog.h#L15

The overload does work, but I keep getting the following warning, 
`warning C4239: nonstandard extension used: argument: conversion from slog::detail::anonymous-namespace::log<slog::base_gate,20> to slog::log_statement &` that why I made the change.

PR #420, opened 2024-11-29
Use if_indextoname to log the interface name for DNS-SD functions

garethsb:
On Linux, this helps because you see the familiar name in the log, e.g. `eth0`.

On Windows, it's not much use because `if_indextoname` returns a name like "ethernet_32778" that seems to be pretty meaningless. We have _cpprest/host_utils.cpp_ code using the `GetAdaptersAddresses` API to get `AdapterName`, but not `FriendlyName`, already but we don't currently have any dependencies from the **mdns** library on the **nmos-cpp** library, so I didn't want to do add coupling just for this. For now, I think you can continue to use PowerShell `Get-NetAdapter -InterfaceIndex N` to look up interface index N.

### DIFF
--- a/Development/mdns/service_discovery_impl.cpp
+++ b/Development/mdns/service_discovery_impl.cpp
@@ -1,5 +1,12 @@
 #include "mdns/service_discovery_impl.h"
 
+#if defined(_WIN32)
+#include <ws2tcpip.h>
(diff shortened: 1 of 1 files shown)


lo-simon (approved):
LGTM

PR #423, opened 2024-12-12
Add client assertion restrictions, and code tidy up

lo-simon:
- Replace `system_clock` with `steady_clock` for request-to-response duration calculation.
- Add client assertion restrictions.
- Add tests against client assertion restrictions.


jonathan-r-thorpe (approved):
LGTM

PR #424, opened 2024-12-18
Changed urns of default Node

PallesenJulius:
Not sure if the Urn here is just invalid and rusts Urn crate is more stringent or if the Urn crate is TOO stringent.

Nmos-cpp-registry always includes a default node when starting up which we could not parse using the translator because of this urn. pretty annoying.

### DIFF
--- a/Development/nmos/registry_resources.cpp
+++ b/Development/nmos/registry_resources.cpp
@@ -30,7 +30,7 @@ namespace nmos
                         .set_scheme(nmos::http_scheme(settings))
                         .set_port(nmos::experimental::fields::mdns_port(settings))
                         .set_path(U("/x-dns-sd/") + make_api_version(version));
-                    auto type = U("urn:x-dns-sd/") + make_api_version(version);
+                    auto type = U("urn:x-dns-sd:mdns-advertisement");
 
                     for (const auto& host : hosts)
                     {


garethsb:
RFC 8141 updated the spec from RFC 2141 to reflect the generic URI spec, explicitly making '/' a valid character in the namespace-specific string (NSS). See https://datatracker.ietf.org/doc/html/rfc8141#appendix-B.1.

However, I agree that per RFC 8141, the current URN for the DNS-SD API is invalid as it doesn't include both a NID (namespace identifier) and the NSS separated by ':'.

Mea culpa.

I suggest we use the NID 'x-dns-sd' and the NSS 'discover/vX.Y' (allowing for a potential NSS 'advertise/vX.Y').

I think the [sony/nmos-js](https://github.com/sony/nmos-js) client doesn't look up the API via the URN, so it's probably safe to make this change without reference to any other project...

Having said that, the API path might be better with an extra path segment as well, i.e., _/x-dns-sd/discover/vX.Y_...

PallesenJulius:
@garethsb Thanks a lot for the reply, i just submitted this here by accident since i was a bit too impatient when creating a PR to our fork and submitted it here by accident, sorry for the inconvenience.

PR #426, opened 2025-01-07
Add _or_defaults versions of the  format-specific parameters parsing functions

garethsb:
Add `_or_defaults` versions of the  format-specific parameters parsing functions; the current functions throw when required parameters are omitted. The intention is the new functions are for parsing partial `fmtp` lines, not handling invalid values, so they can still throw in those circumstances.

I should probably add `_or_defaults` versions for the other three media types even though none of those have required parameters, just so the naming and intended usage is clear?


lo-simon:
It looks good to me, and yes, it would definitely be beneficial to add the _or_defaults to the other media types for completeness.

Issue #414, opened 2024-10-27
MacOS/Cmake Supplied-JWT Linker Error


christopherbaine:
Hi, 
I've been attempting to building on MacOS, with the NMOS_CPP_USE_SUPPLIED_JWT_CPP flag to true, causes `Linking CXX static library libjwt-cpp.a` Error 

Assuming Boost/cpprestsdk are installed, 
To Replicate: 
`cmake -B build -DNMOS_CPP_USE_SUPPLIED_JWT_CPP=TRUE -DNMOS_CPP_USE_SUPPLIED_JSON_SCHEMA_VALIDATOR=TRUE -DWEBSOCKETPP_INCLUDE_DIR="Release/libs/websocketpp"`
`cmake --build build` 

Thanks

christopherbaine:
It looks like the section in `NmosCppDependencies.cmake` that compiles this creates a static library for a header only library. It even has a comment `# hm, header-only so should be INTERFACE library?`

garethsb:
I've checked and confirmed `NMOS_CPP_USE_SUPPLIED_JSON_SCHEMA_VALIDATOR` works OK on Linux... it constructs an empty _libjwt-cpp.a_ (8 Byte file). I suspect difference in behaviour with empty archives on macOS. The CI matrix only tests this flag on one platform. As the original author of the "hm" comment after the library was first added, I agree an `INTERFACE` library would be more elegant anyway... but I note one or two other potential issues - I think missing `install(FILES)` and I think wrong `target_include_directories`. I'll look at this when I get a chance.

christopherbaine:
I believe that is exactly the case, MacOS fails on an empty archive. Thanks for getting back to me!

lo-simon:
Closes, as this issue has been resolved in #431.

PR #431, opened 2025-02-05
Fix jwt-cpp INTERFACE library

lo-simon:
The `jwt-cpp` is a header-only library and should be set as an INTERFACE library. The issue was reported on #414.


jonathan-r-thorpe (approved):
LGTM

PR #437, opened 2025-02-26
IS-12 Bug Fix: Stop resource events being sent to control clients

jonathan-r-thorpe:
- IS-12 re-uses the data grain subscription mechanism for sending notifications to IS-12 Clients
- A side effect (bug) is that any resource events generated, for instance by deleting an NMOS resource, are also sent to the IS-12 Client
- This is unexpected as IS-12 Clients are only expecting IS-12 notifications
- Bug fix ignores resource events in the control protocol WebSocket subscription handling
- **This raises the question: could IS-12 notifications be inadvertently sent over IS-07 event and tally WebSocket when both are in use?**

### DIFF
--- a/Development/nmos/control_protocol_methods.cpp
+++ b/Development/nmos/control_protocol_methods.cpp
@@ -335,7 +335,7 @@ namespace nmos
                     // do notification that the specified property has changed
                     if (property_changed)
                     {
-                        property_changed(resource, nmos::fields::nc::name(property), index);
+                        property_changed(resource, nmos::fields::nc::name(property), -1);
                     }
 
                 }, make_property_changed_event(nmos::fields::nc::oid(resource.data), { { details::parse_nc_property_id(property_id), nc_property_change_type::type::sequence_item_removed, nc_id(index) } }));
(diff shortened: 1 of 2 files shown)


Development/nmos/control_protocol_methods.cpp
```diff
@@ -335,7 +335,7 @@ namespace nmos
                     // do notification that the specified property has changed
                     if (property_changed)
                     {
-                        property_changed(resource, nmos::fields::nc::name(property), index);
+                        property_changed(resource, nmos::fields::nc::name(property), -1);
```
lo-simon, line 338:
```suggestion
                        property_changed(resource, nmos::fields::nc::name(property), -2);
```

lo-simon: - Currently, we use `-1` for non-sequence changed; maybe we can use `-2` to identify this as a deleted sequence item changed. 
- We must update the comment for the `control_protocol_property_changed_handler` in control_protocol_handlers.h.
- Also, update the example, `make_node_implementation_control_protocol_property_changed_handler` in node_implementation.cpp

lo-simon (changes requested):
What do you think?

PR #438, opened 2025-03-07
Fix SDP session ID length limitation

lo-simon:
According to the [RFC 4566 5.2. Origin](https://tools.ietf.org/html/rfc4566#section-5.2)
```
<sess-id> is a numeric string such that the tuple of <username>,
 
<sess-id>, <nettype>, <addrtype>, and <unicast-address> forms a
 
globally unique identifier for the session.  The method of
 
<sess-id> allocation is up to the creating tool, but it has been
 
suggested that a Network Time Protocol (NTP) format timestamp be
 
used to ensure uniqueness [[13](https://datatracker.ietf.org/doc/html/rfc4566#ref-13)].
```
The current implementation limit for the session ID is UINT64_MAX (18446744073709551615), i.e., its maximum length is 20 digits. To fulfill the RFC above, the string should not be limited in length.

### DIFF
--- a/Development/nmos/sdp_utils.cpp
+++ b/Development/nmos/sdp_utils.cpp
@@ -1624,4 +1624,14 @@ namespace nmos
     {
         details::validate_sdp_parameters(details::format_constraints, sdp_params, details::get_format(sdp_params), details::get_format_parameters(sdp_params), receiver);
     }
+
+    // Validate the numeric string
+    const utility::string_t& valid_numeric_string(const utility::string_t& s)
+    {
+        if (!std::all_of(s.begin(), s.end(), ::isdigit))
+        {
+            throw std::invalid_argument("not a numeric string");
+        }
+        return s;
+    }
 }
--- a/Development/nmos/sdp_utils.cpp
+++ b/Development/nmos/sdp_utils.cpp
@@ -1628,7 +1628,7 @@ namespace nmos
     // Validate the numeric string
     const utility::string_t& valid_numeric_string(const utility::string_t& s)
     {
-        if (!std::all_of(s.begin(), s.end(), ::isdigit))
+        if (!std::all_of(s.begin(), s.end(), [](utility::char_t c) { return std::isdigit(c); }))
         {
             throw std::invalid_argument("not a numeric string");
         }
--- a/Development/nmos/sdp_utils.h
+++ b/Development/nmos/sdp_utils.h
@@ -5,6 +5,7 @@
 #include <functional>
 #include <map>
 #include <stdexcept>
+#include <boost/lexical_cast.hpp>
 #include "bst/any.h"
 #include "bst/optional.h"
 #include "cpprest/basic_utils.h"
@@ -88,7 +89,7 @@ namespace nmos
             origin_t() : session_id(), session_version() {}
             origin_t(const utility::string_t& user_name, uint64_t session_id, uint64_t session_version)
                 : user_name(user_name)
-                , session_id(std::to_wstring(session_id))
+                , session_id(boost::lexical_cast<utility::string_t>(session_id))
                 , session_version(session_version)
             {}
             origin_t(const utility::string_t& user_name, const utility::string_t& session_id, uint64_t session_version)
--- a/Development/nmos/sdp_utils.h
+++ b/Development/nmos/sdp_utils.h
@@ -84,24 +84,31 @@ namespace nmos
         {
             utility::string_t user_name;
             utility::string_t session_id;
-            uint64_t session_version;
+            utility::string_t session_version;
 
             origin_t() : session_id(), session_version() {}
             origin_t(const utility::string_t& user_name, uint64_t session_id, uint64_t session_version)
                 : user_name(user_name)
                 , session_id(boost::lexical_cast<utility::string_t>(session_id))
-                , session_version(session_version)
+                , session_version(boost::lexical_cast<utility::string_t>(session_version))
             {}
-            origin_t(const utility::string_t& user_name, const utility::string_t& session_id, uint64_t session_version)
+            origin_t(const utility::string_t& user_name, const utility::string_t& session_id, const utility::string_t& session_version)
                 : user_name(user_name)
                 , session_version(session_version)
             {
                 // validate session_id is a numeric string
                 if (!std::all_of(session_id.begin(), session_id.end(), ::isdigit))
                 {
-                    throw std::invalid_argument("session_id must be a numeric string");
+                    throw std::invalid_argument("session id must be a numeric string");
                 }
                 this->session_id = session_id;
+
+                // validate session_version is a numeric string
+                if (!std::all_of(session_version.begin(), session_version.end(), ::isdigit))
+                {
+                    throw std::invalid_argument("session version must be a numeric string");
+                }
+                this->session_version = session_version;
             }
             origin_t(const utility::string_t& user_name, uint64_t session_id_version)
                 : origin_t{ user_name, session_id_version, session_id_version }
--- a/Development/nmos/sdp_utils.h
+++ b/Development/nmos/sdp_utils.h
@@ -82,19 +82,28 @@ namespace nmos
         struct origin_t
         {
             utility::string_t user_name;
-            uint64_t session_id;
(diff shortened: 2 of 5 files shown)


garethsb:
This is at the least incomplete since the same argument applies to `sess-version`.

lo-simon:
> This is at the least incomplete since the same argument applies to `sess-version`.

Well spotted @garethsb. Will extend the change to `sess-version`.

Development/sdp/sdp_grammar.cpp
```diff
@@ -68,6 +68,22 @@ namespace sdp
...
+            if (v.is_number()) return jn2s(v);
+            else return js2s(v);
+        }
+        inline web::json::value digits2jns(const std::string& s)
+        {
+            if (!std::all_of(s.begin(), s.end(), ::isdigit)) throw sdp_parse_error("expected a sequence of digits");
+
+            uint64_t v;
+            std::istringstream is(s);
+            is >> v;
```
garethsb, line 83: This doesn't seem to solve the problem that 1*DIGITS strings that begin with leading zeroes are not roundtrippable. If you're going to store these as strings do you want to try to solve that too?

Cf. https://github.com/sony/nmos-cpp/commit/062be351ebd58565f91ba1a51c858b9cf19a2d0a#diff-159bd613b3d4d427540279e8cb51487660592ac0bf8634ba88a92dabb7527470R61

Development/sdp/test/sdp_test.cpp
```diff
@@ -571,3 +571,35 @@ a=fmtp:96)";
...
+    BST_REQUIRE_NO_THROW(sdp::parse_session_description(before + "184467440737095516150" + after)); // session id greater than UINT64_MAX
+    // an invalid session id results in "sdp parse error - expected a sequence of digits at line 2"
+    BST_REQUIRE_THROW(sdp::parse_session_description(before + "foo" + after), std::runtime_error);
+    BST_REQUIRE_THROW(sdp::parse_session_description(before + "0foo" + after), std::runtime_error);
+    BST_REQUIRE_THROW(sdp::parse_session_description(before + "0.0" + after), std::runtime_error);
+    BST_REQUIRE_THROW(sdp::parse_session_description(before + "0x0" + after), std::runtime_error);
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////
+BST_TEST_CASE(testSdpSessionVesrion)
```
garethsb, line 592:
```suggestion
BST_TEST_CASE(testSdpSessionVersion)
```

Development/nmos/sdp_utils.h
```diff
@@ -5,6 +5,7 @@
 #include <functional>
 #include <map>
 #include <stdexcept>
+#include <boost/lexical_cast.hpp>
```
garethsb, line 8: I think you should be able to use `utility::conversions::to_string_t` or `utility::conversions:: details::to_string_t`, instead?

Development/sdp/sdp_grammar.cpp
lo-simon: For allowing leading zeroes, I think I do the following will do?
```
inline web::json::value digits2jns(const std::string& s)
{
    if (!std::all_of(s.begin(), s.end(), ::isdigit)) throw sdp_parse_error("expected a sequence of digits");
    return s2js(s);
}
```

Development/nmos/sdp_utils.h
```diff
@@ -82,19 +83,35 @@ namespace nmos
...
-                , session_version(session_version)
+                , session_id(boost::lexical_cast<utility::string_t>(session_id))
+                , session_version(boost::lexical_cast<utility::string_t>(session_version))
             {}
-            origin_t(const utility::string_t& user_name, uint64_t session_id_version)
+            origin_t(const utility::string_t& user_name, const utility::string_t& session_id, const utility::string_t& session_version)
                 : user_name(user_name)
-                , session_id(session_id_version)
-                , session_version(session_id_version)
+                , session_version(session_version)
```
garethsb, line 97: You've still got a member initializer here as well as the code in the constructor body...

Might be nice to use the member initializers for both, like this:

```
const utility::string_t& valid_numeric_string(const utility::string_t& s)
{
  if (!std::all_of(s.begin(), s.end(), ::isdigit))
  {
    throw std::invalid_argument("not a numeric string");
  }
  return s;
}
...
  , session_id(valid_numeric_string(session_id))
  , session_version(valid_numeric_string(session_version))
```
Although you do lose the precision of the error message of course.

Development/nmos/sdp_utils.h
lo-simon: Thanks, `utility::conversions::details::to_string_t` will do the trick.

Development/sdp/sdp_grammar.cpp
garethsb: Yeah, I think so!

Development/sdp/sdp_grammar.cpp
```diff
@@ -68,6 +68,17 @@ namespace sdp
...
         }
 
+        inline std::string jns2s(const web::json::value& v)
+        {
+            if (v.is_number()) return jn2s(v);
+            else return js2s(v);
+        }
+        inline web::json::value digits2jns(const std::string& s)
+        {
+            if (!std::all_of(s.begin(), s.end(), ::isdigit)) throw sdp_parse_error("expected a sequence of digits");
```
garethsb, line 78: `return s2js(valid_numeric_string(s))`?

Development/sdp/sdp_grammar.cpp
```diff
@@ -96,6 +107,8 @@ namespace sdp
 
         const converter digits_converter{ jn2s, digits2jn };
 
+        const converter long_digits_converter{ jns2s, digits2jns };
```
garethsb, line 110: Very minor quibble... Long is often understood to refer to the `long` type... Maybe `big_digits_converter` as in "BigNum", something which is for holding numbers that are bigger than  the largest built-in type?

Development/nmos/sdp_utils.cpp
```diff
@@ -1624,4 +1624,14 @@ namespace nmos
     {
         details::validate_sdp_parameters(details::format_constraints, sdp_params, details::get_format(sdp_params), details::get_format_parameters(sdp_params), receiver);
     }
+
+    // Validate the numeric string
+    const utility::string_t& valid_numeric_string(const utility::string_t& s)
+    {
+        if (!std::all_of(s.begin(), s.end(), ::isdigit))
```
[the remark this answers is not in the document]
lo-simon: Thank you so much @garethsb, learn something new every time. I'll fix it as suggested.

Development/sdp/test/sdp_test.cpp
```diff
@@ -100,8 +100,8 @@ a=mid:SECONDARY
         { sdp::fields::protocol_version, 0 },
         { sdp::fields::origin, web::json::value_of({
             { sdp::fields::user_name, U("-") },
-            { sdp::fields::session_id, 3745911798u },
-            { sdp::fields::session_version, 3745911798u },
+            { sdp::fields::session_id, U("3745911798") },
```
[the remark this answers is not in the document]
lo-simon: They are covered in the `BST_TEST_CASE(testSdpSessionId)` and `BST_TEST_CASE(testSdpSessionVersion)`?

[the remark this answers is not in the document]
garethsb: No, I don't think so, those check only parsing from textual SDP to JSON. My concern was whether an application developer who assumes the value in the JSON is a number as before will now get an exception either when they process the parsed JSON or when they construct the JSON with a number and generate SDP text. You can't make a value be both a number and a string of course! The generation could handle both (does it? this is the original question I was asking) but the parser can't. I guess given the type of the accessor object has changed they will get a compile error if they use that? Maybe that's good enough if so...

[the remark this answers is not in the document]
lo-simon: Thanks @garethsb, a new test `testSdpSessionIdSessionVersionToSdp` has been added to confirm that session id and session version can be numbers or strings for constructing the texture SDP.

PR #439, opened 2025-03-10
IS-12: Fix sequence item removed bug

jonathan-r-thorpe:
- Example property changed handler was attempting to read a removed value causing an exception.
- Item removed is now explicitly indicated to the property changed handler.


lo-simon (approved):
LGTM

PR #442, opened 2025-04-02
fix the st2110_21_sender_type checker for the jxsv_constraints, it wa…

lo-simon:
…s reported by Niitsu, Keita-san (Sony)


jonathan-r-thorpe (approved):
LGTM

PR #446, opened 2025-04-24
Fix c++11 build for Ubuntu 14.04

lo-simon:


Development/nmos/mutex.h
```diff
@@ -1,6 +1,7 @@
 #ifndef NMOS_MUTEX_H
 #define NMOS_MUTEX_H
 
+#include <condition_variable>
```
garethsb, line 4: This header should come from _bst/shared_mutex.h_ and if there's a case where it should be but isn't, the fix belongs there, I think?

lo-simon: The <condition_variable> header only comes from _bst/shared_mutex. h_ when BST_THREAD_BOOST is not defined. We have the `BST_THREAD_BOOST` set as the default in the build, i.e., we always use boost as the default.

we should include the <mutex> for std::unique_lock instead of the <condition_variable>

PR #447, opened 2025-04-26
Update README.md

Joysephk:
In the README under the description of the "nmos-cpp-registry application", the word "utilising" is used (British English). Depending on the style guide you want (e.g., if you expect American English throughout), it would be better to change it to "utilizing" for consistency if the rest is in American English.


Joysephk:
In the README under the description of the "nmos-cpp-registry application", the word "utilising" is used (British English). Depending on the style guide you want (e.g., if you expect American English throughout), it would be better to change it to "utilizing" for consistency if the rest is in American English.

PR #454, opened 2025-07-18
Windows RDS becomes unresponsive when many devices connect

lo-simon:
In Windows, there is a maximum number of exclusive locks that can be held. Once it reaches the maximum value, the lock will throw a lock exception. Although the registration API can handle the exception, many of the nmos-cpp threads, such as `send_events_ws_messages_thread` and `erase_expired_events_resources_thread`, will terminate because they do not handle the exception. This makes RDS unresponsive.  To resolve this issue, it is essential to ensure that the exception is handled correctly.

### DIFF
--- a/Development/nmos/thread_utils.h
+++ b/Development/nmos/thread_utils.h
@@ -15,7 +15,21 @@ namespace nmos
         public:
             typedef BasicLockable mutex_type;
             explicit reverse_lock_guard(mutex_type& m) : m(m) { m.unlock(); }
-            ~reverse_lock_guard() { m.lock(); }
+            ~reverse_lock_guard()
+            {
+                for (;;)
+                {
+                    try
+                    {
+                        m.lock();
+                        break;
+                    }
+                    catch (...)
+                    {
+                        // ignore exception
+                    }
+                }
+            }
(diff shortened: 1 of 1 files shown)


Development/nmos/events_ws_api.cpp
```diff
@@ -504,27 +504,42 @@ namespace nmos
             // otherwise, there's actually work to do...
 
             details::reverse_lock_guard<nmos::read_lock> unlock(lock);
-            // note, without atomic upgrade, another thread may preempt hence the need to recalculate expire_health/forget_health and least_health
-            auto upgrade = model.write_lock();
+            // note 1, without atomic upgrade, another thread may preempt hence the need to recalculate expire_health/forget_health and least_health
+            // note 2, the try-catch block is used here because the Windows version of the `boost::shared_mutex::lock` throws lock exceptions when
+            //         it has reached the maximum number of 128 exclusive_waiting locks. Another implementation could be replacing the initial
```
jonathan-r-thorpe, line 509:
```suggestion
            //         it has reached the maximum number of 128 exclusive_waiting locks. As an alternative we could replace the initial
```

PR #455, opened 2025-07-18
Fix condition variable wait on Windows

lo-simon:
In Windows, there is a maximum number of exclusive locks that can be held. Once it reaches the maximum value, the lock will throw a lock exception. This enhancement improves the low-level relocker's handling of lock exceptions in its destructor.


jonathan-r-thorpe (approved):
LGTM

Issue #399, opened 2024-06-24
Conan version


garethsb:
Latest Conan version is 2.4.1.
DBus recipe now requires Conan >=2.2.0.

I also misunderstood that the version specifier `~=2.0.5` was equivalent to `>=2.0.5,<3`, whereas in fact it is like `>=2.0.5,==2.0.*`.

So I think we should probably change guidance in docs from:
```sh
pip install --upgrade conan~=2.0.5
```
to:
```sh
pip install --upgrade conan>=2.2.0,<3
```

lo-simon:
Thanks @garethsb; we will update the doc; now using Conan 2.4.1

`pip install --upgrade conan~=2.0.5`
to
`pip install --upgrade conan~=2.4`

garethsb:
@lo-simon there are still install commands in both docs and the GitHub action workflow using the `~=x.y.z` form which is probably not desirable?

PR #465, opened 2025-09-26
Up conan version

lo-simon:
Update the Conan version to 2.20.1, refer to #399


lo-simon:
> As described in #399, it would be better to use x.y not x.y.z unless you really want to limit to this specific minor version...

Thanks, will update to 2.20

PR #474, opened 2025-11-14
Add IS-11 support

lo-simon:
- Add I[S-11 NMOS Stream Compatibility Management](https://specs.amwa.tv/is-11/) support.
- This is superseded #271 which was initially created by @N-Nagorny


Development/nmos-cpp-node/node_implementation.cpp
```diff
@@ -203,6 +215,19 @@ namespace impl
...
+        if (nmos::media_types::video_jxsv == nmos::get_media_type(sdp_params))
+        {
+            nmos::validate_video_jxsv_sdp_parameters(receiver, sdp_params);
+        }
+        else
+        {
+            // validate core media types, i.e., "video/raw", "audio/L", "video/smpte291" and "video/SMPTE2022-6"
+            nmos::validate_sdp_parameters(receiver, sdp_params);
+        }
+    };
```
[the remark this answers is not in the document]
lo-simon: Yes, the same `validate_sdp_parameters` function can be reused, now fixed.

Development/nmos-cpp-node/node_implementation.cpp
```diff
@@ -938,6 +964,113 @@ void node_implementation_init(nmos::node_model& model, nmos::experimental::contr
...
+            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00
+        };
+        utility::string_t edid(edid_bytes, edid_bytes + sizeof(edid_bytes));
+
+        const auto output_id = impl::make_id(seed_id, nmos::types::output);
+        const auto receiver_ids = impl::make_ids(seed_id, nmos::types::receiver, { impl::ports::video, impl::ports::audio }, how_many);
+
+        auto output = edid_support
+            ? nmos::experimental::make_streamcompatibility_output(output_id, device_id, true, boost::variant<utility::string_t, web::uri>(edid), receiver_ids, model.settings)
+            : nmos::experimental::make_streamcompatibility_output(output_id, device_id, true, receiver_ids, model.settings);
```
jonathan-r-thorpe: Hmmm, what's the behaviour if EDID support is set to true in the config? Does this pass any of the EDID tests in the NMOS Testing tool?

Development/nmos/constraints.cpp
```diff
@@ -0,0 +1,271 @@
...
+                {
+                    insert_if_not_meta(result, lhs_iter++);
+                }
+            }
+
+            return result;
+        }
+
+        // Constraint B is a subconstraint of Constraint A if:
+        // 1. Constraint B has enum keyword when Constraint A has it and enumerated values of Constraint B are a subset of enumerated values of Constraint A
```
jonathan-r-thorpe: "Constraint B has enum keyword when Constraint __*also*__ has it" perhaps?

simonbaren:
@jonathan-r-thorpe I had another look on the IS-11 project - the EDID code (constraints_sets) is just a stub and doesn't actually modify the effective_edid. Might that be the issue you are seeing?

N-Nagorny:
Hello guys! Glad to see the PR was revived.

Yes, three tests fail.

| Test ID | Test name | Description | Resolution |
| ------------- | ------------- | ------------- | ------------- |
| test_01_03 | Inputs with Base EDID support reject an invalid EDID  | The test PUTs a nulled binary to Base EDID. EDID parsing was out of scope for the reference impementation, so `nmos-cpp` accepts any EDID (returns [204](http://httpstatus.es/204)).| We can implement basic EDID support. https://github.com/N-Nagorny/libedid could come in handy here, though I can't commit to maintaining it. Alternatively, we can `--ignore` it in `run_nmos_testing.sh`. |
| test_02_03_05_01, test_02_03_05_02 | Verify for inputs supporting EDID that the version and the effective EDID change when applying constraints | The test tries to constrain a Sender to certain `grain_rate` for video and `sample_rate` for audio respectively. `node_implementation.cpp` declares Sender Capabilities that allow only single value of `grain_rate` and `sample_rate`, so Node can't adhere to `constraint_sets` from the tests and returns [422](http://httpstatus.es/422). This valid response should result into "Could Not Test" but the test suite has a bug in it (fixed in https://github.com/AMWA-TV/nmos-testing/pull/852). | Revive https://github.com/AMWA-TV/nmos-testing/pull/852 |

simonbaren:
Hi @N-Nagorny .
About test_02_03_05_01 and test_02_03_05_02 - those are very Generator reliant. 
I mean - if your source Generator comply to EDID changes - test will fail.
The solution to that (was applied by IPMX I think) is to add a manual step for those tests. When test runs, the tester is prompted to go and read the EDID from the device and check that it changed.

As a semi-automatic test, it is possible to read the effective_edid and validate that it was constrained according to the request, but this is kind of wrong, because the test should check what is on the wire side and not the nmos side.

regarding test_01_03 - it pass for the implementation we run (we do use the libedid: Edid::parse_edid_binary(edid)) .

N-Nagorny:
> I mean - if your source Generator comply to EDID changes - test will fail.

Should it have been "doesn't comply" or I'm missing something?

> was applied by IPMX I think

Could you please share links?

> because the test should check what is on the wire side and not the nmos side

Effective EDID just tells what the modeled device gives to its upstream input. NMOS Tests don't check deeper than the NMOS side by design, like they don't capture UDP streams after activating a Sender.

Could you please tell more about your implementation and run https://github.com/AMWA-TV/nmos-testing/pull/852 against it?

simonbaren:
@N-Nagorny 
> Should it have been "doesn't comply" or I'm missing something?

Yes, "doesn't comply" is correct — my mistake.

> Could you please share links?

Sorry but I don't have a direct link. It's an additional test configuration where the tester is prompted to read the EDID from the DUT at a certain point and verify it changed on the wire.

> Effective EDID just tells what the modeled device gives to its upstream input. NMOS Tests don't check deeper than the NMOS side by design, like they don't capture UDP streams after activating a Sender.

The UDP analogy doesn't hold here. A sender that doesn't stream what it's supposed to would still fail downstream — the UDP stream itself is independently verifiable. But an EDID change that never reaches the wire has no such natural verification. A node could trivially update effective_edid on the NMOS side without ever pushing the change to the physical input — making it look compliant while not actually being so. The example node_implementation is actually a perfect illustration of this: there is no real HW behind it, so effective_edid is just a value in memory with no physical effect. That's exactly the gap the manual step is designed to close, and it's also why a node that only "looks good" on the NMOS side is not truly IS-11 compliant.

PR #483, opened 2026-03-05
AMWA NMOS IS-05 1.2-dev / MXL Support

jonathan-r-thorpe:
- Prototype used to demonstrate MXL integration with NMOS


Development/nmos-cpp-node/config.json
```diff
@@ -22,9 +22,10 @@
 
     // senders, receivers: controls which kinds of sender and receiver are instantiated by the example node
     // the values must be an array of unique strings identifying the kinds of 'port', like ["v", "a", "d"], see impl::ports
+    // for MXL Senders and Receivers, the values must be ["xv", "xa", "xd"]
     // when omitted, all ports are instantiated
-    //"senders": ["v", "a"],
-    //"receivers": [],
+    //"senders": ["v", "a", "d"],
+    //"receivers": ["v", "a", "d"],
```
garethsb: Some of the comment examples in this file show what the equivalent of the default would be, some show interesting non-default examples. This was the latter (only create senders, don't create any receivers), now it's the former. Change made intentionally?

Development/nmos-cpp-node/config.json
```diff
@@ -51,8 +52,8 @@
     // component_depth: controls the bits per component sample of video flows
     //"component_depth": 10,
 
-    // video_type: media type of video flows, e.g. "video/raw" or "video/jxsv", see nmos::media_types
-    //"video_type": "video/jxsv",
+    // video_type: media type of video flows, e.g. "video/raw", "video/jxsv", or "video/v210" for MXL Senders and Receivers, see nmos::media_types
```
garethsb:
```suggestion
    // video_type: media type of video flows, e.g. "video/raw", "video/jxsv" for ST 2110 senders and receivers
    // or "video/v210" for MXL senders and receivers, see nmos::media_types
```

Development/nmos-cpp-node/config.json
```diff
@@ -22,9 +22,10 @@
 
     // senders, receivers: controls which kinds of sender and receiver are instantiated by the example node
     // the values must be an array of unique strings identifying the kinds of 'port', like ["v", "a", "d"], see impl::ports
+    // for MXL Senders and Receivers, the values must be ["xv", "xa", "xd"]
```
garethsb: Although in docs we normally Title Case the NMOS keywords, that seems not to be the case in this file. Let's not end up with a mix anyway?

```suggestion
    // for MXL senders and receivers, the values must be like ["xv", "xa", "xd"]
```

Development/nmos/connection_api.cpp
```diff
@@ -483,6 +519,20 @@ namespace nmos
                     throw std::logic_error("matching IS-04 and IS-05 resources not found");
                 }
 
+                const nmos::transport transport_subclassification(nmos::fields::transport(matching_resource->data));
+
+                // BCP-007-03: PATCH /staged for MXL receivers must not contain transport_file.
+                // See https://specs.amwa.tv/bcp-007-03/branches/publish-auto-null/docs/NMOS-With-MXL.html
```
garethsb: @jonathan-r-thorpe I think this is a mistake in the spec? Given that the response schema requires that `transport_file` is present, I think it should be valid to stage:
```json
"transport_file": {
  "type": null,
  "data": null
}
```

Development/nmos/is05_schemas/is05_schemas.h
```diff
@@ -7,6 +7,28 @@ namespace nmos
 {
     namespace is05_schemas
     {
+        namespace v1_2_x
```
garethsb:
```suggestion
        namespace v1_2_dev
```

Development/nmos/json_fields.h
```diff
@@ -172,6 +172,9 @@ namespace nmos
         const web::json::field_as_value_or broker_authorization{ U("broker_authorization"), {} }; // string or bool
         const web::json::field_as_value_or broker_topic{ U("broker_topic"), {} }; // string or null
         const web::json::field_as_value_or connection_status_broker_topic{ U("connection_status_broker_topic"), {} }; // string or null
+        // for urn:x-nmos:transport:mxl (see AMWA BCP-007-03 NMOS With MXL)
+        const web::json::field_as_value_or mxl_domain_id{ U("mxl_domain_id"), {} }; // UUID string, auto, or null
+        const web::json::field_as_value_or mxl_flow_id{ U("mxl_flow_id"), {} }; // senders: UUID, auto, or null; receivers: UUID or null (BCP-007-03)
```
garethsb: not sure why this needs the parenthetical? the common above has the spec reference for both transport parameters

Development/nmos/json_schema.cpp
```diff
@@ -77,6 +77,16 @@ namespace nmos
             return{ _XPLATSTR("https://github.com/AMWA-TV/is-05/raw/") + tag + _XPLATSTR("/APIs/schemas/") + ref };
         }
 
+        // See https://github.com/AMWA-TV/is-05/blob/v1.2.x/APIs/schemas/
+        namespace v1_2
+        {
+            using namespace nmos::is05_schemas::v1_2_x;
+            const utility::string_t tag(_XPLATSTR("v1.2.x"));
```
garethsb:
```suggestion
            using namespace nmos::is05_schemas::v1_2_dev;
            const utility::string_t tag(_XPLATSTR("v1.2-dev"));
```

Development/nmos/json_schema.cpp
```diff
@@ -77,6 +77,16 @@ namespace nmos
             return{ _XPLATSTR("https://github.com/AMWA-TV/is-05/raw/") + tag + _XPLATSTR("/APIs/schemas/") + ref };
         }
 
+        // See https://github.com/AMWA-TV/is-05/blob/v1.2.x/APIs/schemas/
```
garethsb:
```suggestion
        // See https://github.com/AMWA-TV/is-05/blob/v1.2-dev/APIs/schemas/
```

Development/third_party/is-05/v1.2-dev/APIs/schemas/activation-response-schema.json
garethsb, line 1: Move all of these files to `v1.2-dev`?

Development/nmos-cpp-node/node_implementation.cpp
```diff
@@ -400,13 +412,18 @@ void node_implementation_init(nmos::node_model& model, nmos::experimental::contr
     const auto seed_id = nmos::experimental::fields::seed_id(model.settings);
     const auto node_id = impl::make_id(seed_id, nmos::types::node);
     const auto device_id = impl::make_id(seed_id, nmos::types::device);
+    const nmos::id mxl_domain_id = model.settings.has_field(impl::fields::mxl_domain_id)
+        ? impl::fields::mxl_domain_id(model.settings)
+        : nmos::make_repeatable_id(seed_id, U("/x-nmos/mxl/domain"));
```
garethsb: Does it make sense to provide a default for this? Doing so seems to imply it would be reasonable for a media function to create its own domain (which feels wrong)?

Development/nmos-cpp-node/node_implementation.cpp
```diff
@@ -903,6 +921,168 @@ void node_implementation_init(nmos::node_model& model, nmos::experimental::contr
...
+                    video_type,
+                    model.settings
+                );
+            }
+            else if (impl::ports::mxl_audio == port)
+            {
+                flow = nmos::make_raw_audio_flow(flow_id, source_id, device_id, 48000, 32, model.settings);
+                flow.data[nmos::fields::media_type] = value::string(U("audio/float32"));
+                // add optional grain_rate
+                flow.data[nmos::fields::grain_rate] = nmos::make_rational(frame_rate);
```
garethsb: Given MXL moves us even further from SDI than ST 2022-6 and ST 2110, maybe we drop the (video) grain rate on audio?

Development/nmos-cpp-node/node_implementation.cpp
```diff
@@ -903,6 +921,168 @@ void node_implementation_init(nmos::node_model& model, nmos::experimental::contr
...
+                    frame_width, frame_height, interlace_mode,
+                    colorspace, transfer_characteristic, sampling, bit_depth,
+                    video_type,
+                    model.settings
+                );
+            }
+            else if (impl::ports::mxl_audio == port)
+            {
+                flow = nmos::make_raw_audio_flow(flow_id, source_id, device_id, 48000, 32, model.settings);
+                flow.data[nmos::fields::media_type] = value::string(U("audio/float32"));
```
garethsb: ⬇️ Maybe we should use `make_audio_flow` or have another overload of `make_raw_audio_flow`, so we don't have to rewrite the `media_type`?

Development/nmos-cpp-node/node_implementation.cpp
```diff
@@ -903,6 +921,168 @@ void node_implementation_init(nmos::node_model& model, nmos::experimental::contr
...
+                        value_of({
+                            { nmos::caps::format::profile, nmos::make_caps_string_constraint({ profile.name }) },
+                            { nmos::caps::format::level, nmos::make_caps_string_constraint({ level.name }) },
+                            { nmos::caps::format::sublevel, nmos::make_caps_string_constraint({ nmos::sublevels::Sublev3bpp.name, nmos::sublevels::Sublev4bpp.name }) },
+                            { nmos::caps::format::bit_rate, nmos::make_caps_integer_constraint({}, nmos::no_minimum<int64_t>(), (int64_t)max_format_bit_rate) },
+                            { nmos::caps::transport::bit_rate, nmos::make_caps_integer_constraint({}, nmos::no_minimum<int64_t>(), (int64_t)max_transport_bit_rate) },
+                            { nmos::caps::transport::packet_transmission_mode, nmos::make_caps_string_constraint({ nmos::packet_transmission_modes::codestream.name }) }
+                        })
+                    });
+                }
```
garethsb: Needs fixing to work with `video/v210` and maybe `video/v210a`.

Development/nmos-cpp-node/node_implementation.cpp
```diff
@@ -903,6 +921,168 @@ void node_implementation_init(nmos::node_model& model, nmos::experimental::contr
...
+                receiver.data[nmos::fields::version] = receiver.data[nmos::fields::caps][nmos::fields::version] = value(nmos::make_version());
+            }
+            else if (impl::ports::mxl_audio == port)
+            {
+                receiver = nmos::make_receiver(receiver_id, device_id, nmos::transports::mxl, {}, nmos::formats::audio, { nmos::media_type{ U("audio/float32") } }, model.settings);
```
garethsb: We should add an `nmos::media_types::audio_float32` constant? (And the MXL video ones too.)

garethsb (changes requested):
Forgot to submit this last week, sorry...

Development/nmos/connection_api.cpp
jonathan-r-thorpe: The BCP-007-03 spec is ambiguous on this so I'll raise it as an issue on the spec.
BCP-007-03 says PATCH /staged for MXL receivers is “not expected to contain” transport_file, and that controllers “MUST NOT provide” the attribute — which I read as prohibiting the key entirely. That’s stricter than the IS-05 stage schema, as you point out. I think this might become moot if the position on `transport_file` changes in the spec...

Development/nmos-cpp-node/config.json
jonathan-r-thorpe: No, hang over from modified example config. Will revert,

Development/nmos/json_fields.h
jonathan-r-thorpe: Agreed, redundant and will remove.

Development/third_party/is-05/v1.2-dev/APIs/schemas/activation-response-schema.json
jonathan-r-thorpe: Yes, agreed

Development/nmos/connection_api.cpp
garethsb: Yeah, I think that MUST NOT is a mistake that makes an edge case where there need not be one.

Development/nmos-cpp-node/node_implementation.cpp
garethsb: @jonathan-r-thorpe do we need an `nmos/mxl.h` which has:

```C++
#ifndef NMOS_MXL_H
#define NMOS_MXL_H

#include "nmos/media_type.h"

namespace nmos
{
    namespace media_types
    {
        // MXL video media types

        const media_type video_v210{ U("video/v210") };
        const media_type video_v210a{ U("video/v210a") };

        // MXL audio media types

        const media_type audio_float32{ U("audio/float32") };

        // MXL data media types

        //const media_type video_smpte291{ U("video/smpte291") };
    }
}

#endif
```

Development/nmos-cpp-node/node_implementation.cpp
jonathan-r-thorpe: Agreed - although it is defining the ID not the MXL domain, in the limit it makes no sense for each and every NMOS Node to create it's own domain ID for the same MXL domain.

Development/nmos-cpp-node/node_implementation.cpp
jonathan-r-thorpe: Use `nmos::make_coded_audio_flow` perhaps?

Development/nmos-cpp-node/node_implementation.cpp
jonathan-r-thorpe: Agreed

Development/nmos-cpp-node/node_implementation.cpp
jonathan-r-thorpe: Yes agreed.

Development/nmos-cpp-node/node_implementation.cpp
garethsb: Yeah... except we want `bit_depth`, I think?

Development/nmos-cpp-node/node_implementation.cpp
```diff
@@ -119,6 +120,9 @@ namespace impl
         // video_type: media type of video flows, e.g. "video/raw" or "video/jxsv", see nmos::media_types
         const web::json::field_as_string_or video_type{ U("video_type"), U("video/raw") };
 
+        // mxl_video_type: media type of MXL video flows and receivers, e.g. "video/v210" or "video/v210a", see nmos/mxl.h
+        const web::json::field_as_string_or mxl_video_type{ U("mxl_video_type"), nmos::media_types::video_v210.name };
+
```
garethsb, line 128: @jonathan-r-thorpe I can't decide if it's a good idea to use the `nmos::media_types::video_v210.name` here (so long as it never suffers from static initialization order fiasco? that might be the reason that we didn't do it above for `"video/raw"`?)

Sandbox/run_nmos_testing.sh
```diff
@@ -202,7 +202,7 @@ if $secure; then
   do_run_test BCP-003-01 $expected_disabled_BCP_003_01 --host "${host}" --port 1080 --version v1.0
 fi
 
-do_run_test IS-04-01 $expected_disabled_IS_04_01 --host "${host}" --port 1080 --version v1.3
+do_run_test IS-04-01 $expected_disabled_IS_04_01 --host "${host}" --port 1080 --version v1.3 --ignore test_19
```
garethsb: Do we need to raise an issue or PR on the NMOS Testing Tool?

jonathan-r-thorpe: Already did it https://github.com/AMWA-TV/nmos-testing/pull/895 👍

Development/nmos-cpp-node/node_implementation.cpp
jonathan-r-thorpe: Good catch - will update

Development/nmos-cpp-node/node_implementation.cpp
jonathan-r-thorpe: Keep it consistent I think.

garethsb:
@jonathan-r-thorpe @lo-simon If this isn't going to get merged imminently (although that would be nice!) would you consider extracting d499aaccc06665b4f98a4bc0000d25aa85c9c4d4 into a separate PR so we can have running CI on Windows again?

Development/third_party/is-05/v1.2-dev/APIs/schemas/constraints-schema.json
```diff
@@ -0,0 +1,24 @@
...
+        "$ref": "constraints-schema-websocket.json"
+      }
+    },
+    {
+      "type": "array",
+      "items": {
+        "$ref": "constraints-schema-mqtt.json"
+      }
+    }
+  ]
```
garethsb, line 29: Should add `constraints-schema-mxl.json` and an `$ref` to that here.

Development/nmos/connection_api.cpp
garethsb: Removing this will mean that the code ends up calling the `transport_file_parser` supplied by the application. That has an easy implementation for MXL receivers in `make_node_implementation_transport_file_parser()` in `nmos-cpp-node/node_implementation.cpp`:

```
throw std::runtime_error("MXL does not use a transport_file");
```

Development/third_party/is-05/v1.2-dev/APIs/schemas/constraints-schema.json
jonathan-r-thorpe: Hmmm, it seems that these `constraints` schemas are not consistent with the ones in IS-05 1.2-dev.  For instance the `constraints-schema.json` in v1.2-dev has been relaxed to allow any transport.  i.e. there is no `constraints-schema-mxl.json` in IS-05 v1.2-dev. We should harmonize these schemas with the v1.2-dev specs and allow "any" transport, or should we explicitly check MXL?

jonathan-r-thorpe: After some thinking, I decided that allowing *any* transport means the schema has little value considering the code only supports a defined subset.  So I have added the `constraints-schema-mxl.json` 👍

garethsb: Yes, I think that was the rationale for how to glue the two things (extensible IS-05 v1 2 schemas and the specific set of transports that nmos-cpp implements at any one time) together. 👍

Development/nmos/node_resources.cpp
```diff
@@ -397,13 +397,20 @@ namespace nmos
 
     // See https://specs.amwa.tv/is-04/releases/v1.2.0/APIs/schemas/with-refs/flow_audio_raw.html
     nmos::resource make_raw_audio_flow(const nmos::id& id, const nmos::id& source_id, const nmos::id& device_id, const nmos::rational& sample_rate, unsigned int bit_depth, const nmos::settings& settings)
+    {
+        return make_raw_audio_flow(id, source_id, device_id, sample_rate, nmos::media_types::audio_L(bit_depth), bit_depth, settings);
+    }
+
+    // See https://specs.amwa.tv/is-04/releases/v1.2.0/APIs/schemas/with-refs/flow_audio_raw.html
+    // (media_type must *not* be nmos::media_types::audio_L(bit_depth); cf. the bit_depth overload)
```
garethsb: I don't think that's true, if you use it correctly, so `must *not*` seems a bit strong? I'd just delete this unless I'm missing something.

Development/nmos/connection_api.cpp
```diff
@@ -304,7 +349,12 @@ namespace nmos
 
                 for (const auto& constraint : leg.as_object())
                 {
-                    if (type_auto_constraints.end() != type_auto_constraints.find(constraint.first))
+                    if (nmos::transports::mxl == transport_base)
```
garethsb: I don't understand yet why this is necessary. For everything except auto we can rely on the embedded JSON Schema can't we?

Development/nmos-cpp-node/node_implementation.cpp
```diff
@@ -1851,6 +2019,13 @@ nmos::transport_file_parser make_node_implementation_transport_file_parser()
     // (if this callback is specified, an 'empty' std::function is not allowed)
     return [](const nmos::resource& receiver, const nmos::resource& connection_receiver, const utility::string_t& transport_file_type, const utility::string_t& transport_file_data, slog::base_gate& gate)
     {
+        // BCP-007-03: MXL receivers do not use transport_file (non-null data is rejected here when parsed)
+        const nmos::transport transport_subclassification(nmos::fields::transport(receiver.data));
```
garethsb:
```suggestion
        const nmos::transport transport_subclassification{ nmos::fields::transport(receiver.data) };
```

Development/nmos/connection_api.cpp
garethsb: Hey, @jonathan-r-thorpe, OK,  I see the challenge. You should not need to add the UUID pattern, that's taken care of by validation against the `‎{sender,receiver}_transport_params_mxl.json‎` schemas, but you do need to allow `null` as well as `"auto"` in addition to the schema produced from the `/constraints` endpoint. That shouldn't be MXL specific, that should be an extension to the existing auto constraints functions above to also identify if the parameter supports `null`.

Development/nmos/connection_api.cpp
```diff
@@ -483,6 +580,8 @@ namespace nmos
                     throw std::logic_error("matching IS-04 and IS-05 resources not found");
                 }
 
+                const nmos::transport transport_subclassification(nmos::fields::transport(matching_resource->data));
+
```
garethsb: This change could now be reverted so that `transport_subclassification` is defined just before `details::validate_staged_constraints(...)` again?

Development/nmos/node_resources.cpp
```diff
@@ -415,7 +421,6 @@ namespace nmos
     }
 
     // See https://specs.amwa.tv/is-04/releases/v1.2.0/APIs/schemas/with-refs/flow_audio_coded.html
-    // (media_type must *not* be nmos::media_types::audio_L(bit_depth); cf. nmos::make_raw_audio_flow)
```
garethsb, line 418: That comment shouldn't have gone, I think? It didn't apply to the above new overload of `make_raw_audio_flow` but it does apply to coded audio flows.

Development/nmos/node_resources.h
```diff
@@ -84,6 +84,8 @@ namespace nmos
 
     // See https://specs.amwa.tv/is-04/releases/v1.2.0/APIs/schemas/with-refs/flow_audio_raw.html
     nmos::resource make_raw_audio_flow(const nmos::id& id, const nmos::id& source_id, const nmos::id& device_id, const nmos::rational& sample_rate, unsigned int bit_depth, const nmos::settings& settings);
+    // (media_type must *not* be nmos::media_types::audio_L(bit_depth); cf. the bit_depth overload)
```
garethsb: This comment should be deleted.

Development/nmos-cpp-node/node_implementation.cpp
```diff
@@ -38,6 +38,7 @@
 #include "nmos/lldp_manager.h"
 #endif
 #include "nmos/media_type.h"
+#include "nmos/mxl.h"
```
garethsb: ⬇️ include order? 😅

Development/nmos/connection_api.cpp
```diff
@@ -261,16 +354,20 @@ namespace nmos
...
             {
                 auto properties = value::object();
 
                 for (const auto& constraint : leg.as_object())
                 {
-                    if (type_auto_constraints.end() != type_auto_constraints.find(constraint.first))
+                    const bool auto_value = type_auto_constraints.end() != type_auto_constraints.find(constraint.first);
+                    const bool null_value = type_null_constraints.end() != type_null_constraints.find(constraint.first);
+
+                    if (auto_value || null_value)
```
garethsb: Given `make_staged_param_schema` has an efficient early return, you could probably just call it here and avoid the if/else?

garethsb (approved):
Looking really good now. Two tiny review comments.

Development/nmos-cpp-node/node_implementation.cpp
```diff
@@ -400,13 +416,19 @@ void node_implementation_init(nmos::node_model& model, nmos::experimental::contr
     const auto seed_id = nmos::experimental::fields::seed_id(model.settings);
     const auto node_id = impl::make_id(seed_id, nmos::types::node);
     const auto device_id = impl::make_id(seed_id, nmos::types::device);
+    // If mxl_domain_id is not set, generate a random UUID for the MXL domain
+    const auto mxl_domain_id = impl::fields::mxl_domain_id(model.settings).empty()
+        ? nmos::make_repeatable_id(seed_id, U("/x-nmos/mxl/domain"))
+        : impl::fields::mxl_domain_id(model.settings);
```
lo-simon, line 425: As `mxl_domain_id` can be overridden by the user, shouldn't we validate it against the UUID pattern? Maybe it's better to validate it in `nmos::make_connection_mxl_sender` and `nmos::make_connection_mxl_receiver`, since those functions can be called directly by the user?

Development/nmos/connection_resources.cpp
```diff
@@ -612,4 +612,108 @@ namespace nmos
...
+            using web::json::value_of;
+
+            return value_of({
+                { nmos::fields::mxl_domain_id, U("auto") },
+                { nmos::fields::mxl_flow_id, value::null() }
+            });
+        }
+    }
+
+    nmos::resource make_connection_mxl_sender(const nmos::id& id, const nmos::id& mxl_domain_id, const nmos::id& mxl_flow_id)
```
lo-simon, line 677: As mxl_domain_id can be overridden by the user, shouldn't we validate it against the UUID pattern? Maybe it's better to validate it in nmos::make_connection_mxl_sender and nmos::make_connection_mxl_receiver, since those functions can be called directly by the user?

Development/nmos/connection_resources.cpp
```diff
@@ -612,4 +612,108 @@ namespace nmos
...
+
+        data[nmos::fields::endpoint_active] = data[nmos::fields::endpoint_staged];
+        // The caller must resolve all instances of "auto" in the /active endpoint into the actual values that will be used!
+
+        // Note that the transporttype endpoint is implemented in terms of the matching IS-04 sender
+
+        return{ is05_versions::v1_2, types::sender, std::move(data), false };
+    }
+
+    nmos::resource make_connection_mxl_receiver(const nmos::id& id, const nmos::id& mxl_domain_id)
```
lo-simon, line 699: As mxl_domain_id can be overridden by the user, shouldn't we validate it against the UUID pattern? Maybe it's better to validate it in nmos::make_connection_mxl_sender and nmos::make_connection_mxl_receiver, since those functions can be called directly by the user?

lo-simon (commented):
Shouldn't we validate `mxl_domian_id` against the UUID pattern? I have marked it in 3 places. The rest of the PR looks good to me.

Development/nmos-cpp-node/node_implementation.cpp
garethsb: I don't think we usually validate stuff that's come from the `config.json` file / `nmos::settings`? if we want to, I suggest doing that as a one-shot `validate_settings(const nmos::settings& settings)` function declared in `nmos/settings.h` and then if applications like nmos-cpp-node define their own application-specific settings, they have their own `validate_settings` function that calls the above one.

lo-simon: I am pointing this out because we could end up with an invalid constraint in the sender and receiver constraints endpoints. I think we should validate it in `nmos::make_connection_mxl_sender` and `nmos::make_connection_mxl_receiver` rather than in the settings.

/receivers/receiver_id/constraints
`[
  {
    "mxl_domain_id": {
      "enum": [
        "bad_mxl_flow_id"
      ]
    },
    "mxl_flow_id": {}
  }
]`

/senders/sender_id/constraints
`[
  {
    "mxl_domain_id": {
      "enum": [
        "bad_mxl_flow_id"
      ]
    },
    "mxl_flow_id": {
      "enum": [
        "43bb8367-fc6d-5402-b9e3-77202edf996d"
      ]
    }
  }
]`

garethsb: I get your point that it's "user input" but actually it's only user input from the point of view of the example application, not the library, so I think it's better provided as an opt-in validation function so that you don't pay for what you don't need (e.g. you've done your app level validation of user level input and don't want nmos-cpp to validate it again, or you defined the value programmatically so you _know_ it doesn't need further validation).

From the point of view of the `nmos::make_connection_mxl_sender` function, it's just another `nmos::id` parameter. And if we need to validate that, we should validate `nmos::id` parameters in many functions (after all, it's just a typedef for a `utility::string_t`...).

**Update:** I've had a go at providing settings validation: https://github.com/jonathan-r-thorpe/nmos-cpp/pull/2

lo-simon (approved):
LGTM

PR #484, opened 2026-03-17
Implement TR-10-9 Section 15 DNS-SD browse strategy

simonbaren:
## Summary

Implements the DNS-SD browse strategy required by TR-10-9 Section 15 for IPMX device service discovery.

### Changes

- **New `dns_sd_browse_mode` setting** (`"both"` / `"unicast"` / `"mdns"`) to control DNS-SD browse method
  - `"both"` (default): try unicast DNS first with half the timeout budget; fall back to mDNS if no services found
  - `"unicast"`: unicast DNS only
  - `"mdns"`: mDNS only
- **Domain class filtering** in browse results to prevent mDNS results leaking into unicast DNS queries and vice versa
- Refactored `resolve_service` to delegate to `resolve_service_` with the dual-discovery logic

### Motivation

TR-10-9 Section 15 requires IPMX devices to support both mDNS and unicast DNS for DNS-SD browse operations, defaulting to using both methods. When using both, unicast DNS must be attempted first, with mDNS as a fallback only if unicast is unsuccessful.

### Files changed

- `Development/nmos/settings.h` — new `dns_sd_browse_mode` field
- `Development/nmos/mdns.cpp` — dual-browse logic and domain class filtering

https://static.vsf.tv/download/technical_recommendations/VSF_TR-10-9_v2_2025-05-13.pdf


lo-simon:
Thank you, @simonbaren, for your PR, we will look into it shortly.

simonbaren:
Hi @lo-simon, kindly reminder.

lo-simon:
Hi @simonbaren 

So sorry for the late reply.

I am happy with your PR. There are a few points I would like to address and offer suggestions on.

1. For `dns_sd_browse_mode`, it would be better to use an enum rather than a string, following the style of [href_mode](https://github.com/sony/nmos-cpp/blob/8e2e17f572feb932e268116e41f1e0e24b56f580/Development/nmos/settings.cpp#L122-L128).
2. Add `dns_sd_browse_mode` to the example nmos-cpp-node/config.json.
3. Mark the unused `resolve_service` function (line 711 in mdns.cpp) as deprecated in both the cpp and header files by adding a comment.
4. Update the `resolve_service` and `resolve_service_` comments in the header and in the cpp.
5. I think it would be better to keep the same `discovery timeout` for both `primary` and `fallback` resolves, rather than halving it. What do you think, @jonathan-r-thorpe and @garethsb?
6. Shall we add the following table to settings.h and config.json as a reminder?

dns_sd_browse_mode | domain | expected-resolve-behaviour 
--- | --- | --- 
both | example.com | unicast->mdns 
both | local. | mdns 
unicast | example.com | unicast
unicast | local. | mdns 
mdns | example.com | mdns 
mdns | local. | mdns

simonbaren:
Hi @lo-simon, thanks for the review. Pushed a commit addressing items 1–4 and 6:

1. **`dns_sd_browse_mode` as enum** — used the `string_enum` pattern (`DEFINE_STRING_ENUM`) rather than the integer `href_mode` style. Rationale:  The string-based pattern is consistent with most other JSON-string enums in nmos-cpp (`interlace_mode`, `activation_mode`, `event_type`, etc.) and keeps the config human-readable. Happy to switch to int-enum if you'd prefer.
2. **`config.json`** — added a commented-out `dns_sd_browse_mode` entry next to `domain`, with the behaviour table inline.
3. **Deprecated marker** — added `// DEPRECATED: …` comments above the unused `web::uri`-returning `resolve_service` overload (and its inline template wrapper) in both `mdns.h` and `mdns.cpp`.
4. **Refreshed comments** — `resolve_service` and `resolve_service_` doc comments in both `mdns.h` and `mdns.cpp` now distinguish the `web::uri` vs `resolved_service` return shapes and reference the TR-10-9 mode selection.
5. **Timeout halving** — left as-is for now, awaiting input from @jonathan-r-thorpe / @garethsb.
6. **Behaviour table** — added next to the `dns_sd_browse_mode` field declaration in `settings.h` and next to the entry in `config.json`.

Let me know if anything needs adjusting.

jonathan-r-thorpe:
@lo-simon @simonbaren on 5. I agree with @lo-simon that I would expect the primary service  timeout to remain unchanged, even when there is a fallback service. My expectation would therefore be that the timeout applied per service (i.e. don't halve it).

simonbaren:
Thanks @jonathan-r-thorpe. Done in the latest commit - primary and fallback now both use the full `discovery_backoff_max` timeout (no more halving).

Development/nmos-cpp-node/config.json
```diff
@@ -85,6 +85,21 @@
     // domain [registry, node]: the domain on which to browse for services or an empty string to use the default domain (specify "local." to explictly select mDNS)
     //"domain": "",
 
+    // dns_sd_browse_mode [node]: DNS-SD browse method per TR-10-9 Section 15
+    //   "both" (default) = unicast DNS first, mDNS fallback if unsuccessful
+    //   "unicast"        = unicast DNS only
+    //   "mdns"           = mDNS only
```
lo-simon:
```suggestion
    //   both(0) (default) = unicast DNS first, mDNS fallback if unsuccessful
    //   unicast(1)        = unicast DNS only
    //   mdns(2)           = mDNS only
```

Development/nmos-cpp-node/config.json
```diff
@@ -85,6 +85,21 @@
...
+    // Expected resolve behaviour for each (mode, domain) combination:
+    //   mode    | domain      | behaviour
+    //   --------+-------------+--------------------
+    //   both    | example.com | unicast -> mdns
+    //   both    | local.      | mdns
+    //   unicast | example.com | unicast
+    //   unicast | local.      | mdns
+    //   mdns    | example.com | mdns
+    //   mdns    | local.      | mdns
+    //"dns_sd_browse_mode": "both",
```
lo-simon:
```suggestion
    //"dns_sd_browse_mode": 0,
```

Development/nmos/dns_sd_browse_mode.h
```diff
@@ -0,0 +1,19 @@
+#ifndef NMOS_DNS_SD_BROWSE_MODE_H
+#define NMOS_DNS_SD_BROWSE_MODE_H
+
+#include "nmos/string_enum.h"
```
lo-simon:
```suggestion
```

Development/nmos/dns_sd_browse_mode.h
```diff
@@ -0,0 +1,19 @@
...
+{
+    // DNS-SD browse method per TR-10-9 Section 15
+    // Selected via the dns_sd_browse_mode setting in nmos::fields
+    DEFINE_STRING_ENUM(dns_sd_browse_mode)
+    namespace dns_sd_browse_modes
+    {
+        const dns_sd_browse_mode both{ U("both") };       // unicast DNS first, mDNS fallback if unsuccessful
+        const dns_sd_browse_mode unicast{ U("unicast") }; // unicast DNS only
+        const dns_sd_browse_mode mdns{ U("mdns") };       // mDNS only
+    }
```
lo-simon:
```suggestion
    enum dns_sd_browse_mode
    {
        both = 0,    // unicast DNS first, mDNS fallback if unsuccessful
        unicast = 1, // unicast DNS only
        mdns = 2     // mDNS only
    };
```

Development/nmos/mdns.cpp
```diff
@@ -710,28 +724,33 @@ namespace nmos
...
 
-        // helper function for resolving instances of the specified service (API) based on the specified settings
-        // with the highest version, highest priority instances at the front, and services with the same priority ordered randomly
+        // Helper function for resolving instances of the specified service (API) based on the specified settings,
+        // returning ((api_version, priority), uri) tuples. Highest version and highest priority first, services
+        // with the same priority ordered randomly.
+        // The browse method is selected by the dns_sd_browse_mode setting per TR-10-9 Section 15:
+        //  - "both" (default): unicast DNS first, mDNS fallback if unsuccessful
+        //  - "unicast"       : unicast DNS only
+        //  - "mdns"          : mDNS only
```
lo-simon:
```suggestion
        //  - both (default): unicast DNS first, mDNS fallback if unsuccessful
        //  - unicast       : unicast DNS only
        //  - mdns          : mDNS only
```

Development/nmos/mdns.h
```diff
@@ -157,39 +157,50 @@ namespace nmos
...
 
-        // helper function for resolving instances of the specified service (API) based on the specified settings
-        // with the highest version, highest priority instances at the front, and services with the same priority ordered randomly
+        // Helper function for resolving instances of the specified service (API) based on the specified settings,
+        // returning ((api_version, priority), uri) tuples. Highest version and highest priority first, services
+        // with the same priority ordered randomly.
+        // The browse method is selected by the dns_sd_browse_mode setting per TR-10-9 Section 15:
+        //  - "both" (default): unicast DNS first, mDNS fallback if unsuccessful
+        //  - "unicast"       : unicast DNS only
+        //  - "mdns"          : mDNS only
```
lo-simon:
```suggestion
        //  - both (default): unicast DNS first, mDNS fallback if unsuccessful
        //  - unicast       : unicast DNS only
        //  - mdns          : mDNS only
```

Development/nmos/settings.h
```diff
@@ -82,6 +82,22 @@ namespace nmos
         // domain [registry, node]: the domain on which to browse for services or an empty string to use the default domain (specify "local." to explictly select mDNS)
         const web::json::field_as_string_or domain{ U("domain"), U("") };
 
+        // dns_sd_browse_mode [node]: DNS-SD browse method per TR-10-9 Section 15
+        // (see nmos::dns_sd_browse_modes for valid values)
+        //   "both" (default) = unicast DNS first, mDNS fallback if unsuccessful
+        //   "unicast"        = unicast DNS only
+        //   "mdns"           = mDNS only
```
lo-simon:
```suggestion
        //   both(0) (default) = unicast DNS first, mDNS fallback if unsuccessful
        //   unicast(1)        = unicast DNS only
        //   mdns(2)           = mDNS only
```

Development/nmos/settings.h
```diff
@@ -82,6 +82,22 @@ namespace nmos
...
+        // Expected resolve behaviour for each (mode, domain) combination:
+        //   mode    | domain      | behaviour
+        //   --------+-------------+--------------------
+        //   both    | example.com | unicast -> mdns
+        //   both    | local.      | mdns
+        //   unicast | example.com | unicast
+        //   unicast | local.      | mdns
+        //   mdns    | example.com | mdns
+        //   mdns    | local.      | mdns
+        const web::json::field_as_string_or dns_sd_browse_mode{ U("dns_sd_browse_mode"), U("both") };
```
lo-simon:
```suggestion
        const web::json::field_as_integer_or dns_sd_browse_mode{ U("dns_sd_browse_mode"), 0 };
```

lo-simon: Use an enum rather than a string for the `dns_sd_browse_mode` to reduce the risk of typos.

lo-simon (changes requested):
Use an enum rather than a string for the `dns_sd_browse_mode` to reduce the risk of typos.

simonbaren:
Hi @lo-simon, thanks for the detailed review. Switched to a typed integer enum.
Two small deviations from your literal suggestions, both for consistency with the existing `href_mode` pattern you originally cited — happy to revert either if you'd prefer:

1. **Enumerator naming** — used `dns_sd_browse_mode_both` / `_unicast` / `_mdns` (prefixed) rather than bare `both` / `unicast` / `mdns`. Bare names would land directly in the `nmos::` namespace, where `nmos::both` etc. are quite generic. The prefixed form matches `href_mode`'s `href_mode_default` / `href_mode_name` / etc.

2. **Header location** — moved the enum into `mdns.cpp` (file-local, under `namespace nmos`) and deleted `Development/nmos/dns_sd_browse_mode.h` that was added in previous commit. This mirrors how `href_mode` lives entirely in `settings.cpp` rather than in a public header. The setting field stays in `settings.h`; the integer value descriptions are inlined in the comment there (per your suggestion).

All other inline suggestions applied verbatim:
- `field_as_integer_or dns_sd_browse_mode{ U("dns_sd_browse_mode"), 0 }`
- `//"dns_sd_browse_mode": 0,` in the example config
- `both(0)` / `unicast(1)` / `mdns(2)` in the settings.h and config.json comment tables
- doc comments in `mdns.h`/`mdns.cpp` use bare `both`/`unicast`/`mdns`

PR #485, opened 2026-03-25
Fix service name generation to enforce RFC 6763 length limit

garethsb:
Truncate mDNS service instance names exceeding 63 bytes with a hash suffix to ensure RFC 6763 compliance, since Avahi does not automatically truncate like mDNSResponder. Also use underscore rather than colon as the host/port separator, and declare service_name in the header. Add unit tests for service_name.

Made-with: Cursor

### DIFF
--- a/Development/nmos/mdns.cpp
+++ b/Development/nmos/mdns.cpp
@@ -1,7 +1,6 @@
 #include "nmos/mdns.h"
 
 #include <functional>
-#include <iomanip>
 #include <boost/algorithm/string/erase.hpp>
 #include <boost/algorithm/string/join.hpp>
 #include <boost/algorithm/string/predicate.hpp>
@@ -332,11 +331,17 @@ namespace nmos
                 return utility::us2s(nmos::fields::service_name_prefix(settings)) + "_" + service_api(service);
             }
 
-            inline std::string hash_string(const std::string& s)
+            // generate a hash string (slightly more compact and possibly slightly more memorable than hex)
+            inline std::string hash_string(const std::string& s, size_t len = 5)
             {
-                std::ostringstream os;
-                os << std::hex << std::setfill('0') << std::setw(8) << (std::hash<std::string>{}(s) & 0xFFFFFFFF);
-                return os.str();
+                auto hash = std::hash<std::string>{}(s);
+                // vowels (and vowel-ish digits) are omitted from the set of available characters
+                // to reduce the chances of "bad words" being formed
+                static const char alphanums[] = "bcdfghjklmnpqrstvwxz2456789";
+                static const size_t base = sizeof(alphanums) - 1;
+                std::string result(len, ' ');
+                for (auto& c : result) { c = alphanums[hash % base]; hash /= base; }
+                return result;
             }
 
             inline std::set<nmos::api_version> service_versions(const nmos::service_type& service, const nmos::settings& settings)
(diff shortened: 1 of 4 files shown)


garethsb:
@lo-simon, @jonathan-r-thorpe, looks like the AMWA test suite is taking 40+ minutes to run, is that expected?

Development/nmos/mdns.cpp
```diff
@@ -1,5 +1,6 @@
 #include "nmos/mdns.h"
 
+#include <functional>
```
lo-simon:
```suggestion
```

lo-simon: Don't think it is required.

Development/nmos/test/mdns_test.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+    // a name that is 64 bytes should be truncated
+    const std::string host_45(45, 'x');
+    const nmos::settings settings = value_of({
+        { nmos::fields::host_name, utility::s2us(host_45) },
+        { nmos::experimental::fields::href_mode, 1 }
+    });
+
+    const auto name = nmos::experimental::service_name(nmos::service_types::node, settings);
+    BST_REQUIRE(name.size() <= 63);
```
lo-simon, line 171:
```suggestion
    BST_CHECK_EQUAL(63, name.size());
```

lo-simon: Would it be better to use the expected returning value, as the service name will always return a length of 63 if it is longer than 63?

lo-simon (changes requested):
Some minor suggestions.

Development/nmos/test/mdns_test.cpp
garethsb: It can be shorter than 63 characters because of this line, designed to avoid "--" or "_-" before the suffix:

```C++
            const auto prefix_length = name.find_last_not_of("-_", max_prefix_length - 1) + 1;
```

garethsb:
@lo-simon any idea why two of the Windows 2022 runners failed fast?

lo-simon:
> @lo-simon any idea why two of the Windows 2022 runners failed fast?

I have seen that before; most of the time, when I re-run the runner, the problem disappears. 

According to Google, in a CI/CD environment, the runner might have multiple network interfaces active.

When Get-NetIPConfiguration runs:

1. It expects to find one primary configuration to output.
2. If it finds multiple configurations, it returns an array (System.Object[]). 
3. Internally, the command attempts to map these results to a property (likely NetAdapter) that is strictly defined to hold only one instance at a time.

To fix that, as we just need any valid IP config to proceed, pipe it to Select-Object -First 1 as shown below:

`Get-NetIPConfiguration | Select-Object -First 1`

garethsb:
> To fix that, as we just need any valid IP config to proceed, pipe it to Select-Object -First 1 as shown below:
> 
> `Get-NetIPConfiguration | Select-Object -First 1`

Hey, @lo-simon, we probably want to put the `Select-Object -First 1` after the `Where-Object`?

lo-simon:
> > To fix that, as we just need any valid IP config to proceed, pipe it to Select-Object -First 1 as shown below:
> > `Get-NetIPConfiguration | Select-Object -First 1`
> 
> Hey, @lo-simon, we probably want to put the `Select-Object -First 1` after the `Where-Object`?

Nice, @garethsb, let's open another PR for this fix.

PR #486, opened 2026-04-02
Enhance Monitor Status callback to include monitor information

lo-simon:
Currently, the callback handlers exposed by nmos-cpp return information for all Interfaces, regardless of whether they have an associated Status Monitor. It is making it impossible to display the Interface information associated with a specific Monitor.

**Current callback**
```
typedef std::function<std::vector<nc::counter>(void)> get_packet_counters_handler;
typedef std::function<void(void)> reset_monitor_handler;
```
**Updated callback**
```
typedef std::function<std::vector<nc::counter>(const nmos::resource&)> get_packet_counters_handler;
typedef std::function<void(const nmos::resource&)> reset_monitor_handler;
```

### DIFF
--- a/Development/nmos-cpp-node/node_implementation.cpp
+++ b/Development/nmos-cpp-node/node_implementation.cpp
@@ -2121,11 +2121,13 @@ nmos::control_protocol_property_changed_handler make_node_implementation_control
     };
 }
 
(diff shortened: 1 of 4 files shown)


jonathan-r-thorpe (approved):
LGTM

PR #488, opened 2026-05-01
Add Network Time Protocol (NTP) support

lo-simon:
The following scenarios illustrate the issue that arises when the system clock is rewound. In both cases, the time used is `system_clock`. To overcome this issue, we will need to use `steady_clock` to obtain the request's received time at the RDS. On the Node side, we will need to use `nmos::strictly_increasing_update` for the resource's version, which will always increment.

NTP study - Rewind clock on Node

NTP study - Rewind clock on RDS


Development/nmos/api_utils.cpp
```diff
@@ -503,8 +503,13 @@ namespace nmos
                 nmos::api_gate gate(gate_, req, parameters);
 
                 const auto received_time = req.headers().find(details::received_time);
+
                 const auto processing_dur = req.headers().end() != received_time
-                    ? bst::chrono::duration_cast<bst::chrono::microseconds>(nmos::tai_clock::now() - nmos::time_point_from_tai(nmos::parse_version(received_time->second))).count() / 1000.0
+                    ? bst::chrono::duration_cast<bst::chrono::microseconds>(
+                        // processing_dur = now - received time
+                        ( tai_clock::time_point(tai_clock::duration(std::chrono::duration_cast<std::chrono::nanoseconds>(std::chrono::steady_clock::now().time_since_epoch()).count()))
```
garethsb, line 510: Can you explain this? Doesn't the `steady_clock` have a different epoch than `tai_clock`?

lo-simon: `tai_clock` now uses the boost implementation. According to the note in https://github.com/sony/nmos-cpp/blob/master/Development/nmos/api_utils.cpp#L710-L711, on Windows, the boost version of time_point::now (system_clock::now) sometimes returns the same value after a small time increment. Therefore, we are using std::chrono::steady_clock::now() here.

PR #490, opened 2026-05-21
Fix Windows CI host IP discovery on windows-2022 runners.

jonathan-r-thorpe:
Get-NetIPConfiguration fails when multiple adapters map to one interface; use the default route and Get-NetIPAddress instead, and only disable vEthernet (nat) when present.


lo-simon (approved):
LGTM

PR #493, opened 2026-06-01
Validate HSTS max-age parsing

### LINKED ISSUE #492: HSTS invalid max-age validation
parse_hsts_header("max-age=meow") currently parses as max_age = 0. RFC 6797 defines max-age as numeric delta-seconds. Would a PR rejecting non-numeric values with a unit test be acceptable?

kshitijshresth:
Reject invalid Strict-Transport-Security max-age values instead of treating them as zero.

This adds validation for non-numeric and out-of-range max-age values and allows the existing unit test coverage for invalid max age input

Fixes #492

### DIFF
--- a/Development/cpprest/http_utils.cpp
+++ b/Development/cpprest/http_utils.cpp
@@ -2,6 +2,8 @@
 
 #include <algorithm>
 #include <cctype>
+#include <cstdint>
+#include <limits>
 #include <map>
 #include <set>
 #include <boost/algorithm/string/predicate.hpp>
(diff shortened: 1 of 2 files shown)


garethsb:
Is an exception the right approach here? The relevant spec would be [RFC 6797: HTTP Strict Transport Security (HSTS)](https://www.rfc-editor.org/info/rfc6797/#section-6.1:~:text=4.%20%20UAs%20MUST,the%20recognized%20directives.) which says among other things:

```
   4.  UAs MUST ignore any STS header field containing directives, or
       other header field value data, that does not conform to the
       syntax defined in this specification.

   5.  If an STS header field contains directive(s) not recognized by
       the UA, the UA MUST ignore the unrecognized directives, and if
       the STS header field otherwise satisfies the above requirements
       (1 through 4), the UA MUST process the recognized directives.
```

kshitijshresth:
> Is an exception the right approach here? The relevant spec would be [RFC 6797: HTTP Strict Transport Security (HSTS)](https://www.rfc-editor.org/info/rfc6797/#section-6.1:~:text=4.%20%20UAs%20MUST,the%20recognized%20directives.) which says among other things:
> 
> ```
>    4.  UAs MUST ignore any STS header field containing directives, or
>        other header field value data, that does not conform to the
>        syntax defined in this specification.
> 
>    5.  If an STS header field contains directive(s) not recognized by
>        the UA, the UA MUST ignore the unrecognized directives, and if
>        the STS header field otherwise satisfies the above requirements
>        (1 through 4), the UA MUST process the recognized directives.
> ```

Thanks for the review! You're right that RFC prescribes that a UA ignore a non-conforming STS header field rather than rejecting it.

I leaned on std::invalid_argument here for consistency parse_hsts_header already throws std::invalid_argument for empty max-age an unterminated quoted-string value and a missing required max-age so this change extends that existing contract to a non-numeric/out-of-range max-age. The caller can then ignore a thrown exception.

I also wanted to avoid the previous behaviour of silently parsing an invalid value as 0 since per 8.1 max-age=0 is an instruction to delete cached policy

if you'd prefer the parser to model "ignore" more directly I'd change parse_hsts_header to return bst::optional<hsts> (nullopt = ignore the header) instead of throwing? It might still be an improvement over deleting the cached policy

PR #494, opened 2026-06-02
Replace fatal log with severe log

lo-simon:
Thanks to @garethsb for pointing out that there are a couple of fatal logs that can be replaced with the severe log. Even when those errors occur, we still want the application to continue running, not terminate.


garethsb (approved):
LGTM

PR #495, opened 2026-06-05
Use `rtp_sender_ports` list to create the relevant sender controls

lo-simon:
Fix the sender control example. It should use `rtp_sender_ports` to iterate over when creating the sender controls. This will allow a mixture of "senders": ["v", "a"] and "receivers": [] configurations.


jonathan-r-thorpe (approved):
LGTM

Issue #38, opened 2018-12-11
Support SDP creation/processing for remaining modes of RTP operation specified by IS-05


garethsb-sony:
IS-05 documentation on [Interpretation of SDP Files](https://github.com/AMWA-TV/nmos-device-connection-management/blob/v1.0.1/docs/4.1.%20Behaviour%20-%20RTP%20Transport%20Type.md#interpretation-of-sdp-files) describes 7 modes of operation. A few of these are not yet implemented by **nmos-cpp**. If you are interested in FEC, SMPTE 2022-7 Temporal Redundancy, or RTCP, help would be welcome.

- [x] [Unicast](https://github.com/AMWA-TV/nmos-device-connection-management/blob/v1.0.1/docs/4.1.%20Behaviour%20-%20RTP%20Transport%20Type.md#unicast)
- [x] [Source Specific Multicast](https://github.com/AMWA-TV/nmos-device-connection-management/blob/v1.0.1/docs/4.1.%20Behaviour%20-%20RTP%20Transport%20Type.md#source-specific-multicast)
- [ ] [Operation with SMPTE 2022-5](https://github.com/AMWA-TV/nmos-device-connection-management/blob/v1.0.1/docs/4.1.%20Behaviour%20-%20RTP%20Transport%20Type.md#operation-with-smpte-2022-5) (FEC)
- [x] [Operation with SMPTE 2022-7 - Separate Source Addresses](https://github.com/AMWA-TV/nmos-device-connection-management/blob/v1.0.1/docs/4.1.%20Behaviour%20-%20RTP%20Transport%20Type.md#separate-source-addresses)
- [x] [Operation with SMPTE 2022-7 - Separate Destination Addresses](https://github.com/AMWA-TV/nmos-device-connection-management/blob/v1.0.1/docs/4.1.%20Behaviour%20-%20RTP%20Transport%20Type.md#separate-destination-addresses)
- [ ] [Operation with SMPTE 2022-7 - Temporal Redundancy](https://github.com/AMWA-TV/nmos-device-connection-management/blob/v1.0.1/docs/4.1.%20Behaviour%20-%20RTP%20Transport%20Type.md#temporal-redundancy)
- [ ] [Operation with RTCP](https://github.com/AMWA-TV/nmos-device-connection-management/blob/v1.0.1/docs/4.1.%20Behaviour%20-%20RTP%20Transport%20Type.md#operation-with-rtcp)

garethsb:
[...]

PR #501, opened 2026-07-14
Add RTP parameter sets and RTCP SDP support

thangldw:
## Summary

- refactor RTP Connection API resource creation into composable core, multicast, FEC, and RTCP parameter sets
- preserve the existing sender and receiver factory overloads and their historical defaults
- add the missing FEC and RTCP JSON field accessors
- parse and generate the RFC 3605 `a=rtcp` SDP attribute, including the port-only form
- add parameter-set, `auto` resolution, specification example, malformed input, and round-trip tests

## Scope

This is the first reviewable slice of issue #38. It provides the Connection API parameter sets and completes RTCP SDP handling. SMPTE 2022-5 FEC SDP and SMPTE 2022-7 temporal redundancy remain follow-up work, so this PR intentionally uses `Refs #38` rather than closing the issue.

## Testing

- Debug build of `nmos-cpp` and `nmos-cpp-test`
- parameter-set tests: 41 assertions in 2 test cases
- RTCP tests: 7 assertions in 2 test cases
- full suite: 2060 assertions in 163 test cases

Refs #38

### DIFF
--- a/Development/nmos/sdp_utils.cpp
+++ b/Development/nmos/sdp_utils.cpp
@@ -631,6 +631,29 @@ namespace nmos
                 );
             }
 
+            if (nmos::fields::rtcp_enabled(transport_param))
+            {
+                // a=rtcp:<port> [<nettype> <addrtype> <connection-address>]
+                // See https://tools.ietf.org/html/rfc3605
+                auto rtcp = value_of({
+                    { sdp::fields::port, transport_param.at(nmos::fields::rtcp_destination_port) }
+                }, keep_order);
+                const auto& rtcp_destination_ip = nmos::fields::rtcp_destination_ip(transport_param);
+                if (!rtcp_destination_ip.is_null())
+                {
+                    const auto& rtcp_address = rtcp_destination_ip.as_string();
+                    rtcp[sdp::fields::network_type] = value::string(sdp::network_types::internet.name);
+                    rtcp[sdp::fields::address_type] = value::string(details::get_address_type_multicast(rtcp_address).first.name);
+                    rtcp[sdp::fields::unicast_address] = value::string(rtcp_address);
+                }
+                web::json::push_back(
+                    media_attributes, value_of({
+                        { sdp::fields::name, sdp::attributes::rtcp },
+                        { sdp::fields::value, std::move(rtcp) }
+                    }, keep_order)
+                );
+            }
+
             if (0 != sdp_params.packet_time)
             {
                 // a=ptime:<packet time>
@@ -954,11 +977,11 @@ namespace nmos
         // * Any Source Multicast
         // * Operation with SMPTE 2022-7 - Separate Source Addresses
         // * Operation with SMPTE 2022-7 - Separate Destination Addresses
+        // * Operation with RTCP
 
         // The following cases are not yet handled:
         // * Operation with SMPTE 2022-5
         // * Operation with SMPTE 2022-7 - Temporal Redundancy
-        // * Operation with RTCP
 
         auto& media_descriptions = sdp::fields::media_descriptions(session_description);
 
@@ -1045,6 +1068,23 @@ namespace nmos
                 }
 
                 params[nmos::fields::destination_port] = value::number(sdp::fields::port(media));
+
+                if (!media_attributes.is_null())
+                {
+                    auto& ma = media_attributes.as_array();
+                    auto rtcp = sdp::find_name(ma, sdp::attributes::rtcp);
+                    if (ma.end() != rtcp)
+                    {
+                        const auto& rtcp_value = sdp::fields::value(*rtcp);
+                        params[nmos::fields::rtcp_enabled] = value::boolean(true);
+                        params[nmos::fields::rtcp_destination_port] = value::number(sdp::fields::port(rtcp_value));
+                        params[nmos::fields::rtcp_destination_ip] = rtcp_value.has_field(sdp::fields::unicast_address)
+                            ? rtcp_value.at(sdp::fields::unicast_address)
+                            : !params[nmos::fields::multicast_ip].is_null()
+                                ? params[nmos::fields::multicast_ip]
+                                : params[nmos::fields::interface_ip];
+                    }
+                }
 
                 params[nmos::fields::rtp_enabled] = value::boolean(true);
 
--- a/Development/nmos/sdp_utils.cpp
+++ b/Development/nmos/sdp_utils.cpp
@@ -1078,11 +1078,12 @@ namespace nmos
                         const auto& rtcp_value = sdp::fields::value(*rtcp);
                         params[nmos::fields::rtcp_enabled] = value::boolean(true);
                         params[nmos::fields::rtcp_destination_port] = value::number(sdp::fields::port(rtcp_value));
-                        params[nmos::fields::rtcp_destination_ip] = rtcp_value.has_field(sdp::fields::unicast_address)
+                        auto rtcp_destination_ip = rtcp_value.has_field(sdp::fields::unicast_address)
                             ? rtcp_value.at(sdp::fields::unicast_address)
                             : !params[nmos::fields::multicast_ip].is_null()
                                 ? params[nmos::fields::multicast_ip]
                                 : params[nmos::fields::interface_ip];
+                        params[nmos::fields::rtcp_destination_ip] = std::move(rtcp_destination_ip);
                     }
                 }
 
--- a/Development/nmos/test/connection_resources_test.cpp
+++ b/Development/nmos/test/connection_resources_test.cpp
@@ -1,0 +1,100 @@
+// The first "test" is of course whether the header compiles standalone
+#include "nmos/connection_resources.h"
+
+#include "bst/test/test.h"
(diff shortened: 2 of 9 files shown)


garethsb:
Thanks, the additions look pretty good to me. It's only been 7 years since I ran out of effort to finish this off. 😀
When the CI is passing, I'll try to do a full review.

thangldw:
Thanks for the feedback. I checked the current CI results: all Linux jobs pass, while all four Windows jobs fail consistently with a segfault in `testInterpretationOfSdpFilesRtcp`.

I will address this before any further feature work. My next update will:

- replace the RTCP whitespace/field parsing path with a simple deterministic token parser that does not depend on regex behavior across standard-library implementations;
- split the RTCP coverage into focused tests for the full RFC 3605 form, the port-only form, and malformed partial forms, so failures are localized;
- run the focused tests and the full local suite, then review the diff and API compatibility before pushing the fix to this PR.

I will wait for the Windows matrix to pass after that update before asking for the full review.

thangldw:
The Windows CI fix is now pushed in `278f2e8`.

The RTCP parser now tokenizes the RFC 3605 fields directly without the regex-based whitespace path, and the coverage is split into full-form, port-only, and malformed-form test cases. Locally, the focused RTCP/parameter-set tests pass, and the full suite passes with 2,062 assertions across 165 test cases. I also reviewed the final diff: it changes no public API or SDP output format, and the temporary Apple Silicon Catch compatibility changes used for local testing were fully reverted.

The new GitHub Actions run is currently marked `action_required`, so it appears to be waiting for repository approval before the Windows/Linux matrix can start.

garethsb:
Can you explain why, in the `rtcp` grammar, you can't just use `strings_converter` (as RFC 3605 uses RFC 2327 `space=%d32` single character) or follow a pattern more similar to the `source_filter` parse function just above with a quick check mirroring the generate function for whether there are more fields after the first one?

Other than that little question, this looks really good to me - nice design, backwards compatible, good test cases. Thank you!

thangldw:
You're right—there isn't a technical reason that `strings_converter` cannot be used here. I introduced the separate tokenization only to enforce the two valid cardinalities (one field or four fields) while accepting horizontal whitespace, but RFC 3605 inherits `space = %d32` from RFC 2327, so accepting broader whitespace was unnecessary and less consistent with the existing grammar.

I'll change the parse function to use `strings_converter.parse(s)`, keep the explicit size check for exactly 1 or 4 fields, and then map those fields into the structured object. That will mirror the generate function and follow the existing project pattern more closely. Thanks for pointing this out.

Development/nmos/sdp_utils.cpp
```diff
@@ -1046,6 +1069,23 @@ namespace nmos
...
+                    if (ma.end() != rtcp)
+                    {
+                        const auto& rtcp_value = sdp::fields::value(*rtcp);
+                        params[nmos::fields::rtcp_enabled] = value::boolean(true);
+                        params[nmos::fields::rtcp_destination_port] = value::number(sdp::fields::port(rtcp_value));
+                        params[nmos::fields::rtcp_destination_ip] = rtcp_value.has_field(sdp::fields::unicast_address)
+                            ? rtcp_value.at(sdp::fields::unicast_address)
+                            : !params[nmos::fields::multicast_ip].is_null()
+                                ? params[nmos::fields::multicast_ip]
+                                : params[nmos::fields::interface_ip];
```
garethsb: The Windows CI segfault was nothing to do with regex, it's this. The `params[...]` on the left-hand and the right-hand side of the assignment are references into the same json value and the order of operations causes a dangling reference.

Possible fix:
```suggestion
                        auto rtcp_destination_ip = rtcp_value.has_field(sdp::fields::unicast_address)
                            ? rtcp_value.at(sdp::fields::unicast_address)
                            : !params[nmos::fields::multicast_ip].is_null()
                                ? params[nmos::fields::multicast_ip]
                                : params[nmos::fields::interface_ip];
                        params[nmos::fields::rtcp_destination_ip] = std::move(rtcp_destination_ip);
```

Development/sdp/sdp_grammar.cpp
```diff
@@ -654,6 +654,42 @@ namespace sdp
...
+                            }
+                            return s;
+                        },
+                        [](const std::string& s) {
+                            std::istringstream is(s);
+                            std::vector<std::string> parts;
+                            for (std::string part; is >> part;)
+                            {
+                                parts.push_back(std::move(part));
+                            }
```
garethsb: Would still prefer to see this using `strings_converter`, I think?

garethsb (changes requested):
Just two small tweaks and I'll approve this. Thanks, @thangldw!

Development/sdp/sdp_grammar.cpp
thangldw: Updated in `bbafe1f` to use `strings_converter.parse(s)` while retaining the explicit one-or-four-field validation, as discussed. Thanks.

garethsb (approved):
LGTM

Development/nmos/test/connection_resources_test.cpp
```diff
@@ -0,0 +1,100 @@
...
+        params[nmos::fields::source_ip] = web::json::value::string(U("192.0.2.10"));
+        params[nmos::fields::destination_ip] = web::json::value::string(U("232.21.21.133"));
+
+        nmos::resolve_rtp_auto(nmos::types::sender, transport_params, 5000);
+
+        BST_REQUIRE_EQUAL(U("232.21.21.133"), nmos::fields::rtcp_destination_ip(params).as_string());
+        BST_REQUIRE_EQUAL(5001, nmos::fields::rtcp_destination_port(params).as_integer());
+        BST_REQUIRE_EQUAL(5001, nmos::fields::rtcp_source_port(params).as_integer());
+    }
+
```
lo-simon, line 87:
```suggestion

    {
        const nmos::rtp_sender_parameter_sets parameter_sets{ true, false };
        auto resource = nmos::make_connection_rtp_sender(U("sender"), false, parameter_sets);
        auto& transport_params = resource.data.at(nmos::fields::endpoint_active).at(nmos::fields::transport_params);
        auto& params = transport_params.at(0);
        params[nmos::fields::source_ip] = web::json::value::string(U("192.0.2.10"));
        params[nmos::fields::destination_ip] = web::json::value::string(U("232.21.21.133"));

        nmos::resolve_rtp_auto(nmos::types::sender, transport_params, 5000);

        BST_REQUIRE_EQUAL(U("232.21.21.133"), nmos::fields::fec_destination_ip(params).as_string());
        BST_REQUIRE_EQUAL(5002, nmos::fields::fec1D_destination_port(params).as_integer());
        BST_REQUIRE_EQUAL(5004, nmos::fields::fec2D_destination_port(params).as_integer());
        BST_REQUIRE_EQUAL(5002, nmos::fields::fec1D_source_port(params).as_integer());
        BST_REQUIRE_EQUAL(5004, nmos::fields::fec2D_source_port(params).as_integer());
    }

```

lo-simon: It may be worth extending the resolveAuto test to cover the sender's FEC.

Development/nmos/test/connection_resources_test.cpp
```diff
@@ -0,0 +1,100 @@
...
+        auto& transport_params = resource.data.at(nmos::fields::endpoint_active).at(nmos::fields::transport_params);
+        auto& params = transport_params.at(0);
+        params[nmos::fields::interface_ip] = web::json::value::string(U("192.0.2.20"));
+
+        nmos::resolve_rtp_auto(nmos::types::receiver, transport_params, 5000);
+
+        BST_REQUIRE_EQUAL(U("192.0.2.20"), nmos::fields::rtcp_destination_ip(params).as_string());
+        BST_REQUIRE_EQUAL(5001, nmos::fields::rtcp_destination_port(params).as_integer());
+    }
+}
```
lo-simon, line 100:
```suggestion

    {
        const nmos::rtp_receiver_parameter_sets parameter_sets{ false, true, false };
        auto resource = nmos::make_connection_rtp_receiver(U("receiver"), false, parameter_sets);
        auto& transport_params = resource.data.at(nmos::fields::endpoint_active).at(nmos::fields::transport_params);
        auto& params = transport_params.at(0);
        params[nmos::fields::interface_ip] = web::json::value::string(U("192.0.2.20"));

        nmos::resolve_rtp_auto(nmos::types::receiver, transport_params, 5000);

        BST_REQUIRE_EQUAL(U("192.0.2.20"), nmos::fields::fec_destination_ip(params).as_string());
        BST_REQUIRE_EQUAL(5002, nmos::fields::fec1D_destination_port(params).as_integer());
        BST_REQUIRE_EQUAL(5004, nmos::fields::fec2D_destination_port(params).as_integer());
    }
}
```

lo-simon: Same to the receiver's FEC.

lo-simon (changes requested):
Just a couple of minor suggestions; the rest are LGTM.

lo-simon:
Thank you, @thangldw. This is done via #514

PR #502, opened 2026-07-14
Support SMPTE 2022-7 temporal redundancy SDP

thangldw:
## Summary

- add structured SDP grammar for `ssrc`, `ssrc-group`, and `duplication-delay`
- represent SSRC-level temporal redundancy explicitly in `sdp_parameters`, including SSRC IDs, CNAMEs, delay, and MID
- generate the RFC 7104 single-media-description form from two identical IS-05 transport legs
- parse the temporal form back into two receiver transport legs while preserving existing separate-source and separate-destination behavior
- reject malformed SSRC attributes, unsupported leg counts, and non-identical temporal transport legs

## Testing

- exact IS-05 temporal redundancy example
- SDP creation/parsing round trip
- malformed grammar and inconsistent-leg negative cases
- existing SMPTE 2022-7 variants
- full suite: 2027 assertions in 161 test cases

This PR is independent of and complementary to #501, so the two remaining modes can be reviewed separately.

Refs #38

### DIFF
--- a/Development/nmos/sdp_utils.cpp
+++ b/Development/nmos/sdp_utils.cpp
@@ -1,5 +1,6 @@
 #include "nmos/sdp_utils.h"
 
+#include <limits>
 #include <boost/algorithm/string/case_conv.hpp>
 #include <boost/asio/ip/address.hpp>
 #include <boost/range/adaptor/filtered.hpp>
@@ -445,6 +446,22 @@ namespace nmos
             const auto& destination_ip = nmos::fields::destination_ip(transport_param);
             return destination_ip.as_string();
         }
+
+        size_t get_ssrc_duplication_count(const web::json::value& media_description)
+        {
+            const auto& attributes = sdp::fields::attributes(media_description);
+            if (attributes.is_null()) return 0;
+
+            const auto& media_attributes = attributes.as_array();
+            const auto ssrc_group = sdp::find_name(media_attributes, sdp::attributes::ssrc_group);
+            if (media_attributes.end() == ssrc_group) return 0;
+
+            const auto& value = sdp::fields::value(*ssrc_group);
+            return sdp::group_semantics::duplication == sdp::group_semantics_type{ sdp::fields::semantics(value) }
+                && 2 == sdp::fields::ssrc_ids(value).size()
+                ? 2
+                : 0;
+        }
     }
 
     // Make a json representation of an SDP file, e.g. for sdp::make_session_description, from the specified parameters; explicitly specify whether 'source-filter' attributes are included to override the default behaviour
@@ -457,8 +474,25 @@ namespace nmos
         // and the rtp_enabled status does not affect the leg's media description
         // see https://github.com/AMWA-TV/is-05/issues/109#issuecomment-598721418
 
+        const bool temporal_redundancy = !sdp_params.temporal_redundancy.synchronization_sources.empty();
+
+        if (temporal_redundancy)
+        {
+            if (2 != transport_params.size() || 2 != sdp_params.temporal_redundancy.synchronization_sources.size())
+            {
+                throw details::sdp_creation_error("SMPTE ST 2022-7 temporal redundancy requires two transport parameter legs and two synchronization sources");
+            }
+            if (sdp_params.temporal_redundancy.media_stream_id.empty())
+            {
+                throw details::sdp_creation_error("SMPTE ST 2022-7 temporal redundancy requires a media stream id");
+            }
+            if (transport_params.at(0) != transport_params.at(1))
+            {
+                throw details::sdp_creation_error("SMPTE ST 2022-7 temporal redundancy requires identical transport parameter legs");
+            }
+        }
         // check to ensure enough media_stream_ids for multi-leg transport_params
-        if (transport_params.size() > 1 && transport_params.size() > sdp_params.group.media_stream_ids.size())
+        else if (transport_params.size() > 1 && transport_params.size() > sdp_params.group.media_stream_ids.size())
         {
             throw details::sdp_creation_error("not enough sdp parameters media stream ids for transport_params");
         }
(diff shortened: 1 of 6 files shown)


Development/sdp/json.h
```diff
@@ -173,6 +178,15 @@ namespace sdp
         const web::json::field_as_string semantics{ U("semantics") }; // see sdp::group_semantics
         const web::json::field_as_array mids{ U("mids") };
 
+        // a=ssrc:<ssrc-id> <attribute>[:<value>]
+        // a=ssrc-group:<semantics> <ssrc-id>...
+        // See https://tools.ietf.org/html/rfc5576 and https://tools.ietf.org/html/rfc7104
+        const web::json::field<uint64_t> ssrc_id{ U("ssrc_id") };
+        const web::json::field_as_string ssrc_attribute{ U("ssrc_attribute") };
+        const web::json::field_as_value_or ssrc_attribute_value{ U("ssrc_attribute_value"), {} };
```
garethsb: I'm wondering whether it would be possible/nice in the JSON model to represent the "source attribute" exactly the same as a top-level attribute is represented but nested here?

I.e. something like:

```C++
{ sdp::fields::ssrc_id, id },
{ sdp::fields::attribute, value_of({
    { sdp::fields::name, U("cname") },
    { sdp::fields::value, cname }
}) }
```

Development/nmos/sdp_utils.cpp
```diff
@@ -445,6 +446,22 @@ namespace nmos
...
+
+            const auto& media_attributes = attributes.as_array();
+            const auto ssrc_group = sdp::find_name(media_attributes, sdp::attributes::ssrc_group);
+            if (media_attributes.end() == ssrc_group) return 0;
+
+            const auto& value = sdp::fields::value(*ssrc_group);
+            return sdp::group_semantics::duplication == sdp::group_semantics_type{ sdp::fields::semantics(value) }
+                && 2 == sdp::fields::ssrc_ids(value).size()
+                ? 2
+                : 0;
```
garethsb: why not return the actual number of `ssrc_ids` here, and **if necessary** bail out higher up if it's not a supported number (`0` or `2`)?

```suggestion
            return sdp::group_semantics::duplication == sdp::group_semantics_type{ sdp::fields::semantics(value) }
                ? sdp::fields::ssrc_ids(value).size()
                : 0;
```

Development/nmos/sdp_utils.cpp
```diff
@@ -1125,6 +1208,59 @@ namespace nmos
...
+            const auto& media_attributes = sdp::fields::attributes(media_description);
+            if (media_attributes.is_null()) continue;
+
+            const auto& attributes = media_attributes.as_array();
+            const auto ssrc_group = sdp::find_name(attributes, sdp::attributes::ssrc_group);
+            if (attributes.end() == ssrc_group) continue;
+
+            const auto& ssrc_group_value = sdp::fields::value(*ssrc_group);
+            if (sdp::group_semantics::duplication != sdp::group_semantics_type{ sdp::fields::semantics(ssrc_group_value) }) continue;
+            if (2 != sdp::fields::ssrc_ids(ssrc_group_value).size()) continue;
```
garethsb, line 1221: This duplicates some of the logic of `get_ssrc_duplication_count`... but given that function isn't directly usable here, and that means `get_ssrc_duplication_count` is only used in one place, maybe better to drop that function and inline its code into the single call-site?

garethsb (commented):
Also looks pretty good, thank you. Just a few comments so far!

Development/sdp/json.h
thangldw: Agreed. Representing the SSRC source attribute as a nested name/value object reuses the same JSON shape as a top-level SDP attribute and avoids baking cname-specific fields into the model. I will replace the separate SSRC attribute name/value fields with an attribute object, then update the grammar, generation/parsing paths, and round-trip tests accordingly.

Development/nmos/sdp_utils.cpp
thangldw: Agreed. Since the helper has a single call site and the temporal-redundancy parser necessarily performs the same group lookup for additional data, keeping the helper does not buy enough reuse. I will remove it and inline the DUP SSRC-group lookup at the transport-parameter call site, while leaving the richer temporal-redundancy parsing logic in its existing layer.

Development/nmos/sdp_utils.cpp
thangldw: Agreed. The detection code should preserve the actual SSRC cardinality, and the NMOS interpretation layer should own the supported-count decision. In combination with the later comment about the helper having only one call site, I will remove the helper, compute the actual SSRC ID array size at that call site, and keep the two-source requirement only where temporal redundancy is interpreted.

thangldw:
The review updates are now pushed in commit 87f9ee2.

This update:

- represents each SSRC source attribute as a nested name/value attribute object;
- removes the single-use duplication-count helper and computes the actual SSRC ID cardinality at its call site;
- keeps the two-source support decision in the temporal-redundancy interpretation path;
- adds direct coverage for the nested JSON representation.

The focused temporal-redundancy tests pass with 18 assertions across 2 test cases, the existing separate-source/destination tests pass, and the full local suite passes with 2,030 assertions across 161 test cases. The temporary Apple Silicon Catch compatibility patch used by the local test build was fully reverted before committing.

The new GitHub Actions suite is currently marked action_required and appears to be waiting for repository approval.

garethsb (approved):
LGTM. I recommend merging.

The only real issue with this is the **pre-existing** one that `nmos::get_session_description_transport_params` is a bit of a convoluted state machine for both classifying the SDP "shape" and simultaneously collecting the transport parameters for each leg. A future refactor might avoid some of the hard-to-understand-in-isolation conditionals* by having a classification function (e.g. enum of `separate_destination`, `separate_source`, `temporal_redundancy`, `single_leg`) and then using shape-specific parsing.

*E.g. `sa.size() <= source_address && !(1 == sa.size() && source_address < ssrc_duplication_count)`

thangldw:
Thanks for the review. Agreed that separating SDP shape classification from shape-specific parsing would be a useful follow-up; I’ll keep that out of this PR.

lo-simon (approved):
LGTM

PR #505, opened 2026-07-18
Support RFC 6364 and IS-05 SMPTE 2022-5 FEC SDP

thangldw:
## Summary

This implements the layered FEC approach discussed in #38:

- add structured RFC 6364 SDP support for `FEC/UDP`, `UDP/FEC`, `FEC-FR`, `fec-source-flow`, `fec-repair-flow`, and `repair-window` in `sdp/`;
- map the IS-05-compatible SMPTE 2022-5 topology (one RTP source flow with one or two repair flows per transport leg) in `nmos/sdp*`;
- keep RFC-only source IDs, encoding IDs, preference, SS-FSSI/FSSI, repair windows, and media stream IDs explicit in `sdp_parameters` rather than synthesizing them from IS-05 values;
- resolve receiver `fec_mode: auto` to the highest configured dimension in the example node.

## RFC / NMOS boundary

The generic SDP representation accepts the RFC 6364 ABNF, including optional parameters and empty FSSI element values. It validates numeric ranges and the RFC token/separator rules.

The NMOS layer deliberately handles only the form IS-05 can represent. RFC-valid multi-source groups remain available in the generic `sdp/` JSON but are not converted to FEC transport parameters. Ambiguous/lossy cases, such as two repair flows with different destination addresses, are rejected because IS-05 exposes only one `fec_destination_ip`.

The `fec_enabled` field addition overlaps with #501. It is included here so this PR remains independently buildable from `master`; I will drop the duplicate during rebase if #501 merges first.

## Tests

- RFC example-style round trip covering optional source tag, preference, SS-FSSI/FSSI, empty FSSI values, and both repair-window units
- malformed ABNF, range, ordering, and separator cases
- exact IS-05 SMPTE 2022-5 example interpretation and text round trip
- 2D generation/interpretation with two repair flows
- unsupported topology, missing media reference, and unrepresentable destination-address cases
- `nmos-cpp-node` build
- full local suite: **2092 assertions in 164 test cases**

Related to #38.

### DIFF
--- a/Development/nmos-cpp-node/node_implementation.cpp
+++ b/Development/nmos-cpp-node/node_implementation.cpp
@@ -2105,6 +2105,13 @@ nmos::connection_resource_auto_resolver make_node_implementation_auto_resolver(c
             const bool smpte2022_7 = 1 < transport_params.size();
             nmos::details::resolve_auto(transport_params[0], nmos::fields::interface_ip, [&] { return web::json::front(nmos::fields::constraint_enum(constraints.at(0).at(nmos::fields::interface_ip))); });
             if (smpte2022_7) nmos::details::resolve_auto(transport_params[1], nmos::fields::interface_ip, [&] { return web::json::back(nmos::fields::constraint_enum(constraints.at(1).at(nmos::fields::interface_ip))); });
+            for (auto& params : transport_params.as_array())
+            {
+                // The IS-05 receiver schema defines "auto" as the highest available number of dimensions.
+                nmos::details::resolve_auto(params, nmos::fields::fec_mode, [&] {
+                    return value::string(!nmos::fields::fec2D_destination_port(params).is_null() ? U("2D") : U("1D"));
+                });
+            }
             // lastly, apply the specification defaults for any properties not handled above
             nmos::resolve_rtp_auto(id_type.second, transport_params);
         }
--- a/Development/sdp/sdp_grammar.cpp
+++ b/Development/sdp/sdp_grammar.cpp
@@ -1,6 +1,7 @@
 #include "sdp/sdp_grammar.h"
 #include "sdp/sdp.h"
 
+#include <limits>
 #include <stdexcept>
 #include "bst/regex.h"
 #include "cpprest/basic_utils.h"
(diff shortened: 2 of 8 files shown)


Development/nmos-cpp-node/node_implementation.cpp
```diff
@@ -2105,6 +2105,13 @@ nmos::connection_resource_auto_resolver make_node_implementation_auto_resolver(c
             const bool smpte2022_7 = 1 < transport_params.size();
             nmos::details::resolve_auto(transport_params[0], nmos::fields::interface_ip, [&] { return web::json::front(nmos::fields::constraint_enum(constraints.at(0).at(nmos::fields::interface_ip))); });
             if (smpte2022_7) nmos::details::resolve_auto(transport_params[1], nmos::fields::interface_ip, [&] { return web::json::back(nmos::fields::constraint_enum(constraints.at(1).at(nmos::fields::interface_ip))); });
+            for (auto& params : transport_params.as_array())
+            {
+                // The IS-05 receiver schema defines "auto" as the highest available number of dimensions.
+                nmos::details::resolve_auto(params, nmos::fields::fec_mode, [&] {
+                    return value::string(!nmos::fields::fec2D_destination_port(params).is_null() ? U("2D") : U("1D"));
+                });
```
garethsb, line 2113: I interpreted that to mean highest available in the receiver implementation... so I would have based this on the `constraints` I think? If it means highest available in the stream being received, then I see the logic here...

Development/sdp/sdp_grammar.cpp
```diff
@@ -505,21 +683,40 @@ namespace sdp
...
+                    v[sdp::fields::media_type] = string_converter.parse(substr_find(s, pos, " "));
+                    if (std::string::npos == pos) throw sdp_parse_error("expected a media port");
+                    const auto port = port_converter.parse(substr_find(s, pos, " "));
+                    web::json::insert(v, port.as_object().begin(), port.as_object().end());
+                    if (std::string::npos == pos) throw sdp_parse_error("expected a media protocol");
+                    v[sdp::fields::protocol] = string_converter.parse(substr_find(s, pos, " "));
+                    v[sdp::fields::formats] = strings_converter.parse(std::string::npos != pos ? substr_find(s, pos) : "");
+                    return v;
+                }
+            };
```
garethsb: The custom `media_converter` feels surprisingly verbose compared with the existing declarative `object_converter` form. Maybe `object_converter` gains an overload for an optional trailing field, with that converter receiving the remaining input? Then this could retain the original shape except the the `{ std::fields::formats, strings_converter }` becomes optional, which would permit UDP/FEC media descriptions without formats.

thangldw: Agreed, updated in d6b3db3. `object_converter` now supports an optional trailing field that consumes the remaining input, and the media converter is declarative again with `formats` as that optional tail. I also added coverage for UDP/FEC without formats and for media lines with multiple formats.

garethsb: I suggested a version which I think is slightly more in keeping with the original code - https://github.com/thangldw/nmos-cpp/pull/1

Development/nmos-cpp-node/node_implementation.cpp
garethsb: Ah, the schema doesn't allow `null` for the ports, so you definitely can't infer the available dimensions from the port values. I'm more confident to say we should base it on the `/constraints` schema.

lo-simon:
Close, now superseded by #515

Issue #187, opened 2021-06-17
Deprecated dependency: cpprestsdk is now in maintenance mode


connorfeeley:
As of March 8th, [cpprestsdk is in maintenance mode](https://github.com/microsoft/cpprestsdk/commit/0ddc6182943f43a71453c97640ae7987d1868ea8#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5).

From the README:
`cpprestsdk is in maintenance mode and we do not recommend its use in new projects. We will continue to fix critical bugs and address security issues.`

There are currently about 700 open issues, and the last commit was on March 11th.

Given that the project appears essentially abandoned (and is unlikely to receive fixes for new compilers or boost releases), are there any plans to migrate to a different REST library?

This currently doesn't seem to be a pressing issue, but in the longer term the `cpprestsdk` dependency may cause compatibility problems.

garethsb:
That's a good question. Personally (and definitely not speaking for the current team at Sony), I think investigating options here is a good idea. At the point that cpprestsdk was selected, it provided the best balance of features and platform and compiler support. Other potential libraries were more bloated or immature. That has changed. However, there are obviously parts of the nmos-cpp design that are currently tied to the way that cpprestsdk provides these features.

I agree that in the short term, there are few concerns, nmos-cpp is fairly stable and both it and cpprestsdk have weathered a series of evolutions of platform and compiler already. There are no obvious missing infrastructure features required for current or near term NMOS enhancements. The cpprestsdk codebase could be maintained fit for this project's purpose for sometime.

That said, open-source C++ cross platform networking libraries like Boost.Beast have matured in the last couple of years, and with json support provided by e.g. nlohmann/json or Boost.JSON, and several options for async task management to replace pplx, nmos-cpp could be transitioned, especially if the minimum C++ standard requirement were updated past GCC 4.8—era C++11.

This issue could be a good place to identify the options between those extremes.

garethsb:
I thought I'd make a start on a list of features to look for when evaluating alternative libraries, besides the obvious ones of OSS licence, level of project maintenance/activity, cross platform, C++ language version/compiler support (CI tests!), sufficient performance, Conan/vcpkg support, and recursively for additional dependencies... 

### Network I/O Features

Used by nmos-cpp now:
* HTTP/1.1 client
* HTTP/1.1 server
* WebSocket client
* WebSocket server (one of nmos-cpp's cpprest extensions)
[...]

PR #507, opened 2026-07-20
Test Visual Studio 2026 and update dependencies

garethsb:
Move the primary Windows matrix to VS 2026 while retaining VS 2022 coverage, and update CMake and Conan dependencies to current supported releases.


garethsb:
@lo-simon, @jonathan-r-thorpe, @aholzinger ... so I think there's an approach here to support VS 2026 and update deps (if we're happy with Boost 1.86.0) ... thoughts?

aholzinger:
@garethsb Boost 1.86 does only help for current VS 2022 due to BoostDetectToolset.cmake:

`  if((MSVC_VERSION GREATER 1929) AND (MSVC_VERSION LESS 1950))`

`    set(BOOST_DETECTED_TOOLSET "vc143")`


In Boost 1.85 it does still look like this (1940 is too small for current VS 2022):

`  if((MSVC_VERSION GREATER 1929) AND (MSVC_VERSION LESS 1940))`

`    set(BOOST_DETECTED_TOOLSET "vc143")`


Regarding VS 2026, please note that there is no "vc145" in the above.

VS 2026 is supported only since Boost 1.91:

`  if((MSVC_VERSION GREATER 1949) AND (MSVC_VERSION LESS 1960))`

`    set(BOOST_DETECTED_TOOLSET "vc145")`

`  elseif((MSVC_VERSION GREATER 1929) AND (MSVC_VERSION LESS 1950))`

`    set(BOOST_DETECTED_TOOLSET "vc143")`

garethsb:
> VS 2026 is supported only since Boost 1.91

Hm, seems to be building and passing tests here though? What am I missing?

aholzinger:
@garethsb It builds, but doesn't specialize for the toolset and emits this message (always BoostDetectToolset.cmake):

`message(STATUS "Boost toolset is unknown (compiler ${CMAKE_CXX_COMPILER_ID} ${CMAKE_CXX_COMPILER_VERSION})")`

garethsb:
I don't think that affects _correctness_, does it? `NmosCppDependencies.cmake` relies on `BOOST_ALL_NO_LIB` so MSVC auto-link is off, the libraries are linked via the `Boost::` targets, so having the toolset identifier in the library names is irrelevant?

Moving all of nmos-cpp, cpprestsdk and websocketpp past Boost 1.86.0 would be a bit of a headache? (Not the code changes, but using forked repos and getting those approved into Conan Center recipes, for example.)

aholzinger:
Well, I'm not a Boost build expert, but I wouldn't use Boost with an unsupported toolset in a system critical application (without prior having checked all the inplications, ergo becoming a Boost build expert). For the time being I can live with VS 2022.

So for the moment I suggest to stick with Boost 1.86 (meaning not officially support VS 2026), but for the long term (Boost 1.86 is already two years old) the cpprestsdk issue has to be resolved anyhow, which is the way bigger issue. If a replacement is found I guess the changes in nmos-cpp regarding updated Boost Asio in newer Boost versions will be the smalles part.

garethsb:
> for the long term (Boost 1.86 is already two years old) the cpprestsdk issue has to be resolved anyhow, which is the way bigger issue. If a replacement is found I guess the changes in nmos-cpp regarding updated Boost Asio in newer Boost versions will be the smalles part.

True... That's https://github.com/sony/nmos-cpp/issues/187.

Our (@lo-simon, @jonathan-r-thorpe) feeling until recently has been that we would maintain a feature-frozen fork of cpprestsdk if that became necessary, because we don't need **new features** for nmos-cpp, just platform/build/security updates. However, with 🤖 support it might now be practical to switch nmos-cpp to other async-task, JSON and HTTP/WebSocket libraries, though I suspect not without a lot of pain for application developers who have directly integrated nmos-cpp (rather than using one of the API wrappers like [NVIDIA/nvnmos](https://github.com/NVIDIA/nvnmos)'s [C API](https://nvidia.github.io/nvnmos/c-api-guide.html) or [gRPC daemon](https://nvidia.github.io/nvnmos/grpc/) or [GStreamer plugin](https://nvidia.github.io/nvnmos/gstreamer/)).

aholzinger:
Well with Mythos you can probably fix the biggest part of any still unknown issues in cpprestsdk :-)

I was able to build nmos-cpp (without Conan) on a freshly installed Ubuntu based 26.04 LTS due to the fact that there was still Boost 1.83 available in the Ubuntu 26.04 repos. Once Boost 1.83 isn't in available anymore in the next LTS it will be less easy. At least then it would be nice to have nmos-cpp2 relying on current libs (at least Boost).

Obviously any changes in nmos-cpp will also hit myself then :-)

Thank you for considering Boost 1.86 for the moment.

Issue #506, opened 2026-07-19
Support more current Boost for supporting VS2022 and probably VS2026


aholzinger:
Latest officially supported Boost version is 1.83.
Boost 1.83 is supporting up to MSC_VER 1939 (LESS 1940) which is VS2022 17.8 which is EOL.
Only VS2022 that is not EOL is VS2022 17.14 which is MSC_VER 1944 which obviously is not supported by Boost version 1.83.

Since Boost version 1.86 MSC_VER up to 1949 (LESS 1950) is supported, but in Development/cpprest/host_utils.cpp there is `#include <boost/asio/io_service.hpp>` which in Boost version 1.86 is no more existing. It's well probable that this is not the only isse with Boost version 1.86.

So currently nmos-cpp does not support VS2022 non-EOL. Neither it does support V2026.

Since Boost version 1.91 (current Boost version as of writing this issue) VS2026 is supported up to MSC_VER 1959 (LESS 1960). Latest VS2026 18.8 comes with MSC_VER 1952.

So as a minum step forward for nmos-cpp would be to support Boost version 1.86.

A better step forward would be support Bost version 1.91.

garethsb:
I've been working on fixes for VS 2026 intermittently, I got it building with Boost 1.83.0 but can look at bumping Boost as well. Ref: https://github.com/NVIDIA/nvnmos/pull/60

aholzinger:
Correcting my own text:
boost/asio/io_service.hpp is still available in Boost 1.86, it's in 1.87 that's no more there.

So Boost version 1.86 could be an intermediate step to support current VS2022.

garethsb:
Yes, I think the removal of the old ASIO APIs that are used by nmos-cpp, cpprestsdk and websocketpp are a bit of a challenge. I'll see what I can do with Boost 1.86.0.

garethsb:
| Component | Boost ≥1.87 status |
|---|---|
| **websocketpp** | https://github.com/zaphoyd/websocketpp/pull/1164 not merged. Distros (FreeBSD, Gentoo, …) carry their own patches. Proposal for a patch for the Conan Center recipe was rejected (https://github.com/conan-io/conan-center-index/issues/26732) |
| **cpprestsdk** | https://github.com/microsoft/cpprestsdk/issues/1815 indicates there’s a fork branch and vcpkg / FreeBSD patches. Microsoft’s repo is archived. We've had a long-time plan to fork and maintain for nmos-cpp. |
[...]

PR #508, opened 2026-07-25
Add support for negative logging categories in the log gateway

### LINKED ISSUE #338: [Enhancement] Negative logging categories
We could add support to the default/example logging gateway so that categories that should _not_ be logged can be specified in the _config.json_, e.g. with a simple prefix like "!access". This would prevent logging of messages with this category even if other categories match positively.

https://github.com/sony/nmos-cpp/blob/14fd314890a85fdf857acfc30821aeaf1bbb35db/Development/nmos/log_gate.h#L57

ZayanKhan-12:
Closes #338

Implements the enhancement suggested in #338: categories prefixed with `!` in the `logging_categories` setting specify messages that should **not** be written to the error log.

## Semantics

- A negative match takes precedence over any positive match, so a message tagged both `send_query_ws_events` and `access` is suppressed by `["send_query_ws_events", "!access"]`.
- When only negative categories are specified, all other messages remain pertinent — so `["!access"]` means "everything except access", which makes the feature usable on its own.
- Existing behaviour is unchanged when `logging_categories` is omitted (everything logged), explicitly empty (nothing logged), or contains only positive categories, including the empty-string entry that selects messages with no category.

The settings schema already accepts these values (`stringArray`), and the `logging_categories` comment in `nmos/settings.h` is updated to document the prefix.

## Testing

New `nmos/test/log_gate_test.cpp` covers all the cases above via a small `log_gate` subclass (20 assertions). Full local run: **all 169 test cases / 2361 assertions pass** (macOS arm64, apple-clang, conan dependencies per `conanfile.txt`).

Two environment notes from getting the suite running on an arm64 Mac, in case they're useful for #506:

- vendored `third_party/catch/catch.hpp` needs an `__aarch64__` branch for `CATCH_TRAP()` (`int $3` doesn't assemble on arm64) and an unsigned `RandomNumberGenerator::result_type` for modern libc++ — happy to submit separately if of interest;
- Boost 1.90 removes the (long header-only) `Boost.System` stub entirely, so `find_package(Boost COMPONENTS system ...)` fails against current Boost — dropping `system` from `FIND_BOOST_COMPONENTS` works fine with 1.83+.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Development/cmake/NmosCppTest.cmake
+++ b/Development/cmake/NmosCppTest.cmake
@@ -56,6 +56,7 @@ set(NMOS_CPP_TEST_NMOS_TEST_SOURCES
     nmos/test/json_validator_test.cpp
     nmos/test/jwt_generator_test.cpp
     nmos/test/jwt_validation_test.cpp
+    nmos/test/log_gate_test.cpp
     nmos/test/mdns_test.cpp
     nmos/test/node_interfaces_test.cpp
     nmos/test/paging_utils_test.cpp
--- a/Development/nmos/log_gate.h
+++ b/Development/nmos/log_gate.h
@@ -63,14 +63,36 @@ namespace nmos
 
                 const auto& pertinent_categories = nmos::fields::logging_categories(model.settings);
 
+                // categories prefixed with '!' specify messages that should not be logged;
+                // a negative match takes precedence over any positive match, and when only
+                // negative categories are specified, all other messages are pertinent
+                const auto is_negative = [](const web::json::value& category)
+                {
+                    const auto& s = category.as_string();
+                    return !s.empty() && U('!') == s.front();
+                };
+                const bool default_pertinent = 0 != pertinent_categories.size()
+                    && pertinent_categories.end() == boost::range::find_if(pertinent_categories, [&](const web::json::value& category)
+                    {
+                        return !is_negative(category);
+                    });
+
                 if (categories.empty())
                 {
                     static const auto no_category = web::json::value::string(utility::string_t());
-                    return pertinent_categories.end() != boost::range::find(pertinent_categories, no_category);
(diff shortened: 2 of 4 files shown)


garethsb (changes requested):
Nice, focused change that matches #338. I recommend a few small tweaks before merge.

**1. Align example config comments**  
`nmos-cpp-node/config.json` and `nmos-cpp-registry/config.json` still only describe the include form. Please mirror the updated `settings.h` wording with one extra line on the `!` prefix (and the “only excluded categories ⇒ everything else included” behaviour).

**2. Treat `"!"` as negated `""`**  
Today `"!"` is classified as negative (so it can enable “negation-only” mode) but the empty-categories path only looks for `""`, so `"!"` never excludes uncategorized messages and is effectively a no-op.

Suggestion: in the `categories.empty()` branch, if `"!"` is present return false (negative wins), else keep `default_pertinent || has ""`. That makes the prefix rule complete and allows useful configs such as `["!"]` (everything except uncategorized) and `["!", "!access"]`.

**3. Extra test**  
Please add a case for `["access", "!access"]` (negative wins over the same positive), and for `"!"` excluding uncategorized once (2) is in.

**4. CI**  
There are no checks reported on this branch yet. @lo-simon please could you allow workflows to run?

Thanks also for the Catch/Boost notes for #506...
* I'm surprised about the need for `__aarch64__` as I thought this has been built on Linux arm64 and that there was a current macOS runner 🤔
* we have bigger issues with recent Boost (since Asio deprecated API has been removed) but this sounds like an easy win in another PR

lo-simon:
Thank you so much, @ZayanKhan-12 and @garethsb, for your contribution. This PR has been superseded by #519, which has now been merged.

PR #512, opened 2026-07-28
Expose SDP grammar converters for application-defined attributes

garethsb:
`session_description` already accepted custom `attribute_converters`, but the helper converters needed to build them were internal, and the lifetime of those converters relative to the returned grammar was undocumented. Declare the reusable converters and generators, document the by-reference capture and static-initialization constraints, and add a test that extends the default grammar with `a=x-example-foo:<pt> <value>`.

Probable merge conflict with #505 - whichever is merged later will need rebasing.

### DIFF
--- a/Development/sdp/sdp_grammar.cpp
+++ b/Development/sdp/sdp_grammar.cpp
@@ -139,7 +139,7 @@ namespace sdp
             };
         }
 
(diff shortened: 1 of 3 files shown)


lo-simon (approved):
LGTM

PR #517, opened 2026-08-01
Fix path traversal vulnerability

lo-simon:
A path traversal vulnerability has been identified in the Registry Admin UI in Windows. The vulnerability exists in `make_filesystem_route`, where the code sanitises user-controlled file paths by checking for forward-slash parent directory sequences (/..) but completely ignores the backslash variant (\..), which is a valid path separator on Windows systems.

How the fix works:
1. Normalise path separators to forward slashes
    -  /foo/./bar\\..\baz -> /foo/./bar/../baz
2. Split the path by/into segments
    -	/foo/./bar/../baz -> ["foo", ".", "bar", "..", "baz"]
3. Processes each segment:
    -	. (current directory) -> Skip it
    -	.. (parent directory) -> Pop the last segment from the stack
    -	Normal segment -> Add to stack
4. Detects traversal attacks:
    -	If .. tries to pop from an empty stack -> Returns false (attempting to escape root)
5. Reconstructs the normalised path:
    -	["foo", "baz"] -> /foo/baz

Examples:
| Input | Normalised Output | Result |
| ----------- | ----------- |----------- |
| /foo/./bar | /foo/bar | ✅ Allowed |
| /foo/../bar | /bar | ✅ Allowed |
| /foo/bar/../../baz | /baz | ✅ Allowed |
| /../secret | N/A | ❌ Forbidden (traversing above root) |
| /foo/../../etc/passwd | N/A | ❌ Forbidden (traversing above root) |
| /../ | N/A | ❌ Forbidden (traversing above root) |


garethsb:
Seems excessive. Previously we just banned any use of "/..". If the problem is Windows, a tiny tweak to that would be enough.

lo-simon:
> Seems excessive. Previously we just banned any use of "/..". If the problem is Windows, a tiny tweak to that would be enough.

In the previous implementation, paths containing "/.." were rejected, but in this implementation we ensure the path is normalised and resolved before the traversal check. i.e. the ".." and "." are both supported for all platforms. Although it seems a little heavier than before, it is only because `std::filesystem::canonical` and `std::filesystem::relative` are not available in C++11.

garethsb:
I think supporting `..` traversal is unnecessary though?

garethsb:
Why not just ban URL paths with `\` completely?

garethsb (approved):
LGTM - "the simplest thing that could possibly work"

Recommend squashing the commits before or while doing the merge.

PR #518, opened 2026-08-07
Parse SDP encoding names and fmtp names case-insensitively

garethsb:
## Summary
- Treat RTP encoding names and fmtp parameter names as case-insensitive when parsing SDP, per RFC 4855.
- Keep `media_type::operator==` exact for stored/configured values; add public `equals_media_type` for SDP-derived comparisons (receivers, validation, JPEG XS gate, node example).
- Move `find_fmtp` / `equals_encoding_name` into `sdp_utils.cpp` so Boost stays out of the headers.

## Test plan
- [x] `nmos-cpp-test testSdpEncodingNameAndFmtpCaseInsensitive`
- [x] `nmos-cpp-test 'testSdp*'`
- [x] `testValidateSdpParameters`, JPEG XS, and related SDP cases
- [x] CI on this PR

### DIFF
--- a/Development/cmake/NmosCppLibraries.cmake
+++ b/Development/cmake/NmosCppLibraries.cmake
@@ -1093,6 +1093,7 @@ set(NMOS_CPP_NMOS_SOURCES
     nmos/manifest_api.cpp
     nmos/mdns.cpp
     nmos/mdns_api.cpp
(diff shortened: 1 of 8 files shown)


lo-simon (approved):
LGTM

PR #519, opened 2026-08-12
Fix/negative logging categories

### LINKED ISSUE #338: [Enhancement] Negative logging categories

garethsb:
Replaces #508, to resolve #338. Based on @ZayanKhan-12 original PR, just adding my requested tweaks.


lo-simon (approved):
LGTM

PR #520, opened 2026-08-19
Fall back for invalid Node interface port IDs

### LINKED ISSUE #496: Node registration fails when host has tunnel/virtual interfaces with non-MAC port_id
When running nmos-cpp Node in Docker (host networking), registration with the Registry fails with 400 Bad Request due to IS-04 schema validation failure. The node resource includes a tunnel interface (tunl0) whose port_id is 4 octets instead of the required 6-octet MAC.

- nmos-cpp Node and Registry in separate Docker containers
- Host networking / multiple host_addresses (physical NIC + Docker bridge IPs)
- Linux host with interfaces such as enP7s7, docker0, br-*, tunl0

### Registry log
```json
warning: JSON error: schema validation failed at root - no subschema has succeeded ...
{"type":"node","data":{...,"interfaces":[
  {"name":"enP7s7","port_id":"4c-bb-47-81-1f-e8",...},
  {"name":"docker0","port_id":"be-e8-aa-9a-16-c2",...},
  {"name":"br-47379b15f1c5","port_id":"d2-38-b9-d4-7e-80",...},
  {"name":"tunl0","port_id":"00-00-00-00",...}
]}}
```
### Node log
```
Registration creation error: 400 Bad Request
```
Discovery and the POST to the Registration API succeed; only schema validation fails.

### Root cause
On Linux, `host_utils.cpp` reads link-layer addresses from `AF_PACKET` and formats `sll_halen` bytes as-is. Tunnel interfaces like `tunl0` report `sll_halen == 4`, producing `"00-00-00-00"`.

IS-04 requires `port_id` to match `^([0-9a-f]{2}-){5}([0-9a-f]{2})$` (6 octets).

`make_node_interfaces_port_id()` in `node_interfaces.cpp` falls back to `"00-00-00-00-00-00"` only when `port_id` is empty — not when it is non-empty but invalid:
```c
return value::string(!port_id.empty() ? port_id : U("00-00-00-00-00-00"));
```
### Workaround
Restrict `host_addresses` to the single IP clients should use (e.g. the physical NIC), so tunnel interfaces are excluded:
```json
"host_address": "192.168.6.54",
"host_addresses": ["192.168.6.54"]
```
### Suggested fix
When `physical_address` is not a valid 6-octet MAC, either:

- normalize to `"00-00-00-00-00-00"` (already used for missing addresses), or
- exclude the interface from the node resource

This would make container/virtualised deployments more robust without requiring users to know which kernel-reported interfaces have non-MAC link addresses.

shin4141:
## Summary

- use the repository's existing `00-00-00-00-00-00` fallback when a Node interface port ID cannot be represented in the IS-04 six-octet lowercase-hyphen form
- leave valid port IDs unchanged
- keep host discovery and interface membership unchanged

Fixes #496.

## Why

Tunnel and virtual interfaces may expose non-6-octet link-layer addresses. Since `interfaces[].port_id` requires the IS-04 six-octet lowercase-hyphen form, those values can make the Node resource schema-invalid. This change addresses the invalid `port_id` path reported in #496 by applying the existing null-address fallback at serialization.

## Verification

- valid lowercase-hyphen port ID remains unchanged
- empty, 4/5/7-octet, malformed-hex, uppercase, and colon-separated inputs use the fallback
- focused tests: 3 test cases / 8 assertions
- affected NodeInterface tests: 6 test cases / 21 assertions
- full native suite: 186 test cases / 2597 assertions
- `git diff --check`

## Scope

The regression uses controlled inputs and does not claim a Linux `tunl0` end-to-end reproduction. This PR does not claim a maintainer preference for normalize over exclude, Sony adoption or acceptance, or cross-platform execution beyond the tested macOS arm64 / Apple clang 21 environment.

### DIFF
--- a/Development/nmos/node_interfaces.cpp
+++ b/Development/nmos/node_interfaces.cpp
@@ -20,13 +20,35 @@ namespace nmos
             return chassis_id.is_null() ? utility::string_t{} : chassis_id.as_string();
         }
 
+        bool is_valid_node_interfaces_port_id(const utility::string_t& port_id)
+        {
+            if (17 != port_id.size()) return false;
+
+            for (size_t index = 0; index < port_id.size(); ++index)
+            {
+                const auto character = port_id[index];
+                if (2 == index % 3)
+                {
+                    if (U('-') != character) return false;
+                }
+                else if (!((U('0') <= character && character <= U('9')) || (U('a') <= character && character <= U('f'))))
+                {
+                    return false;
+                }
+            }
+
+            return true;
+        }
+
         // Port ID must be a MAC address
         web::json::value make_node_interfaces_port_id(const utility::string_t& port_id)
         {
             using web::json::value;
-            // when no physical address is available, use the common null value of all zeros
+            // IS-04 port_id requires the six-octet lowercase-hyphen form. Any other representation,
+            // including uppercase or colon-separated MAC addresses, is not representable in this field
+            // and uses the existing null-address fallback of all zeros
             // see https://standards.ieee.org/content/dam/ieee-standards/standards/web/documents/tutorials/eui.pdf
-            return value::string(!port_id.empty() ? port_id : U("00-00-00-00-00-00"));
+            return value::string(is_valid_node_interfaces_port_id(port_id) ? port_id : U("00-00-00-00-00-00"));
         }
     }
 
--- a/Development/nmos/node_interfaces.cpp
+++ b/Development/nmos/node_interfaces.cpp
@@ -1,6 +1,8 @@
 #include "nmos/node_interfaces.h"
 
 #include <boost/range/adaptor/transformed.hpp>
+#include "bst/regex.h"
+#include "cpprest/basic_utils.h"
 #include "cpprest/host_utils.h"
 #include "nmos/json_fields.h"
 
@@ -20,24 +22,13 @@ namespace nmos
             return chassis_id.is_null() ? utility::string_t{} : chassis_id.as_string();
         }
 
+        // Port ID must be a MAC address, strictly following the lowercase hexadecimal format and separated by hyphens (not colons)
+        // It should match the regular expression pattern ^([0-9a-f]{2}-){5}([0-9a-f]{2})$
+        // see https://specs.amwa.tv/is-04/branches/v1.2.x/APIs/schemas/with-refs/node.html
         bool is_valid_node_interfaces_port_id(const utility::string_t& port_id)
         {
-            if (17 != port_id.size()) return false;
-
-            for (size_t index = 0; index < port_id.size(); ++index)
-            {
-                const auto character = port_id[index];
-                if (2 == index % 3)
-                {
-                    if (U('-') != character) return false;
-                }
-                else if (!((U('0') <= character && character <= U('9')) || (U('a') <= character && character <= U('f'))))
-                {
-                    return false;
-                }
-            }
-
-            return true;
+            static const bst::regex port_id_regex(R"(([0-9a-f]{2}-){5}[0-9a-f]{2}$)");
+            return bst::regex_match(utility::us2s(port_id), port_id_regex);
         }
 
         // Port ID must be a MAC address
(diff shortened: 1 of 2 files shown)


Development/nmos/node_interfaces.cpp
lo-simon, line 3:
```suggestion
#include <boost/range/adaptor/transformed.hpp>
#include "bst/regex.h"
#include "cpprest/basic_utils.h"
```

lo-simon: It is better to use `regex` to validate `port_id`, see the suggestion in is_valid_node_interfaces_port_id()

Development/nmos/node_interfaces.cpp
```diff
@@ -20,13 +20,35 @@ namespace nmos
...
+                    if (U('-') != character) return false;
+                }
+                else if (!((U('0') <= character && character <= U('9')) || (U('a') <= character && character <= U('f'))))
+                {
+                    return false;
+                }
+            }
+
+            return true;
+        }
```
lo-simon, line 32:
```suggestion
        // Port ID must be a MAC address, strictly following the lowercase hexadecimal format and separated by hyphens (not colons)
        // It should match the regular expression pattern ^([0-9a-f]{2}-){5}([0-9a-f]{2})$
        // see https://specs.amwa.tv/is-04/branches/v1.2.x/APIs/schemas/with-refs/node.html
        bool is_valid_node_interfaces_port_id(const utility::string_t& port_id)
        {
            static const bst::regex port_id_regex(R"(([0-9a-f]{2}-){5}[0-9a-f]{2}$)");
            return bst::regex_match(utility::us2s(port_id), port_id_regex);
        }
```

Development/nmos/test/node_interfaces_test.cpp
```diff
@@ -45,6 +61,36 @@ BST_TEST_CASE(testMakeParseNodeInterfaceNullChassisId)
...
+////////////////////////////////////////////////////////////////////////////////////////////
+BST_TEST_CASE(testMakeNodeInterfaceInvalidPortIdFallback)
+{
+    const std::vector<utility::string_t> invalid_port_ids{
+        U("00-00-00-00"),
+        U("00-00-00-00-00"),
+        U("00-00-00-00-00-00-00"),
+        U("gg-00-00-00-00-00"),
+        U("AA-BB-CC-DD-EE-FF"),
+        U("00:00:00:00:00:00")
```
lo-simon:
```suggestion
        // Uppercase MAC addresses
        U("AA-BB-CC-DD-EE-FF"),
        U("AA-bb-cc-dd-ee-ff"),
        // Colon-separated MAC addresses
        U("00:00:00:00:00:00"),
        U("aa:bb:cc:dd:ee:ff"),
        U("12:34:56:78:9a:bc"),
        // Various malformed MAC addresses
        // Wrong length
        U("00-00-00-00"),
        U("00-00-00-00-00"),
        U("00-00-00-00-00-00-00"),
        U("aa-bb-cc-dd-ee"),
        U("aa-bb-cc-dd-ee-ff-gg"),
        // Invalid hex characters
        U("gg-00-00-00-00-00"),
        U("aa-bb-cc-dd-ee-GG"),
        // Missing separators
        U("aabbccddeeff"),
        U("00000000000000"),
        // Wrong separator positions
        U("aab-bcc-dde-eff"),
        U("aa-bbccdd-ee-ff"),
        // Mixed separators
        U("aa:bb-cc-dd-ee-ff"),
        U("aa-bb:cc:dd-ee-ff"),
        // Extra characters
        U(" aa-bb-cc-dd-ee-ff"),
        U("aa-bb-cc-dd-ee-ff "),
        U("aa-bb-cc-dd-ee-ff-"),
        U("-aa-bb-cc-dd-ee-ff"),
        // Special characters
        U("aa.bb.cc.dd.ee.ff"),
        U("aa_bb_cc_dd_ee_ff"),
        // Empty octets
        U("--bb-cc-dd-ee-ff"),
        U("aa--cc-dd-ee-ff"),
        // Single character octets
        U("a-b-c-d-e-f"),
        U("0-0-0-0-0-0"),
        // Three character octets
        U("aaa-bbb-ccc-ddd-eee-fff")
```

lo-simon: Extend the test to cover various malformed port IDs

- Wrong lengths (too short/too long)
- Invalid hex characters
- Missing/wrong separators
- Mixed separators
- Extra whitespace/characters
- Special characters
- Empty/single/triple character octets

lo-simon (changes requested):
Thank you, @shin4141, for your contribution. I have some suggestions to simplify the port ID validation and extend the test for various malformed port IDs.

Issue #109, opened 2020-01-16
Logging Boost error messages as UTF8


aholzinger:
By default Boost error messages when using Visual Studio are ANSI encoded and not UTF-8. To force Boost to emmit strings in UTF-8 the `#define BOOST_SYSTEM_USE_UTF8` is needed. It could easily be added to Development/cmake/NmosCppCommon.cmake after line 147 like this:
`add_definitions(/DBOOST_SYSTEM_USE_UTF8)`

Unfortunately the console also needs a hint that printed text is UTF-8. I had to fiddle a little bit, but finally I found a portable solution:
```
#include <clocale>

static bool make_utf8()
{
    std::string locale = setlocale(LC_CTYPE, "");
    if (!locale.empty()) {
        auto dot = locale.find_last_of('.');
        if (dot != std::string::npos) {
            locale = locale.substr(0, dot) + ".utf8";
            char* result = setlocale(LC_CTYPE, locale.c_str());
            return result != nullptr;
        }
    }
    return false;
}
```
After initialising the log:
```
if (!make_utf8())
{
    slog::log<slog::severities::error>(gate, SLOG_FLF) << "Setting console to UTF-8 mode failed";
}
```

Tested under Windows and Linux.

garethsb-sony:
Thanks for the thorough investigation!

garethsb-sony:
A couple of questions:

* Presumably we need to also build Boost libraries (those which aren't header-only) and C++ REST SDK with `BOOST_SYSTEM_USE_UTF8`?
* Could we achieve the same effect for logging via a call to e.g. `std::cerr.imbue(...)` (getting its current locale first if required) rather than setting the global locale with `setlocale`?

aholzinger:
1. Good question: I really don't know, but I guess it would make sense. On my dev machine I did **not** compile boost with `BOOST_SYSTEM_USE_UTF8` and the ASIO error messages are correct ("_... ungültig ..._"), but I can't imagine that compiled Boost code can emmit correct UTF-8 messages if Boost is not compiled with `BOOST_SYSTEM_USE_UTF8`.
2. I'm not an expert. I made a short recherche and I doubt that it can be done via imbue, but what about this?
[https://www.boost.org/doc/libs/master/index.html](url)
Well it needs upcoming Boost version 1.73.

garethsb-sony:
Hi Axel,

Sorry, did you mean to put in a more specific link in 2?

[...]

PR #523, opened 2026-09-22
Fix crash logging non-UTF-8 narrow strings

marcusfreisleben:
## Problem

Logging a message containing non-UTF-8 narrow bytes (e.g. a `std::exception::what()` built from a Windows ACP/CP1252 string, or a `__FILE__` path with an accented character) crashes the app.

`utility::s2us` requires valid UTF-8 and throws `std::range_error` otherwise. `nmos::experimental::details::json_from_message` calls it directly on `message.str()`, `message.file()`, `message.function()` and categories, on the async logging worker thread. `slog::message_service::run` only catches its own `stopped_exception`, so the throw escapes the thread function and terminates the process.

Confirmed on a german Windows 11 installation.

## Fix

- Add `utility::s2us_lenient` (cpprest/basic_utils.h/.cpp): UTF-8 fast path, Windows ANSI code page fallback, Latin-1 fallback otherwise. Never throws, never discards the message. `s2us`'s contract and existing call sites are unchanged.
  - Validates with `conversions::to_utf16string`, the same decoder `s2us` uses on a wide build, so the fast path accepts exactly what `s2us` accepts. Note this decoder is not a strict UTF-8 validator (tolerates surrogates, overlongs, code points above U+10FFFF) - a hand-rolled check would have diverged from it and reopened the bug.
  - Converts the original string rather than the successfully-decoded value, so on the fast path the result is byte-identical to what `s2us` would have returned, on narrow builds too.
  - Skips the ANSI code page when it is itself UTF-8 (Windows 10+), where `MultiByteToWideChar` would just repeat the failed decode and substitute U+FFFD, losing the characters at issue.
- Use it in `json_from_message` (message, file, function, categories) and at the `what()`/error-message boundaries in the API error responses (`api_utils`, `connection_api`, `channelmapping_api`, `authorization_redirect_api`, `control_protocol_ws_api`).
- Guard `log_gate::service` as a last-resort safety net, so no future logging failure can terminate the app.

## Testing

- `basic_utils_test.cpp`: valid UTF-8 fast path, invalid-byte fallback (CP1252 umlaut and others), decoder-tolerated sequences pass through unchanged, empty input.
- `log_gate_test.cpp`: `testInsertLogEventNonUtf8` builds a log message with CP1252 umlaut bytes through the same path that crashed and asserts the emitted JSON is valid UTF-8. This fails against the pre-fix code (confirmed) - on non-Windows too, since `s2us` is a pass-through there and the invalid bytes previously reached the JSON unmodified.
- Built and tested on macOS (narrow build) and confirmed on Windows.

## Open question for maintainers

There are ~15 other `utility::s2us(...)` call sites taking narrow strings of uncertain origin (mDNS host names, JWT claims, OpenSSL PEM output, etc). Left out of this PR to keep it focused on the crash. Happy to follow up with `s2us_lenient` at those sites too if that's wanted, or to fold `s2us_lenient` behavior into `s2us` itself if you'd rather not carry two functions - whichever you prefer.


garethsb:
How does this relate to #109?

garethsb:
I wonder what the minimum blast radius is... is it better to wrap every external API call that can return/throw a non-UTF8 string or to handle such strings as close as possible to outputting them?

marcusfreisleben:
> How does this relate to #109?

It does sound like it could be related, yeah. Fixing it defensively at the consumer side has the benefit of also covering other non-Boost sources and we don't have to diverge from the official conan boost recipe.

